PR #23, opened 2019-08-29
Removes TG maps and updates the basemap.dm file

crossedfall:
## About The Pull Request

This removes all of the standard TG maps from the repo and updates the basemap.dm file to correctly reflect what we have.

## Why It's Good For The Game

It makes map development easier since we don't have a bunch of junk maps sitting around anymore. The `basemap.dm` updates will also allow the map diff bot to do its job.


crossedfall:
> 
> 
> Far as I understand it from Kmc, whenever we pull from our upfork we tend to get these maps once more.
> Non the less I don't entirely opose this, we'll just most likely get them back later on.

This isn't a fork and that'd only happen if we lazily bulk migrate all of TG's changes. Something I'd _**hope**_ we wouldn't/don't do.

Kmc2000:
Why?
Those maps are great reference for complicated setups like atmos, buttons, machines, decals etc.

GoatMonarch:
> Why?
> Those maps are great reference for complicated setups like atmos, buttons, machines, decals etc.

Good point. I'd argue we could thrim the fat somewhat though.

crossedfall:
> 
> 
> Why?
> Those maps are great reference for complicated setups like atmos, buttons, machines, decals etc.

Wouldn't it be better for new mappers to reference our maps for all of that? Also, you can freely reference them from TG's codebase. There's honestly no reason to keep them here as it just adds bloat.

GoatMonarch:
> > Why?
> > Those maps are great reference for complicated setups like atmos, buttons, machines, decals etc.
> 
> Wouldn't it be better for new mappers to reference our maps for all of that? Also, you can freely reference them from TG's codebase. There's honestly no reason to keep them here as it just adds bloat.

Currently as is we're kind of lacking some base mapping features. Mule bot pathing, disposal delivery systems and the like.
We're more on par with space_ship than the Aetherwhisp to put it into FTL13 terms of polish here, and it's a fairly neat thing to copy-paste certain objects from one map to another.

crossedfall:
> 
> 
> > > Why?
> > > Those maps are great reference for complicated setups like atmos, buttons, machines, decals etc.
> > 
> > 
> > Wouldn't it be better for new mappers to reference our maps for all of that? Also, you can freely reference them from TG's codebase. There's honestly no reason to keep them here as it just adds bloat.
> 
> Currently as is we're kind of lacking some base mapping features. Mule bot pathing, disposal delivery systems and the like.
> We're more on par with space_ship than the Aetherwhisp to put it into FTL13 terms of polish here, and it's a fairly neat thing to copy-paste certain objects from one map to another.

So then having a properly made map more-or-less solves this issue? Why not upgrade the current ship to have these systems so that we can use it as our basemap? Rather than keeping a load of maps we'll never use in practice.

GoatMonarch:
> > > > Why?
> > > > Those maps are great reference for complicated setups like atmos, buttons, machines, decals etc.
> > > 
> > > 
> > > Wouldn't it be better for new mappers to reference our maps for all of that? Also, you can freely reference them from TG's codebase. There's honestly no reason to keep them here as it just adds bloat.
> > 
> > 
> > Currently as is we're kind of lacking some base mapping features. Mule bot pathing, disposal delivery systems and the like.
> > We're more on par with space_ship than the Aetherwhisp to put it into FTL13 terms of polish here, and it's a fairly neat thing to copy-paste certain objects from one map to another.
> 
> So then having a properly made map more-or-less solves this issue? Why not upgrade the current ship to have these systems so that we can use it as our basemap? Rather than keeping a load of maps we'll never use in practice.

Yes, it'd be perfect if we had a properly made map, though the current goal is to have a map where player tests can be made on, later once the base systems are in I'll be able to get to work on a more fleshed out map which probably will take more time to get going to a playable stage.
We'll most certainly get there, but for the moment I'll just leave the decision up to one of the project leads.

crossedfall:
> > So then having a properly made map more-or-less solves this issue? Why not upgrade the current ship to have these systems so that we can use it as our basemap? Rather than keeping a load of maps we'll never use in practice.
> 
> Yes, it'd be perfect if we had a properly made map, though the current goal is to have a map where player tests can be made on, later once the base systems are in I'll be able to get to work on a more fleshed out map which probably will take more time to get going to a playable stage.
> We'll most certainly get there, but for the moment I'll just leave the decision up to one of the project leads.

We are starting to get a little off topic, but I will say this much. I've always been of the mind that if things are worth doing, they're worth doing right the first time. Pushing the goal further out doesn't do us any good. Having a well made map right from the start is arguably critical for having successful testing. If we're looking to settle now, this early, that sets a bad precedent for the future.

Having a mediocre first test vs having a fantastic first test should be a no brainier. You only get one first impression after all, and in a niche community like this that counts for a lot.

Kmc2000:
The Aegis isn't that bad anymore, as people have helped fix it. It was originally intended as just a testship, but I think it's as viable as any other map, and should be used as the reference until a better one is made. For non ship related things though, it's really useful to have a reference. So, I'd say remove all maps but meta if we want reduced bloat, as meta's probably the most fleshed out /tg/ map.

sephora.dme
```diff
@@ -1872,7 +1872,6 @@
 #include "code\modules\jobs\job_types\station_engineer.dm"
 #include "code\modules\jobs\job_types\virologist.dm"
 #include "code\modules\jobs\job_types\warden.dm"
-#include "code\modules\jobs\map_changes\map_changes.dm"
```
francinum: Replace the pubby data in this fine with an include, Don't uninclude it.

crossedfall: The file only exists for pubby, there's not much reason to keep it since it'd be a "blank" file that's entirely commented out. If that's fine with everyone then sure.

crossedfall:
> 
> 
> The Aegis isn't that bad anymore, as people have helped fix it. It was originally intended as just a testship, but I think it's as viable as any other map, and should be used as the reference until a better one is made. For non ship related things though, it's really useful to have a reference. So, I'd say remove all maps but meta if we want reduced bloat, as meta's probably the most fleshed out /tg/ map.

I still think you could very easily reference meta, or any of the TG maps, via TG's codebase rather than our own. But, sure. I'll re-add meta for the time being.

sephora.dme
crossedfall: Got added clarification from @francinum in discord. It was decided that he wanted the file even if it would be nothing but comments. Re-adding it per his request and with his comment suggestions.

PR #29, opened 2019-09-01
Floyd's Mining Rework

Qustinnus:
:cl:
 add: Reworks mining completely
/:cl:

Reworks mining completely, replacing with a system that relies heavily on the new /datum/material system instead.

You can now go to a mineral magnet, activate it to pull a nearby asteroid in, and start mining. By default asteroids are mostly trash unless you find specific asteroids by flying around and looking for them (this part is WIP)

Instead of mining rocks to get ore, the current system involves around boulders. You can use your scanner on any object to find out what materials it contains. This is useful as it will tell you what rocks are made of before you mine them. This is usually a composition of mostly sand, and any other material inside. For example a rock could be 70% sand, 20% iron, and 10% plasma

Once you find one with materials you want you can dig the wall. It will leave behind a large boulder with the materials. By default a boulder has 40k units of material. (20 sheets). 

Once you've mined a boulder, you can drag it back to your mining base and put it into the crusher, this'll leave behind 5 rocks each made out of the same composition of the boulder but 5 times less the units. (5 * 8000 = 40k).  Now these rocks will go down the conveyor belt into various machines, which you can either turn on, or off. The effects they have on rocks vary on what materials are inside of the rock. You can see a full chart of the current effects below.

image

As an example, if you were to have a rock with plasma in it, the rock would evaporate if it goes through the rock cleaner, and even blow up if it goes into the rock burner. These effects combined with the proceduraly generated nature of the rock, prevent you from leaving the same machines on by default. A picture of the default set-up can be seen below. (Excuse the sprites, theyre placeholder)

image

The question still remains if we have all the machines unlocked at roundstart, or add progression to them so you can't get them all at roundstart.


nsv13/code/modules/mining/core-mining/asteroid_magnet.dm
```diff
@@ -0,0 +1,53 @@
...
+		pull_in_asteroid()
+	if(href_list["push_asteroid"])
+		push_away_asteroid()
+
+	attack_hand(usr) //Refresh window
+
+/obj/machinery/mineral_magnet/proc/pull_in_asteroid()
+	current_asteroid = new /datum/map_template/asteroid(pick("_maps/asteroids/asteroid1.dmm")) //Set up an asteroid, can specify a core composition as argument
+	current_asteroid.load(target_location, TRUE)
+	message_admins("gay")
```
ike709: 👍

nsv13/code/modules/mining/core-mining/asteroid_magnet.dm
```diff
@@ -0,0 +1,53 @@
...
+	//addtimer(CALLBACK(src, .proc/load_asteroid), 20) //2 seconds to fuck off
+
+/obj/machinery/mineral_magnet/proc/load_asteroid()
+	current_asteroid.load(target_location, TRUE)
+	message_admins("gay2")
```
ike709: 👍2

PR #31, opened 2019-09-04
Boarding Template Expansions

Vasily2013:
## Why It's Good For The Game

This adds damaged variants of boarding maps already created by me.

## Changelog
:cl:
add: broken versions of each ship, boarder, corvette, prison ship, and yacht.
add: munitions assets to corvettes.
/:cl:
image
image
image
image
image
image
image
image


Qustinnus:
It might be better to use FTL13's damage system, which would generate new damage each time

Vasily2013:
> It might be better to use FTL13's damage system, which would generate new damage each time

Well, unfortunately I don't know what damage system FTL13 used, and therefore can't use it here. If you'd like to implement it once this is merged I would be more than happy to offer map support to make it feasible.

PR #68, opened 2019-09-29
Icon state of turfs refactoring

UndeadNappist:
moves the handlers up for mineral floors up and genericizes them. Should fix a few issues, but creates others under unqiue conditions.

Note to self: Test before using - codebase and updates may have messed with things.

have been used in same codebase from previous vers. to fix various bugs.

## Changelog
refactor: refactored some code
/:cl:


UndeadNappist:
Testing done. Works fine. Now with bonus commentary.

UndeadNappist:
Works fine, tested fine. However, there's a runtime involved that is inherent to the master branch and not my changes.
[13:17:17] Runtime in atoms.dm,924: type mismatch: 0 |= the plating (122,125,2) (/turf/open/floor/plating)

Looks like someone screwed up typing with crowbars somewhere.

code/game/turfs/simulated/floor.dm
```diff
@@ -21,6 +21,13 @@
 	var/floor_tile = null //tile that this floor drops
 	var/list/broken_states
 	var/list/burnt_states
+/*
```
Kmc2000:
```suggestion
/*Nsv13 - Turfs
```

code/game/turfs/simulated/floor.dm
```diff
@@ -56,6 +63,8 @@
 		icon_regular_floor = icon_state
 	if(mapload && prob(33))
 		MakeDirty()
+	if (!icons)
```
Kmc2000:
```suggestion
	if(!islist(icons))
```

code/game/turfs/simulated/floor.dm
Kmc2000: We do this to track changes across rebases more easily.

PR #74, opened 2019-10-03
Reverts the electrical system back to Classic Wires

Emotions211:
## About The Pull Request
Reverts this PR https://github.com/tgstation/tgstation/pull/44265 and subsequent PR's related to it, returning classic cables in a stable form.

## Why It's Good For The Game
Allows multiple power nets in close proximity, even overlapping, and removes the need to map in wall cables in tight spaces to separate power nets.

## Changelog
:cl:
del: any reference in the /tg/ base to should_have_node
add: classic cables
add: returns Rapid Cable Layers to true cable laying capacity.
add: mecha RCL's have been re-enabled
map: re-did Aegis wiring to classic cables.
config: fixed cable coil spawns to be random colours
/:cl:
// reverting admin/verb/atmosdebug.dm seems to cause instability. someone with more knowledge should look into that if it causes issues.


Qustinnus:
This pr is both broken and just bad, we will never be able to sync tg again, if we do this lol

PR #100, opened 2019-10-23
Stormdrive Percentile Control Rod Insertion Values

Karmic-Skink:
## About The Pull Request
This pull request changes the stormdrive control rod system to allow for tweaking of power outputs, rather than 4 set states.

[https://cdn.discordapp.com/attachments/602204840180842497/636362916806787095/unknown.png](https://cdn.discordapp.com/attachments/602204840180842497/636362916806787095/unknown.png)

## Why It's Good For The Game
Allows engineers greater control over the engine.

## Changelog
:cl:
add: added manual control rod insertion depth option for the storm drive
tweak: tweaked the calculation for heat generation via control rods
/:cl:


nsv13/code/modules/power/stormdrive.dm
```diff
@@ -199,18 +195,28 @@ Takes  plasma and outputs superheated plasma and a shitload of radiation.
...
+		reactor.control_rod_percent = 0
 		message_admins("[key_name(usr)] has fully raised reactor control rods in [get_area(usr)] [ADMIN_JMP(usr)]")
 		reactor.update_icon()
 	if(href_list["rods_2"])
-		reactor.control_rod_state = RODS_HALFRAISED
+		reactor.control_rod_percent = 33.42
 		reactor.update_icon()
 	if(href_list["rods_3"])
-		reactor.control_rod_state = RODS_HALFLOWERED
+		reactor.control_rod_percent = 23.49
```
gryffinp: These numbers are backwards: according to the control_rod_percent formula on line 445, position 2 will now have a target heat of 100, and position 3 will have a target heat of 200. Because position 1 is the maximum heat position and position 4 is the maximum cooling position, this means that the heat level of the control rod presets would be 1>3>2>4.

Karmic-Skink: resolved as outdated

nsv13/code/modules/power/stormdrive.dm
```diff
@@ -199,18 +195,28 @@ Takes  plasma and outputs superheated plasma and a shitload of radiation.
...
-		reactor.control_rod_state = RODS_LOWERED
+		reactor.control_rod_percent = 100
 		reactor.update_icon()
+	if(href_list["rods_c"])
+		if(alert(usr, "Manually setting control rods positions should only be attempted by experienced nuclear engineers.", "EXTREME CAUTION ADVISED", "Continue", "Abort")=="Continue")
+			var/new_control_rod_percent = input(usr, "Set new control rod insertion depth percent", "Manual Control Rods Positional Control", reactor.control_rod_percent) as num|null
+			if(!usr.canUseTopic(src, BE_CLOSE))
+				return
+			new_control_rod_percent = CLAMP(new_control_rod_percent, 0, 100)
+			reactor.control_rod_percent = new_control_rod_percent
```
gryffinp:
```suggestion
			if (!isnull(new_control_rod_percent))
				new_control_rod_percent = CLAMP(new_control_rod_percent, 0, 100)
				reactor.control_rod_percent = new_control_rod_percent
```
If a player hits cancel on the percentage prompt, the control rod percentage gets set to null. This is approximately the worst possible thing, because it sets the icon of the reactor to display as if the rods are 100% inserted, but actually causes the reactor to not cool at _all._ Wrapping the assignment in an isnull() if statement should result in what I think is more intuitive behavior, which is retaining the current percentage on cancelling the prompt.

nsv13/code/modules/power/stormdrive.dm
```diff
@@ -353,7 +359,15 @@ Takes  plasma and outputs superheated plasma and a shitload of radiation.
...
-		add_overlay("rods_[control_rod_state]")
+		switch(control_rod_percent)
+			if(0 to 24)
+				add_overlay("rods_1")
+			if(25 to 49)
+				add_overlay("rods_2")
+			if(50 to 74)
+				add_overlay("rods_3")
+			if(75 to 100)
+				add_overlay("rods_4")
```
gryffinp: Bug: Setting the control rod percentage to a number in between one of these ranges (e.g. 24.5) results in the rods disappearing entirely, as if set to rods_4. I don't know byond code well enough to know the proper fix.

Suggestion: In this current distribution, rod position 1 and 2 use the same icon, rod 3 uses icon 2, and icon 3 never gets used except for mostly-pointless manual assignments.

I suggest the ranges should be 0-10 for "rods_1"/fully raised (because that is the range at which the reactor will actually explode), 11-24 for "rods_2"/mostly-raised (between ~450-200 degrees, i.e. what was considered at least "hot" in the old definitions), 25-52  for "rods_2"/mostly-inserted ("safe" temperatures, within the realm of rod position 3), and 53-100 for "rods_4"/SCRAM (This is the range at which the reaction will actually just die, slowly at about 53% and pretty rapidly once you go above 55%.

That way, the visual positions of the control rods will provide hints to engineers who didn't do the math of the expected results of their terrible decisions.

Karmic-Skink: Not so much of a bug as a lazy get around while I did other things.

Karmic-Skink: Fixed.

nsv13/code/modules/power/stormdrive.dm
Karmic-Skink: tbh I wanted a TGUI solution rather than what I did.

Karmic-Skink: Should function correctly now.

PR #106, opened 2019-10-29
Remaps atmospherics to be better by a good degree?

Jalleo:
Does what it says on the tin made a proper distro link and waste link removed some bad placing of pipes. Added in some port mixing and some atmos tanks along with more scrubbers and some air pumps.
I plan to add in space heaters and a incinerator next week as the focus was getting a good atmos setup for now. The gas miners are there mainly to assist to that and to add something for atmospheric technicians to work towards with science to make some gasses.
Old 
Old
New
new


Vasily2013:
Pictures please.

Vasily2013:
Change Requests:
-Remove the gas miners from the supply tanks. Atmospherics is meant to be a limited resource, even though it contains enough to normally keep the station/ship running for a couple hours.
-CO2 supply console is not aligned with the filter, while all the other consoles are.
-Add space heaters? I don't see any, and true KMC didn't put any in there, but atmos should have them just like they should have pumps and scrubbers. 
-A huge scrubber or two might not be amiss either.

Suggestions from an atmos perspective
-move the O2 pump from 66,121,2 to 67,123,2 and point it east towards the mix filter. This allows for finer control between mix and pure.
-Do the same sort of move for the air mix to distro pump, leave the vent exposed until the first pump going to either pure or distro.
-Move the freezer from 77,128,2 to 79,128,2 and point the connector West. If you still need the third connector there then you can swap it, that leaves the open space normally instead of keeping a machine there.



Overall this looks like a good change to the atmospherics workspace, and I think that the added assets of scrubbers and pumps will prove useful during gameplay. Apart from the few change requests I have no other complaints or comments.

GoatMonarch:
I'd like to see some pictures as well, given I am not really able to view this otherwise till the upcomming weekend.

Though I wish to note that the gass miners within the tanks are a fine addition as long as there's a turbine somewhere aboard, as they can be quite a hog related to gasses, especially on smaller tanks like here.

Jalleo:
I plan to add the turbine next week.
I will fix the CO2 filter tonight.
The other requests I shall conside tonight. I have made the setup inefficient to a degree so its not perfect with stuff like the freezer and heater.
The atmos room itself will also probably change when I add in the incinerator stuff next week.

Jalleo:
I have moved the CO2 computer to line up with the filters as per request.
The O2 pump I have not moved as it would change its position from where all the tanks have it and it does not require one right before the mixer.
I have added three space heaters as well now.

Vasily2013:
Fair enough. I just don't see much of a point of having the pumps double stacked right there in one direction of flow.  I still think the gas miners are cheesy on a ship, but i'm not going to let that stop this PR. Again, I would say put in two huge scrubbers. Perhaps one right along the wall between the Plasma and N20 tank, then another between the o2 and air mix tank.

Other than the above considerations I approve.

Jalleo:
If you are talking about the ones usually seen in research I will say those cant be added as they are too effective as scrubbers.

GoatMonarch:
I am fairly certain TG does not have large scrubbers aviable for a reason.
Even if they were too efficient as a scrubber, the fact they can't connect to the atmos piping is kind of shid. So I say don't bother with them.

Vasily2013:
> If you are talking about the ones usually seen in research I will say those cant be added as they are too effective as scrubbers.

Rebuttle: Toxins on box starts with one, and they are rarely used even though they are effective due to their lack of mobility. Bolt them to the deck at round-start and I promise you nine out of ten times they'll never even be turned on.

Jalleo:
Either way its something which isn't really needed its a preference that you are asking for. The focus is to get atmos remapped to make the map better overall. Then next week in another PR the incinerator gets added to allow a new focus on remapping the entire rear of the vessel over two weeks or so after that.

It will take some time until a proper ship gets added by a amazing mapper due to their situation or feeling of need so the focus is getting things better for the temporary ship. As well as giving it the needs that people have for it at this time.

ta3370:
If you're gonna have a plasma gas miner in the plasma tank, a large perma-bolted scrubber outside the tank as Vasily suggested doesn't seem too unreasonable. 
The super powerful large scrubber would perhaps be required to deal with any breach of the plasma tank since it has infinite output which will flood the atmos deck and the engine bay.

PR #179, opened 2019-12-25
Fixes travis maybe

ike709:
Maybe fixes travis.


ike709:
Since the check doesn't support lists, it'll no longer cause travis to fail.

Aside from lack of list support, this is a fairly important check. It should be rewritten to support lists at some point. @Kmc2000

PR #204, opened 2019-12-29
Fix: power sink's power_drained change limited to APC cell's charge 

covertcorvid:
## About The Pull Request
In current code, if a powersink finds an APC cell to drain from, it increases its power drained by 50 watts even if the APC has less than 50 watts of charge. This fix makes power_drained increase by the minimum of either the cell's charge or 50.

## Why It's Good For The Game
Power is not created from nothing

## Changelog
:cl:
fix: corrected powersink behavior where APC cell has less than 50 watts left to drain
/:cl:


code/game/objects/items/devices/powersink.dm
```diff
@@ -141,8 +141,8 @@
 				if(istype(T.master, /obj/machinery/power/apc))
 					var/obj/machinery/power/apc/A = T.master
 					if(A.operating && A.cell)
```
Jalleo: add a extra && check seeing if the cell.charge is greater than zero.

code/game/objects/items/devices/powersink.dm
```diff
@@ -141,8 +141,8 @@
 				if(istype(T.master, /obj/machinery/power/apc))
```
Jalleo: needs a CHECK_TICK or STOP_LAG() before this line encase we start getting way too many nodes in the far future. This could technically run for a long ass time

covertcorvid:
BTW, this fix was already merged on Bee without the performance improvements https://github.com/BeeStation/BeeStation-Hornet/pull/1075

PR #208, opened 2019-12-29
Attempt at seperating the overmap from station at backend code

Jalleo:
## Why It's Good For The Game
KMC wanted it done
## Changelog
:cl:
code: touched mapping.dm to try and seperate the ship and overmap code (NULL RODS WONT GO FLYING ON A OVERMAP FOR A CHANGE!)
/:cl:

### DIFF
--- a/code/datums/map_config.dm
+++ b/code/datums/map_config.dm
@@ -24,6 +24,9 @@
 
 	var/minetype = "lavaland"
 
+	var/overmap = null //NSV13 Stuff with overmap code
+	var/over_traits = null //NSV13 Stuff with overmap code
+
 	var/allow_custom_shuttles = TRUE
 	var/shuttles = list(
 		"cargo" = "cargo_box",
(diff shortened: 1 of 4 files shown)


Jalleo:
Current status. It works and compiles. The problem is that the FTL is now broken. The specific line is probably
https://github.com/DDMers/NSV13/blob/02bf2eff8266b4d010247757d339e720fe45886b/nsv13/code/modules/overmap/ftl.dm#L204

code/datums/map_config.dm
```diff
@@ -143,6 +143,7 @@
 			if (!fexists("_maps/[map_path]/[file]"))
 				log_world("Map file ([map_path]/[file]) does not exist!")
 				return
+
```
Jalleo: Or not I was just tired

Kmc2000:
Supremely based jalleo

Can you mod the other JSONS for me?

Jalleo:
I will only change the ones that are in active use and are playable otherwise they are not going to be active until the rework or fixing of issues for those maps. Better to force a focus on that. I am starting on the aegis already.

PR #246, opened 2020-01-08
Configurable ranks

covertcorvid:
## About The Pull Request
Loads ranks for each job from a configuration file, and allows admins to switch between rank configurations.

## Why It's Good For The Game
Rank adjustments no longer require modifications to individual job files. Also we can do memes.

## Changelog
:cl:
add: Ability to set rank-to-job mappings from configuration file
refactor: Moved NSV13 compose_rank say.dm function into NSV folder
refactor: Moved NSV13 config entries into NSV13 folder
config: Added a config option for rank file, and two rank config files.
admin: Added admin verb to change rank structures
server: New config/ranks directory and config files
/:cl:


francinum:
Looks good, does file access still bring up the view window as was seen during the showcase snippet?

covertcorvid:
> 
> 
> Looks good, does file access still bring up the view window as was seen during the showcase snippet?

The View and Open options? Yeah, those work exactly like the log viewer which is where I stole the code from.

francinum:
You may want to edit that interaction to be more clear.

nsv13/code/controllers/subsystem/ranks.dm
```diff
@@ -0,0 +1,59 @@
...
+
+/client/proc/browse_rank_configs(path = "config/ranks/")
+	path = browse_files(path)
+	if(!path)
+		return
+
+	if(file_spam_check())
+		return
+
+	message_admins("[key_name_admin(src)] changed rank configuration to: [path]")
```
francinum: This should be called ONLY when it's actually loaded, otherwise it should report an access.

code/controllers/configuration/entries/general.dm
```diff
@@ -497,6 +497,4 @@
 
 /datum/config_entry/flag/grant_metacurrency
 
-/datum/config_entry/flag/respect_global_bans
-
-/datum/config_entry/flag/show_ranks //nsv13
\ No newline at end of file
+/datum/config_entry/flag/respect_global_bans
```
Jalleo: needs a newline after this line.

config/ranks/corporate.txt
```diff
@@ -0,0 +1,50 @@
+Captain=Director
```
Jalleo: I personally think Captain should be captain on both because that is how it is on all servers basically.

nsv13/code/controllers/subsystem/ranks.dm
```diff
@@ -0,0 +1,59 @@
+//For code/controllers/subsystem/job.dm
+/datum/controller/subsystem/job/proc/LoadRanks(rankfile="config/ranks/military.txt")
+	if (fexists("[rankfile]"))
+		var/rankstext = file2text("[rankfile]")
+		for(var/datum/job/J in occupations)
+			var/regex/jobs = new("[J.title]=(.+)")
+			jobs.Find(rankstext)
```
Jalleo: Test to see what happens if a job doesnt have a rank and what happens ingame when that occurs. See about adding a alert that is visible in world when that happens while this is loading so everyone can notice and report it to get it fixed.

config/ranks/corporate.txt
covertcorvid: Talk to @Vasily2013 about this one.

Vasily2013: Request Denied.

Further explanation: The director of the ship is the captain, captainship is a position, not a rank under the corporate rank structure. It's like saying that the "captain" of a passenger plane isn't actually a captain, he's a commercial pilot. The title is an honorific, not indicative of position other than job, which is still called in the codebase 'captain'.

nsv13/code/controllers/subsystem/ranks.dm
covertcorvid: All other jobs are updated and the missing job retains its old rank. What do you think is desirable behavior - default to no rank, or abort the change entirely?

covertcorvid:
> 
> 
> You may want to edit that interaction to be more clear.

@francinum What's unclear about it so that I can make it better?

nsv13/code/controllers/subsystem/ranks.dm
Jalleo: Ask Kmc.

covertcorvid:
Substituting missing ranks with Assistant rank:
useassistant

Substituting missing ranks with random rank:
noassistant

PR #248, opened 2020-01-08
False walls and correct resources from duranium walls

### LINKED ISSUE #63: Duranium Deconstruction returns only 1 sheet
## Round ID:

11

## Testmerges:

#49 - Irrelevant for this

## Reproduction:

Deconstruct a Duranium wall, expect to get 2 sheets back, only get 1
Be unable to patch the hole you just made.

### LINKED ISSUE #241: Dissasembling a renforced wall gives incorrect resources
## Round ID:
Any round where you weld down a wall, or a breach is caused by an explosion.

## Testmerges:
**As if Kmc would test merge his code**

## Reproduction:
Disasemble a renforced wall with tools will give you one durasteel plate, and you later aquire a sheet of plasteel once you remove the renforcement of the wall. 
This makes me angerry.

### LINKED ISSUE #242: False Walls are not a thing with NSV sprites
As titled. For the love of antags, KMC. For the love of antags.

covertcorvid:
## About The Pull Request
Adds false walls for both wall types. Fixes bug where constructing duranium wall took 2 duranium but returned 1 duranium and some plasteel.
Fixes #63 
Fixes #241 
Fixes #242 

## Why It's Good For The Game
Antags and morale support officers can be sneaky. Walls that are taken down can be patched.

## Changelog
:cl:
add: Fake walls
fix: Fixes incorrect materials from deconstructed r-walls
imageadd: Fake walls
/:cl:


covertcorvid:
Not sure where the hourglass giving me a runtime is. Will investigate later if no one knows.

francinum:
Why is this entire chunk of code copypasted instead of extending the behavior of https://github.com/BeeStation/NSV13/blob/master/code/game/objects/structures/girders.dm#L167

covertcorvid:
@francinum It was already like this, probably to stop people from building TG walls: https://github.com/BeeStation/NSV13/blob/master/nsv13/game/custom_turfs.dm I just indented it another level so I could check if the girder was wrenched or not.
Did you want me to subclass girder instead? I think that'd let us still return before calling the parent if it's a type of sheet we don't want to handle.

nsv13/game/custom_turfs.dm
```diff
@@ -4,28 +4,52 @@
...
-			to_chat(user, "<span class='warning'>You need two sheets of [W] to finish a wall!</span>")
-			return
-		if(!S.turf_type)
-			to_chat(user, "<span class='warning'>You can't build anything with [W].</span>")
+			to_chat(user, "<span class='warning'>You need two sheets of [S] to finish a wall!</span>")
 			return
+
+		to_chat(user, "<span class='notice'>You start plating the wall with [S]...</span>")
 		if (do_after(user, 40, target = src))
 			if(S.get_amount() < 2)
```
Jalleo: you should add in the to_chat warning here that you need extra sheets I believe.

nsv13/game/custom_turfs.dm
```diff
@@ -36,7 +60,31 @@
...
+
+	else //Build false wall
+		if(S.get_amount() < 2)
+			to_chat(user, "<span class='warning'>You need two sheets of [S] to create a false wall!</span>")
+			return
+
+		to_chat(user, "<span class='notice'>You start building a false wall with [S]...</span>")
+		if (do_after(user, 40, target = src))
+			if(S.get_amount() < 2)
+				return
```
Jalleo: Add in a warning here that you need the two sheets otherwise the person might not know if they are making more than one false wall at a time.

nsv13/game/custom_turfs.dm
```diff
@@ -4,28 +4,52 @@
...
+	smooth = TRUE
+	walltype = /turf/closed/wall/r_wall/ship
+	canSmoothWith = list(/turf/closed/wall/ship,/obj/machinery/door,/obj/structure/window,/turf/closed/wall/r_wall/ship, /obj/structure/falsewall/duranium, /obj/structure/falsewall/durasteel)
+	mineral = /obj/item/stack/sheet/duranium
+
+/obj/structure/girder/proc/try_nsv_walls(obj/item/stack/sheet/S, mob/user)
+	if(!S.turf_type) //Let the girder code handle it
+		return
+
+	if (state != GIRDER_DISPLACED) //Build regular wall
```
Jalleo: remove the space before the (

nsv13/game/custom_turfs.dm
francinum: This is the success path. `S.use(2)` needs to be called as well to actually use the material from the stack.

francinum: As well as actual wall creation unless I'm missing something.

nsv13/game/custom_turfs.dm
```diff
@@ -36,7 +60,31 @@
...
+		to_chat(user, "<span class='notice'>You start building a false wall with [S]...</span>")
+		if (do_after(user, 40, target = src))
+			if(S.get_amount() < 2)
+				return
+			S.use(2)
+			to_chat(user, "<span class='notice'>You create a false wall. Push on it to open or close the passage.</span>")
+
+			var/turf/T = get_turf(src)
+			var/obj/structure/falsewall/FW
+			if (istype(S, /obj/item/stack/sheet/durasteel))
```
francinum: Why are these checks, `turf_type` contains the required path.

Jalleo: var/F = text2path("/obj/structure/falsewall/[M]")
			var/obj/structure/FW = new F (loc)
this removes the need for two istype checks as it gets the stack name to use for the actual path.

Jalleo: This is also what the other false walls do later on in the girder code.

nsv13/game/custom_turfs.dm
covertcorvid: Added warning.
Wall creation was just collapsed by github because it hadn't changed. Visible now that I changed the return.

nsv13/game/custom_turfs.dm
covertcorvid: Only for real walls, not false ones

covertcorvid: Only worked for minerals. Added sheettype to our mats and checked that they are not minerals.

covertcorvid:
> 
> 
> See comments.
> Also normal walls wont be made due to what you have setup so far so this PR cant be merged.

Normal walls were made, github just hid the unchanged lines.

francinum:
Ah no I just forgot that `PlaceOnTop` handled turf creation.

PR #251, opened 2020-01-11
Make AI ships suck less

### LINKED ISSUE #176: The AI on ships is a complete mess
AI pilots are bad at flying

Kmc2000:
## Why It's Good For The Game

God DAMNIT THE AI SHIPS ARE BORING TO FIGHt.

This PR alleviates this.
AI ships now come in a few different flavours:

https://i.gyazo.com/3896482c13b843edc2df4560058c4add.mp4

"Fighters" will aim to rush at your ship and pepper it with bullets. They do not know the meaning of mercy, surrender, or retreat. 
"Light cruisers" will, like fighters, try to ram into the ship and enter CQC range.
"Carriers" are timid, once they have launched their fighters, they will run away from aggressors, and favour long range bombardment and survivability. If you do not take them out quickly, they will continue to launch more and more fighters
"Heavy cruisers" like to stay at medium range and use their railguns. If something gets into their PDC range, theyll try to back off a bit.

All AI ships above fighter class will now try to flee if their health gets too low, however this doesn't mean they won't stop shooting at you.

https://i.gyazo.com/84d3107ecb786156d804b43414b7610e.mp4

HEY ASSHOLE CHECK FOR ISSUES
fixes #176 
HEY ASSHOLE CHECK FOR ISSUES
## Changelog
:cl: Kmc2000 with HAL 9000
add: Added basic behaviours to AI ships to make them more interesting to fight.
add: A new AI ship class, "Carrier" which will pepper you with small fighters.
/:cl:


nsv13/code/modules/overmap/ai.dm
```diff
@@ -24,34 +26,71 @@
...
-						target(ship)
-				if(AI_RETALIATE)
-					if(ship in enemies)
-						target(ship)
+	handle_ai_behaviour()
+
+/obj/structure/overmap/proc/handle_ai_behaviour()
+	if(!ai_controlled)
+		return
+	for(var/X in GLOB.overmap_objects)
```
Jalleo: var/obj/structure/overmap/X
That should remove the need for the if(!istype right?

nsv13/code/modules/overmap/ai.dm
```diff
@@ -24,34 +26,71 @@
...
-				if(AI_GUARD)
-					if(get_dist(ship,src) <= guard_range)
-						target(ship)
-				if(AI_RETALIATE)
-					if(ship in enemies)
-						target(ship)
+	handle_ai_behaviour()
+
+/obj/structure/overmap/proc/handle_ai_behaviour()
+	if(!ai_controlled)
```
Jalleo: This already returns in proc its called from at line 18 of this file.

nsv13/code/modules/overmap/ai.dm
```diff
@@ -24,34 +26,71 @@
...
+					target(ship)
+		if(locate(ship in enemies))
+			if(get_dist(ship, src) <= 3)
+				user_thrust_dir = 0 //Don't thrust towards ships we're already close to.
+				brakes = TRUE
+			else
+				user_thrust_dir = 1
+				brakes = FALSE
+
+/obj/structure/overmap/syndicate/ai/carrier/handle_ai_behaviour()
```
Jalleo: This is a bad method to do this work out which areas of the proc you can  convert into individual procs to override at that level instead. You dont need to have seperate procs (A even better method is using signals to call the proper behaviour ask @ninjanomnom )

ninjanomnom: You could register on any overmap object that gets added to know when they move rather than using a processing proc to check every tick when they are close

Plotron:
Syndicate fighters are either too strong, or our own fighters are too weak. A Syndicate carrier usually spells doom for the NT crew. Enemy fighters respawn, they are very tanky, and are hard to shoot down.

Maybe there should be a limit on how many fighters the carrier is able to spawn (if there isn't one already), but I believe that we have a balance issue at hand. Our own fighters get shot down constantly in an open engagement. My own guess is that Syndicate fighters should be weak since they can be spammed by the enemy. The optimal strategy would be to rush the enemy carrier, of course, but DRADIS is kinda unhelpful in locating it when you've got so many simultaneous enemy blips on the radar.

nsv13/code/modules/mining/asteroid.dm
```diff
@@ -235,7 +235,10 @@ GLOBAL_LIST_EMPTY(overmap_asteroids) //For dradis to be able to track asteroids.
 	linked.relay('nsv13/sound/effects/ship/tractor.ogg', "<span class='warning'>DANGER: Magnet has locked on to an asteroid. Vacate the asteroid cage immediately.</span>")
 	cooldown = TRUE
 	addtimer(VARSET_CALLBACK(src, cooldown, FALSE), 1 MINUTES)
-	current_asteroid = new /datum/map_template/asteroid(pick("_maps/templates/nsv13/asteroid1.dmm","_maps/templates/nsv13/asteroid2.dmm","_maps/templates/nsv13/asteroid3.dmm"), null, FALSE, AS.core_composition) //Set up an asteroid
+	if(prob(10)) //I hate this but it works so fuck you
```
Jalleo: Put this in another PR instead so it can get merged faster fool!

nsv13/code/modules/overmap/weapons.dm
```diff
@@ -195,6 +187,11 @@
 			relay('nsv13/sound/effects/ship/reload.ogg')
 			swap_to(FIRE_MODE_TORPEDO)
 
+/obj/structure/overmap/proc/get_max_firemode()
```
Jalleo: signals can probably manage stuff like this better.

nsv13/code/modules/overmap/weapons.dm
```diff
@@ -253,6 +250,8 @@
 /obj/structure/overmap/proc/fire_torpedo(atom/target)
 	if(!linked_areas.len && !main_overmap) //AI ships don't have interiors
 		if(torpedoes <= 0)
+			if(ai_controlled)
+				addtimer(VARSET_CALLBACK(src, torpedoes, initial(src.torpedoes)), 60 SECONDS)
```
Jalleo: A hacky way to fix this. You may want to rework the entire code with all of this later on so its more modular and easy to use/expand.

nsv13/code/modules/overmap/weapons.dm
```diff
@@ -79,22 +74,21 @@
...
 				if(prob(pdc_miss_chance)) //Gives them a chance to actually hit a torpedo, so it's not a perfect smokescreen.
-					var/turf/T = get_turf(pick(orange(6,target))) //Pick a random tile within 6 turfs, this isn't a flat out miss 100% of the time though
+					var/turf/T = get_turf(pick(orange(4,target))) //Pick a random tile within 6 turfs, this isn't a flat out miss 100% of the time though
 					fire_pdcs(T, lateral=TRUE)
 				else
 					fire_pdcs(target, lateral=TRUE)
 	if(!found_target) //Can't see a torpedo to shoot, try find an enemy ship to shoot
 		for(var/obj/structure/overmap/ship in GLOB.overmap_objects)
-			if(!istype(ship, /obj/structure/overmap))
+			if(!ship || !istype(ship, /obj/structure/overmap))
```
Jalleo: null ship in overmap_objects? Also not merging the if below with this one?

Jalleo: A better method to manage GLOB.overmap_objects might be separating it into factions or something. That would let you to spawn in meteors as flying objects which are neutral to avoid.
You can then more easily manage it in separate code that signals use to manage for the ships instead as they are needed.

nsv13/code/modules/overmap/weapons.dm
Kmc2000: I'd prefer to run 1 simple compare operation than have to create and manage signal objects in memory for this imho

nsv13/code/modules/overmap/weapons.dm
Kmc2000: noted

nsv13/code/modules/overmap/ai.dm
```diff
@@ -69,4 +122,14 @@
 	if(OM.tactical)
 		var/sound = pick('nsv13/sound/effects/computer/alarm.ogg','nsv13/sound/effects/computer/alarm_3.ogg','nsv13/sound/effects/computer/alarm_4.ogg')
 		var/message = "<span class='warning'>DANGER: [src] is now targeting [OM].</span>"
-		OM.tactical.relay_sound(sound, message)
\ No newline at end of file
+		OM.tactical.relay_sound(sound, message)
+	try_launch_fighters()
+
+/obj/structure/overmap/proc/try_launch_fighters()
+	if(ai_can_launch_fighters) //Found a new enemy? Launch the CAP.
```
Jalleo: move this check into the /obj/structure/overmap/syndicate/ai/carrier/ai_target( and create a child of the target proc above for the carrier which literally adds the try_launch_fighters after inheritance with the check too. 

The reasoning is that each time its processing the AI it will go into this proc for the carrier when it doesnt need to. 
(wastes processing time)
Nothing else after that calls this proc. Your safe to remove the check inside the proc.

nsv13/code/modules/overmap/ai.dm
```diff
@@ -24,34 +33,78 @@
...
+/obj/structure/overmap/proc/retreat()
+	if(!last_target)
+		return
+	desired_angle = -Get_Angle(src, last_target) //Turn the opposite direction and run.
+	user_thrust_dir = 1
 
 /obj/structure/overmap/proc/target(atom/target)
 	if(!istype(target, /obj/structure/overmap)) //Don't know why it wouldn't be..but yeah
 		return
+	if(obj_integrity <= max_integrity / 3)
```
Jalleo: isnt * 0.33 faster than / still or is that out of date?

nsv13/code/modules/overmap/ai.dm
```diff
@@ -24,34 +33,78 @@
...
+			retreat()
+		else
+			target(ship)
+			try_launch_fighters()
+
+/obj/structure/overmap/proc/retreat()
+	if(!last_target)
+		return
+	desired_angle = -Get_Angle(src, last_target) //Turn the opposite direction and run.
+	user_thrust_dir = 1
```
Jalleo: is this var always a 0 or 1 ? If so please do a TRUE and FALSE across the files.

nsv13/code/modules/overmap/ai.dm
```diff
@@ -24,38 +33,91 @@
...
 	if(!istype(target, /obj/structure/overmap)) //Don't know why it wouldn't be..but yeah
 		return
+	if(obj_integrity <= max_integrity / 3)
+		retreat()
+		return
 	add_enemy(target)
 	desired_angle = Get_Angle(src, target)
 	last_target = target
 
+/obj/structure/overmap/target(atom/target)
```
Jalleo: ?!?!?

nsv13/code/modules/overmap/ai.dm
```diff
@@ -24,38 +33,91 @@
...
+			user_thrust_dir = 1
+			brakes = FALSE
+
+/**
+*
+* Carriers are more timid, and prefer to run away from enemy ships before firing, engaging at extreme range.
+*
+*/
+
+/obj/structure/overmap/syndicate/ai/carrier/ai_target(obj/structure/overmap/ship)
```
Jalleo: So it doesnt move towards the target?

PR #281, opened 2020-01-14
Restores the older station maps

Jalleo:
So its easier for quickly updating them from bee and for actually having maps to quickly reference what we are missing from our maps (Looking is faster and its a easy copy paste)

## Why It's Good For The Game
Mappers have a easier time and I restored them locally for myself so why not put them on the repo itself.
We dont need to maintain them just copy the upstreams as new stuff happens to see what they have and how its laid out to show us.


Jalleo:
They arent on the repo maps.txt so it should be fine. They also havent been given jsons. Only meta's exist because meta was already about just not compiling it should be disabled or not on the txt already.

PR #299, opened 2020-01-26
Bump spacemandmm to 1.2

spookydonut:
https://github.com/tgstation/tgstation/pull/48785
https://github.com/tgstation/TerraGov-Marine-Corps/pull/3647

### DIFF
--- a/code/__HELPERS/game.dm
+++ b/code/__HELPERS/game.dm
@@ -425,7 +425,7 @@
 	var/list/result = list()
 	for(var/m in group)
 		var/mob/M = m
-		if(!M.key || !M.client || (ignore_category && GLOB.poll_ignore[ignore_category] && M.ckey in GLOB.poll_ignore[ignore_category]))
+		if(!M.key || !M.client || (ignore_category && GLOB.poll_ignore[ignore_category] && (M.ckey in GLOB.poll_ignore[ignore_category])))
 			continue
 		if(be_special_flag)
 			if(!(M.client.prefs) || !(be_special_flag in M.client.prefs.be_special))
@@ -561,7 +561,7 @@
 	for(var/turf/found_turf in turfs)
 		var/area/turf_area = get_area(found_turf)
 
-		// We check if both the turf is a floor, and that it's actually in the area. 
+		// We check if both the turf is a floor, and that it's actually in the area.
 		// We also want a location that's clear of any obstructions.
 		if (specific_area)
 			if (!istype(turf_area, specific_area))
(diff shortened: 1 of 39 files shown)


Jalleo:
Hello

code/__HELPERS/game.dm
```diff
@@ -425,7 +425,7 @@
 	var/list/result = list()
 	for(var/m in group)
 		var/mob/M = m
-		if(!M.key || !M.client || (ignore_category && GLOB.poll_ignore[ignore_category] && M.ckey in GLOB.poll_ignore[ignore_category]))
+		if(!M.key || !M.client || (ignore_category && GLOB.poll_ignore[ignore_category] && (M.ckey in GLOB.poll_ignore[ignore_category])))
```
Jalleo: What does this fix?

code/__HELPERS/mobs.dm
```diff
@@ -436,7 +436,7 @@ GLOBAL_LIST_EMPTY(species_list)
 			continue
 		if(M.stat != DEAD && !override)
 			continue
-		if(speaker_key && speaker_key in prefs.ignoring)
+		if(speaker_key && (speaker_key in prefs.ignoring))
```
Jalleo: I presume this is a fix too

code/datums/hud.dm
```diff
@@ -35,7 +35,7 @@ GLOBAL_LIST_INIT(huds, list(
 
 /datum/atom_hud
 	var/list/atom/hudatoms = list() //list of all atoms which display this hud
-	var/list/mob/hudusers = list() //list with all mobs who can see the hud
+	var/list/hudusers = list() //list with all mobs who can see the hud
```
Jalleo: ?

code/__HELPERS/game.dm
spookydonut: `x && y in z` doesn't work how you expect it to, dreamchecker will warn about it and this change disambiguates the operations being done.

code/datums/hud.dm
spookydonut: hudusers is an assoc list but the value isnt a /mob, only the assoc key is.
`var/list/type/foo` doesnt do anything different in the byond compiler than `var/list/foo`, the typepath after /list exists to tell dreamchecker what type exists in the list's values.
in this case dreamchecker thinks you are doing (/mob)++ below which is a bad use of operator++ since /mob doesnt define `mob/proc/operator++()`
In reality the operation is safely doing (int)++ since the values in the list are ints

code/__HELPERS/mobs.dm
spookydonut: Yes

PR #327, opened 2020-02-12
Weapons PR fixes - machine frame runtime, startup admin messages

covertcorvid:
## About The Pull Request
Different way of handling directional machine building that doesn't runtime overrides.
Removes the overmap linkage messages from startup.

## Changelog
:cl:
fix: fixes machine frame runtime
fix: silences weapon initialization error messages during startup
/:cl:

### DIFF
--- a/code/game/machinery/constructable_frame.dm
+++ b/code/game/machinery/constructable_frame.dm
@@ -180,7 +180,11 @@
 						break
 				if(component_check)
 					P.play_tool_sound(src)
-					var/obj/machinery/new_machine = new circuit.build_path(loc, ndir=dir) //NSV13 - added ndir argument
+					var/obj/machinery/new_machine = new circuit.build_path(loc)
+					//NSV13 - start directional machine handling
+					if(circuit.directional_build == TRUE)
+						new_machine.setDir(dir)
+					//NSV13 - end directional machine handling
 					new_machine.setAnchored(anchored)
 					new_machine.on_construction()
 					for(var/obj/O in new_machine.component_parts)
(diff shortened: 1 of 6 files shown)


code/game/objects/items/circuitboards/circuitboard.dm
```diff
@@ -14,6 +14,7 @@
 	w_class = WEIGHT_CLASS_SMALL
 	grind_results = list(/datum/reagent/silicon = 20)
 	var/build_path = null
+	var/directional_build = FALSE //NSV13 - do we want the machine to have the frame's dir?
```
Jalleo: You can technically define variables in another file with the same object. We might want to do this if we wantto make updates easier.

covertcorvid: From a developer standpoint I'm not a fan of split class definitions since it makes it harder to get a good overview of what you're working with.

Jalleo: Thats fine it may make merge conflicts more likely though.

code/game/machinery/constructable_frame.dm
```diff
@@ -180,7 +180,11 @@
 						break
 				if(component_check)
 					P.play_tool_sound(src)
-					var/obj/machinery/new_machine = new circuit.build_path(loc, ndir=dir) //NSV13 - added ndir argument
+					var/obj/machinery/new_machine = new circuit.build_path(loc)
+					//NSV13 - start directional machine handling
+					if(circuit.directional_build == TRUE)
+						new_machine.setDir(dir)
```
Kmc2000: Would it not make more sense to override the "on_construction()" method to do all this?

covertcorvid: Good call, fixed

PR #335, opened 2020-02-16
"Better" ranks

Vasily2013:
## About The Pull Request

Makes a decent rank structure where all crew are granted the ranks that their positions warrant.

## Why It's Good For The Game

Because everyone deserves to receive credit for the ranks that they warrant.

## Changelog
:cl:
add: new set of ranks
/:cl:


Kmc2000:
While this is hilarious. These ranks are so totally robotic that as a non militaristically-inclined person, I wouldn't be able to tell much from them.

covertcorvid:
actually, one request - make the filename not have spaces, use underscores or something instead

covertcorvid:
(it seems rubbish because it's a joke)

Vasily2013:
> While this is hilarious. These ranks are so totally robotic that as a non militaristically-inclined person, I wouldn't be able to tell much from them.

bigger number higher rank. Fairly simple with warrants.

Vasily2013:
> This is actually rubbish. I want something with effort. If you do not do that by tomorrow I will close this PR. Just having CW and WO whatever seems rubbish.

this is about 80% memes

francinum:
Good jape. It's entirely optional and not the default.

PR #337, opened 2020-02-17
Fix: use of overmap camera takes away mob's ability to move

covertcorvid:
## About The Pull Request
Following the ship vibe check the mob's relaymove wasn't reset after use of overmap cameras. This includes both bridge consoles and any seegson ship viewscreen. After stopping the use of the camera, the mob would be unable to move and use of movement keys would cause runtime errors.

## Why It's Good For The Game
Movement is good

## Changelog
:cl:
fix: mobs can move normally after exiting overmap camera view
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/camera.dm
+++ b/nsv13/code/modules/overmap/camera.dm
@@ -56,11 +56,8 @@
 	if(M.client)
 		M.client.check_view()
 	M.overmap_ship = null
-	var/mob/camera/aiEye/remote/overmap_observer/eyeobj = M.remote_control
-	if(eyeobj?.off_action)
-		qdel(eyeobj.off_action)
 	M.cancel_camera()
-	M.remote_control = null
+	QDEL_NULL(M.remote_control)
 	return TRUE
 
 /obj/structure/overmap/proc/CreateEye(mob/user)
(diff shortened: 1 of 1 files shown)


nsv13/code/modules/overmap/camera.dm
```diff
@@ -60,6 +60,7 @@
 	if(eyeobj?.off_action)
 		qdel(eyeobj.off_action)
```
francinum: This doesn't look right to me...

nsv13/code/modules/overmap/camera.dm
```diff
@@ -60,6 +60,7 @@
...
 	M.cancel_camera()
+	M.remote_control = null
```
Jalleo: This likely fixes the issue but can you explain to Fran why the code qdel's above?

covertcorvid:
With regards to the implicit null, qdel waits 2 minutes before deleting to see if garbage collection takes care of it. https://github.com/Baystation12/Baystation12/wiki/Guide-to-Garbage-Collection
Not sure why it qdels, not my code, will try to analyze it more for you later

covertcorvid:
@francinum @Jalleo Looking at this further, eyeobj itself wasn't being deleted, off_action was, so even if deletion was immediate, which it's not, relaymove's reference still wouldn't be null because eyeobj would still exist.
Since off_action maintains an internal reference to eyeobj, even without relaymove pointing to eyeobj, neither of these would ever be GC'd if you don't either free one of the references manually or qdel one of them.

PR #353, opened 2020-03-01
Fixing more Enterprise debt

Cenrus:
- Adds intercoms and fire alarms to areas that didn't have them
- Fire extinguishers EVERYWHERE.
- Adds missing pilot RC console
- AI/Evac displays around the map.
- Maintenance uses the correct area types (starboard and port types)
- Pilots no longer have to pass through maintenance to reach the flight deck ladder.

image

- Fixes xenobio shield generators so that monsters can't smash them within the secure containment room
- Fixes xenobiology disposals leading to cargo instead of space
- Medbay+brig entrance firelocks
- Fixed a few instances of cameras on light tubes
- Air alarms in areas that needed them, such as the engine room and the engineering hangar.
- Renamed bridge RC to CIC RC.

### DIFF
--- a/_maps/map_files/Enterprise/Enterprise1.dmm
+++ b/_maps/map_files/Enterprise/Enterprise1.dmm
@@ -329,7 +329,6 @@
 /turf/open/floor/monotile,
 /area/crew_quarters/heads/captain)
 "aX" = (
-/obj/machinery/camera/autoname,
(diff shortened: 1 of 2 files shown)


Vasily2013:
I assume you've checked to make sure those doors do not start bolted?

Cenrus:
The ones leading to the flight deck ladder don't start bolted, and use the correct access types.

PR #375, opened 2020-03-04
Security level lowers to blue after dispatching all targets in a system

Cenrus:
## About this PR

Updates the Blue downto text and automatically sets the security level to blue once all the targets in a star system have been destroyed.

## Why it's good for the game

The red lighting can be annoying, and if there's no Captain with access to change it to blue using the bridge console, you're stuck with it.

## Changelog

:cl:
tweak: Ship security level now automatically drops to code blue after all syndicate targets have been dispatched.
/:cl:


francinum:
The current security levels are probably not useful on a combat ship actively patrolling, and should probably be reconsidered.

Karmic-Skink:
Sounds like a solution to an IC issue.

francinum:
Besides possibly a check to see if they've already dropped (or elevated beyond) red, this seems good.

PR #376, opened 2020-03-07
Port FTLiver

NIXC:
## About The Pull Request

Ports Faster Than Liver from the FTL13 code base. Made it less janky and easier to make than it used to be.

image
(It's animated, but I'm too lazy to make a gif)

## Why It's Good For The Game

It's alcohol, that means its automatically good.
## Changelog
:cl:
add: Faster-Than-Liver, the drink you probably didn't even realize existed in FTL13
/:cl:


code/modules/food_and_drinks/drinks/drinks.dm
```diff
@@ -264,6 +264,13 @@
 	list_reagents = list(/datum/reagent/consumable/ethanol/beer = 30)
 	foodtype = GRAIN | ALCOHOL
 
+/obj/item/reagent_containers/food/drinks/ftliver
```
Karmic-Skink: Please move all of these to files within the NSV13 folder for better compatibility.

NIXC: Two questions for you: Do I simply create a copy of the exact file path inside the NSV13 folder (eg nsv13/modules/food_and_drinks/drinks/drinks.dm, which doesn't exist currently) with the added code and nothing else in it?

And should I remake this PR from a clean branch to ensure that my commits have not touched the upstream files?

francinum: Moving them out should *probably* be fine, and yes, make a path-identical blank file with your new code in it.

francinum:
This could probably use a new sprite.

NIXC:
Here is the sprite as a gif, it looks slightly better in motion:
image
I have to leave the icon for it inside the normal drinks dmi, as drinking glasses assume the icon will be in that file and are unable to be configured otherwise.

PR #397, opened 2020-03-16
Fixes some issues with the MAA

Cenrus:
## About The Pull Request

Fixes the MAA's PDA.
The MAA's headset now has access to loudmode
The master at arms now has their own jumpsuit with suit sensors.

## Why It's Good For The Game

The munitions channel is usually very busy, so loudmode is good for coordination.

## Changelog

:cl:
add: The MAA now has access to loudmode.
fix: You no longer need to swipe your ID on the MAA's PDA.
add: the MAA now has their own jumpsuit.
/:cl:


code/modules/clothing/under/syndicate.dm
```diff
@@ -14,6 +14,7 @@
 	icon_state = "tactifool"
 	item_state = "bl_suit"
 	item_color = "tactifool"
+	has_sensor = HAS_SENSORS //NSV13 - MAA suit sensors
```
Karmic-Skink: This is a syndicate item and is only being used by the MAA as a place holder, please do not change the suit sensors on it.

Cenrus: This is a subtype of the turtlenecks worn by syndicate operatives, and can be purchased in ClothesMate vendors. I'll remove it for now

Karmic-Skink: See below.

Issue #416, opened 2020-03-23
Cargo impossible bounties


Cenrus:
Cargo can have bones, Goliath hides and other lavaland specific loot as bounties while we have no lavaland on the codebase.

Corsaka:
needs closure

PR #418, opened 2020-03-23
Resolve 416 "Cargo impossible bounties": Disable Mining Bounties

Tennessee116:
## About The Pull Request

These commits disable mining bounties, resolving #416. 
## Why It's Good For The Game

Cargo needs to be able to complete their bounties in order to get more bounties and therefore more station credits, affecting the outcome of the round.

## Changelog
:cl: Tennessee116
tweak: Disabled mining bounties
/:cl:


nsv13.dme
```diff
@@ -1568,7 +1568,7 @@
 #include "code\modules\cargo\bounties\item.dm"
 #include "code\modules\cargo\bounties\mech.dm"
 #include "code\modules\cargo\bounties\medical.dm"
-#include "code\modules\cargo\bounties\mining.dm"
+//#include "code\modules\cargo\bounties\mining.dm" //NSV13
```
francinum: This doesn't work. The compiler will automatically clear this line. Simply leave the file unticked.

code/modules/cargo/bounties/mining.dm
```diff
@@ -1,3 +1,4 @@
+/*
```
francinum: There's no need to comment out the entire file.

nsv13.dme
Tennessee116: so just leave #include "code\modules\cargo\bounties\mining.dm" be?

code/modules/cargo/bounties/mining.dm
Tennessee116: Alright. Would I just disable it from generating as a bounty in `bounty.dm` and leave everything else be?

PR #453, opened 2020-04-03
Oversight repair of a specific FUN verb.

Jalleo:
I caused a issue last round due to messing with it as it was turning to roundstart.

## Why It's Good For The Game

Prevents idiots like me pressing yes post roundstart as there wasn't a prevention check

I will push a PR of this to bee and TG tomorrow evening.


francinum:
This is going stale for no good reason.

PR #465, opened 2020-04-05
[Balance] SHIP FIGHTER, SHIP FIGHTER, WHERE WILL HE BOMB???

Kmc2000:
## About The Pull Request

Due to multiple issues, fighters are having their capabilities nerfed when shipside. 

## Why It's Good For The Game

Ship fighters are pretty unstoppable. Let's fix that.

## Changelog
:cl: Kmc2000
add: Master at arms now gets a console to better keep track of / control the fighters. Traitors can emag this to remove its access restriction.
del: Fighters can no longer fire their weapons when on the main ship without the MAA giving them firing authorisation via his console
add: Fighters are now much harder to fly in pressurized environments. If you don't keep your speed up, you scrape along the floor and damage your fighter.
add: I've remedied an issue causing fighters to be able to squeeze through 1 tile doors. They have also now lost their ability to open doors by default
add: Emagging a fighter lets you shoot its guns ANYWHERE you like. It also removes it from the MAA's fighter tracking console, at the cost of marking you as clearly hostile on DRADIS, and allowing any other fighter (Even other emagged ones, or Syndicate ones!) to fire on you, and vice versa.
add: Fighters now take a huge amount of damage from EMPs, specifically, the ION rifle is extra good at dealing with them now.
/:cl:


WhyIsCaeciliusTaken:
Why not give the console to ATC? They're the one that's currently supposed to be keeping track of what the fighters are doing, and having a console to actually do something about stolen/rogue fighters rather than just whining at MaA to force eject the pilot.

Karmic-Skink:
> Why not give the console to ATC? They're the one that's currently supposed to be keeping track of what the fighters are doing, and having a console to actually do something about stolen/rogue fighters rather than just whining at MaA to force eject the pilot.

Because the ATC is an RP role that is literally there to click the mag cat buttons.

Karmic-Skink:
Bring on the merge conflicts.

francinum:
Hammerhead doesn't really need the fighter tugs, as the hangar is too small to make use of them.
Likewise for Vago.

PR #479, opened 2020-04-10
Remove pointlessly excessive LRP spray

Corsaka:
## About The Pull Request
Removes the "yiff in hell furry" spray from the options in the spraycan

## Why It's Good For The Game
Promotes LRP, would be best removed. While in-game racism is allowed, having a pre-set spray for it I think would imply it's common, which would be LRP.

## Changelog
:cl:
del: Removed Yiff In Hell graffiti
/:cl:


francinum:
Tag as NSV Change.

Firefox13:
imo "spraycan option" for changelog is a poor description

PR #528, opened 2020-04-19
gives the detective on hammerhead, enterprise, and vago a sechud

BeloneX:
## About The Pull Request

replaces dets normal sunglasses with a sechud

## Why It's Good For The Game

cause it is

## Changelog
:cl:
add: gave the detective a sechud
/:cl:


GoatMonarch:
The detective isn't expected to do arrests though. Reason for someone to wear a sec-hud is to set someone to arrest / to locate people with a warrant, and arrest them.
So they do not really need security huds, and other sec members should keep in mind before they hand out sechuds like candy.

You also seem to have alot of pixel movement changes compared to the base maps you've edited on which is _od_.

Cdey78:
GoatMonarch has a point. Detective was not designed to arrest anybody -- That's why he was made at all. Giving him SecHuds is reinforcing a mindset that shouldn't have been there in the first place. Go play security officer if you get bored in the role of a detective.

Karmic-Skink:
I don't really believe detectives need a sec hud to function. If they really want one, they can ask the HoS for one.

PR #552, opened 2020-04-29
Increase nicotine OD threshold

Firefox13:
## About The Pull Request

Current nicotine OD threshold is 15, usual cigarette leaves you with 5u nicotine in system after smoking it, smoking premium or robusto cigars straight OD you. I raised OD level to ~~50~~ 35 units, smoking whole pack of cigs isn't dangerous now, but cigars can still OD you

## Why It's Good For The Game

muh smoker trait

## Changelog
:cl:
tweak: Raised nicotine OD threshold.
/:cl:

### DIFF
--- a/code/modules/reagents/chemistry/reagents/drug_reagents.dm
+++ b/code/modules/reagents/chemistry/reagents/drug_reagents.dm
@@ -41,7 +41,7 @@
 	addiction_threshold = 10
 	taste_description = "smoke"
 	trippy = FALSE
-	overdose_threshold=15
+	overdose_threshold=50 // usual cigarette leaves you with 5u of nicotine after smoking, Havanian cigs contain 15u, premium - 25u, cohiba robusto contain 40u, usual cigs contain 15-20u
 	metabolization_rate = 0.125 * REAGENTS_METABOLISM
 
 /datum/reagent/drug/nicotine/on_mob_life(mob/living/carbon/M)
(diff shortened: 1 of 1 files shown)


GoatMonarch:
Ever considered to smoke reasonably, or move to a more healthy alternative such as vape pens?

ta3370:
Whole pack should OD you, heck, half a pack should. Spacemans isn't spaced over the course of decades. I too use the quirk and you just learn how to balance it. Space out cigs, and with cigars RP it up a bit by taking it out to speak or the such, and it'll innately stop you from ODing by giving you regular pause.

Karmic-Skink:
While 35 is more reasonable then 50. I do not think this change is necessary and negates part of the quirk's penanlties, ie requiring to you manage your addiction.

Firefox13:
> 
> 
> While 35 is more reasonable then 50. I do not think this change is necessary and negates part of the quirk's penanlties, ie requiring to you manage your addiction.

The current problem is getting way too easily ODed by smoking cigarettes. We can probably negate raising OD level by decreasing metabolization rate(currently it's 0.125 per tick)

Karmic-Skink:
You can negate it by taking the cigarette/cigar out of your mouth.

Firefox13:
> 
> 
> You can negate it by taking the cigarette/cigar out of your mouth.

it's basically the same as "you can stop smoking not to get ODed", the OD threshold should get tweaked more to be in line with other drug reagents, 15 is too low, what do you say for 25-30 threshold then?

Cdey78:
"I dislike having a negative trait so I removed the negative of the trait."
Pay attention and take the cigarette out every so often instead of changing the game code for your own sake. There's a reason it's a negative trait.

PR #561, opened 2020-05-01
"Cw3040 Fixes KMC's technical debt" part 1

Cw3040:
Replaces all /obj/.../maintenance/defaultaccess with obj/.../maintenance{req_one_access_txt = "12"} because the former is bad practice and KMC should be ashamed of himself


Cw3040:
Closing to PR to Karmic's branch

PR #566, opened 2020-05-02
Makes the AI no longer able to connect to the CTF and syndicate radio channels with the intercoms

### LINKED ISSUE #253: Telecom code is a hot fucking mess [THANKS DOOHL]
AI sat intercom has access to red and blue CTF radio channels, can both hear and talk over them. It can also talk over syndicate channels but can't hear anything. It also has access to centcom channel which seems like it might of not been intended.

## Reproduction:

Be AI, click on intercom, turn the mic on, set channel to 121.3 (Syndicate, need a human with syndicate headset to hear it since AI doesn't have decryption key to actually hear it), 121.7 (Blue CTF), 121.5 (Red CTF), 133.7 (Centcom), and proceed to talk

Bokkiewokkie:
## About The Pull Request

Changes the minimum allowed frequency devices with 'freerange' can tune to as to avoid the AI (and the clock cult) tuning in to syndicate and CTF channels. Fixes #253, it does not stop the AI from tuning in to central command's radio. If that needs to be fixed aswell yell at me in the comments.
Also changes the atc radios to use a seperate superspace channel than the 'independent' one used by central command and CTF radio's.

## Why It's Good For The Game

Stops the AI from talking to people who are enjoying a friendly game of capture the flag which is highly illegal under space law.

## Changelog
:cl:
fix: the AI intercoms can now no longer change their frequency below 122.1 KHz
refactor: atc radio now uses a seperate superspace channel.
/:cl:

### DIFF
--- a/code/__DEFINES/radio.dm
+++ b/code/__DEFINES/radio.dm
@@ -58,8 +58,9 @@
 #define RADIO_CHANNEL_CTF_BLUE "Blue Team"
 
 
-#define MIN_FREE_FREQ 1201 // -------------------------------------------------
+#define MIN_FREE_FREQ 1221 // -------------------------------------------------
 // Frequencies are always odd numbers and range from 1201 to 1599.
+//(NSV13) Free frequency minimum was increased to 1221 to avoid the AI using comms to access syndicate and CTF channels
 
 #define FREQ_SYNDICATE 1213  //!  Nuke op comms frequency, dark brown
 #define FREQ_CTF_RED 1215  //!  CTF red team comms frequency, red
--- a/code/game/machinery/telecomms/broadcasting.dm
+++ b/code/game/machinery/telecomms/broadcasting.dm
@@ -161,9 +161,9 @@
 					radios += R
 
 		if (TRANSMISSION_SUPERSPACE)
-			// Only radios which are independent
+			// Only radios which are independent //NSV13 - or atc radios
 			for(var/obj/item/radio/R in GLOB.all_radios["[frequency]"])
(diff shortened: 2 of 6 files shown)


Bokkiewokkie:
I've tested it on a hosted server (while connecting with another machine) and my own pc so it should be fine.

Kmc2000:
ATC needs independant so that pilots can hear things across Zlevels but i'm thinking of just giving them a comms relay

Bokkiewokkie:
This makes atc independent from independent while still giving them cross-z-level comms.

MayhemSailor:
If ship comms go down ATC comms should also go down.

Bokkiewokkie:
ATC not having a comms server feels like it should be a seperate PR though and will probably be a bit more delayed since it'll include map changes.

francinum:
ATC physically can't be made to support telecoms equipment without equipping every fighter with a tcom relay. This is simply a limitation of telecomms jank code.

Bokkiewokkie:
Actually this can be done if you have the atc channel on a seperate independent channel, the only thing that would set it apart from the other comms at that point is that when the telecomms server dies the pilots can still talk to anyone who is on the overmap (other pilots) but not anyone one the ship. You could do this by making the server work kind of backwards from what it normally does, and making it stop ATC messages from reaching anyone's headsets when it's offline instead of actively carrying the signal over to the other headsets when it is online. (this would mean you don't need to attach a relay to every ship and can still have the comms break in a way that would be at least semi-logical)

Bokkiewokkie:
Closing this because telecomms is getting remade already

PR #569, opened 2020-05-02
Fixes Tyrosene fuel pump cargo crate

Cenrus:
Fixes tyrosene fuel pump crates, now they actually contain their contents.
Moved the tyrosene cargo pack datum to the correct folder. (/modules/cargo/packs)

:cl:
fix: Aviation fuel can be purchased from cargo
/:cl:


francinum:
I disagree with them coming empty, as that's unlike every other orderable reagent dispenser.

francinum:
This is also not the proper way to make reagent dispensers come empty, as setting `tank_volume = 0` causes it to be *unable to hold reagents at all*

Cenrus:
Good catch.

francinum:
@Kmc2000 I remember you wanting to hold off on merges, I'll let you push the button on this one.

Issue #587, opened 2020-05-10
Resisting from Gauss Guns before your view swaps forces you into overmap


Corsaka:
## Round ID:
316

## Testmerges:
irrelevant

## Reproduction:
1. Get into a gauss chair
2. Resist immediately to exit the chair
3. Wait for your camera to move
4. Try to move your character

PR #594, opened 2020-05-11
Various bugfixes and QOL

Kmc2000:
## About The Pull Request

I want die

## Why It's Good For The Game

fix https://github.com/BeeStation/NSV13/issues/587

TODO:
Stormdrive pass
Pass on weapons (briefly)

## Changelog
:cl:
add: Admins can now click on holopads while ghosted to holocall the crew.
fix: Fighter tugs are less broken
fix: Magcats are less broken
fix: Bridge consoles are replaceable
fix: Fighters can open pod doors with buttons
fix: Fighter occupants can now call elevators from inside a fighter
fix: Stormdrive reactor comp is now less snowflakey
fix: Hull plates have been made more fair and less buggy
fix: You can now build dradis consoles
/:cl:


nsv13/code/modules/overmap/fighters/tug.dm
```diff
@@ -87,9 +81,36 @@
 /obj/vehicle/sealed/car/realistic/fighter_tug/proc/can_launch_fighters()
 	return TRUE
 
+//This proc is cringe. Raw, unadulterated cringe. It'll do for now
```
covertcorvid, line 84: so what was wrong with the angle2dir proc in code/__HELPERS/type2type.dm ?

Kmc2000: youre kidding me, right?

Cenrus:
Helm/tactical console computer boards didn't show up in the science lathe

nsv13/code/modules/overmap/radar.dm
```diff
@@ -1,8 +1,22 @@
+/obj/item/circuitboard/computer/ship/dradis
```
Cenrus, line 1: if you're adding DRADIS circuit boards, can you add a way to replace the mining ship DRADIS console?

Jalleo: How is the mining ship DRADIS different?

nsv13/code/modules/overmap/radar.dm
```diff
@@ -1,8 +1,35 @@
...
+	name = "circuit board (dradis computer)"
+	build_path = /obj/machinery/computer/ship/dradis
+
+/datum/design/board/dradis_circuit
```
Jalleo, line 5: See how the freezer and heater board is done for the atmos machines and do the code needed to let the screwdriver switching between this one and the child mining one.

nsv13/code/modules/vehicles/_vehicle.dm
```diff
@@ -177,7 +179,7 @@ MASSIVE THANKS TO MONSTER860 FOR HELP WITH THIS. HE EXPLAINED PHYSICS AND MATH T
 			if(occupants.len)
 				var/mob/M = return_drivers()[1]
 				var/client/C = M.client
-				if(C && C.keys_held["Alt"])
+				if(!M || C && C.keys_held["Alt"]) //No driver? Brake automatically. Otherwise check if driver wants to break.
```
Jalleo, line 182: does this need to process once nobody is in it and its not moving?
Also I believe it needs a extra set of ( ) somewhere in here.

ta3370:
I feel like the elevator option should be on the tugs instead of the fighters. And if they lose tugs they can just go back to current. Makes tugs more useful and less of a meme. Perhaps the same for pod doors since ATC is likely driving the tug, and as said in the previous sentence, makes tugs have more purpose for ease of use in comparison to just mavericking away the hull tiles in a fighter.

Kmc2000:
> I feel like the elevator option should be on the tugs instead of the fighters. And if they lose tugs they can just go back to current. Makes tugs more useful and less of a meme. Perhaps the same for pod doors since ATC is likely driving the tug, and as said in the previous sentence, makes tugs have more purpose for ease of use in comparison to just mavericking away the hull tiles in a fighter. Also encourages teamplay and coordination more instead of solo play and pilot toxicity since "I don't rely on nobody else" like pilots getting the button in their fighter would.

The button is almost entirely so that they can open the doors from space. Fighters flying without tugs indoors is totally unviable,  tugs will see more use now that I've fixed them up to not be terrible in every way.

PR #628, opened 2020-05-26
Feature freeze

Jalleo:
Time for a update cycle

##What does this mean?
We are going to be updating the repo to be in line with the newest possible beestation version to add a few extra features and give us some performance boosts.

## What features will this give us?
TGUI 3
Monstermos
And more.

## Why now?
We have been meaning to do it for a while but kept holding it off.

## Rules for the freeze
Reworks of existing NSV13 specific code is allowed. Addition of sprites for NSV13 specific files are also allowed. PRs that touch non NSV13 subfolder files need to state UPDATE CONFLICT in their PR title until we tag it properly.

#Afterword
Sorry for any inconvenience we do have ideas of how to do this more often with less conflicts by some ideas that we have heard of but at this time we haven't fully investigated it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@
 **Code:** https://github.com/Beestation/NSV13
 **Wiki:** https://nsv.beestation13.com/wiki/Main_Page
 
+## Kmc is a coding torp
+Nothing can stop his coding ability it sinks boring mechanics with ease to bring more enjoyment to all.
 
 ## DOWNLOADING
 


WhyIsCaeciliusTaken:
Won't you need to remap all the maps for monstermos?

Jalleo:
We will we are first updating the code in another branch and then remapping the maps in that branch.

PR #632, opened 2020-05-28
Adds my changes back to the Galactica

Cenrus:
restores my changes that were lost when 610 was merged, and includes the EVA Storage + CIC security upgrade that were introduced in said PR.

Shouldn't interfere with the upstream merge because it's mapping related.


Kmc2000:
Does this override sergei's changes? or is this a clean re-implementation

Cenrus:
Should be clean, I checked to ensure all the issues were fixed, added the CIC upgrade and EVA storage room

Cenrus:
Issues have been fixed, I didn't remove the rack as it's not a dense object, and you can reach the cabinet even if it's there or not.

Vasily2013:
Remove this disco ball too please.

Also make sure the surgery doors get changed to ID access 5. The sleepers should be set in another room if this is not done.

Cenrus:
Disco ball removed, there seem to be sleepers in the Cryogenics room which uses access type 5, so I'll keep the surgery access on 45 for now.

PR #640, opened 2020-05-31
Fixes Torp Casing bug when constructing in a trolley, but the root cause needs work.

### LINKED ISSUE #620: A torpedo can get stuck in a Trolley

hmkwoc:
## About The Pull Request

- When you construct torpedoes which are contained into a trolley, they may break the trolley.

- The original author of the system probably wanted to allow for "dud" torpedoes. Problem is, the munitions trolleys do a simple check for ammunition, which any torpedo casing unfortunately is. In other words, thats a typeing problem too. (IMHO, they should be -really- parts of a torpedo assembly, not ammunition).

- On the other hand, Munitions trolleys also don't care much about what you are trying to do. If the object contained is trying to get deleted, the capacity to load more ammunition is not updated, which means that the trolley will be bugged forever in this state (root cause). This may render the trolley effectively useless because it will act as if it was already fully loaded when, in fact, it is not.

- This pull request tries to keep the original author's code unchanged, but does a simple check to see if we are trying to mess with the loaded ammo in the trolley.

## Why It's Good For The Game
It solves one "priority bug", at least partially and without having to make a major change.

It is really frustrating and immersion breaking when you either cant use the trolley (if you managed to bug the entire capacity) or seeing icons flying around where they shouldnt be. It was my second round on the server when it happened to me (note, I wasnt the one who bugged it) and I was really trying to do my best to keep the hammerhead railgun loaded not knowing that I was doomed anyway (new player perspective, I know its shocking). Picture related: https://prnt.sc/sr773l

## Changelog
:cl:foehammer
fix: Munitions Carts shouldn't have ghost torps anymore.
/:cl:
Fixes #620

### DIFF
--- a/nsv13/code/modules/munitions/ammunition/torpedos/torpedo_construction.dm
+++ b/nsv13/code/modules/munitions/ammunition/torpedos/torpedo_construction.dm
@@ -191,6 +191,10 @@
 	. = ..()
 
 /obj/item/ship_weapon/ammunition/torpedo/torpedo_casing/welder_act(mob/user, obj/item/tool)
+	//Fixes #620 (https://github.com/BeeStation/NSV13/issues/620)
+	if(istype(src.loc, /obj/structure))
+		to_chat(user, "<span class='notice'>[src] is loaded in [src.loc]. Unload it first.</span>")
+		return
 	switch(state)
 		if(0)
 			to_chat(user, "<span class='notice'>You start disassembling [src]...</span>")


Jalleo:
Thank you for making this PR we shall review it as a priority.

francinum:
Please do not include unused changelog headers. It makes the bot confused.

hmkwoc:
> Please do not include unused changelog headers. It makes the bot confused.

About that, I was not sure what to do with them. Never worked with SS13 code before. I will keep it in mind if I contribute again in the future.

> This is about as sane as a sanity check this is gonna get. I'm not sure of loc actually is the correct check here. Has this been tested ingame?

It was tested in my own test environment. Thats as in game as I can go.

Jalleo:
Just so you are aware @hmkwoc you need the Fixes #whatever in the text area not the title. Thank you very much for this PR and fixing this.

PR #651, opened 2020-06-05
Update 0.6 Map paths fixed. All recent updates to master is merged too.

Jalleo:
If your working on a map and this causes a issue with your map tell me and I will remove your map before merging this.


Jalleo:
I dont know why Galatica is so large. WAS IT NOT TGM?

PR #656, opened 2020-06-09
Adds a lazy version of turbolifts (sorry Ike)

Kmc2000:
(Elevators should be PUCE.)

## About The Pull Request

I am lazy and I like being lazy

## Why It's Good For The Game

Adds lazylifts to the game. 
## Changelog
:cl:
add: Added voice activated, lazy versions of turbolifts
/:cl:


Firefox13:
Can we get some flavor: When you press the button on harm intent you get differen, red, angry press text.

nsv13/game/lazylift.dm
```diff
@@ -0,0 +1,394 @@
...
+	name = "turbolift door"
+	desc = "A bulkhead which opens and closes."
+	resistance_flags = INDESTRUCTIBLE | LAVA_PROOF | FIRE_PROOF | UNACIDABLE | ACID_PROOF | FREEZE_PROOF //Go away greytiders
+
+//Lifts speak a lotta languages, alien isn't one of 'em.
+/datum/language_holder/synthetic/turbolift
+	languages = list(/datum/language/common)
+	shadow_languages = list(/datum/language/common, /datum/language/machine, /datum/language/draconic, /datum/language/drone)
+
+/obj/machinery/lazylift
```
Jalleo, line 85: EW override what you need to on the bee version by being a child there.

nsv13/game/lazylift.dm
```diff
@@ -0,0 +1,394 @@
...
+/obj/machinery/lazylift_button/attack_hand(mob/user)
+	if (stat & NOPOWER || !lift)
+		to_chat(user, "<span class='warning'>Nothing happens...</span>")
+		return
+	if(isliving(user) && user.mouse_opacity) //Don't display things like AIs or admin ghosts clicking the button.
+		visible_message("<span class='notice'>[user] [pick("calmly presses", "mashes", "pokes", "slaps", "hits")] [src] </span>", "<span class='notice'>You push [src].</span>")
+	flick("button_lit", src)
+	lift.master.path_to(lift.deck, lift.master.platform_location.deck)
+
+/obj/machinery/door/airlock/ship/public/glass/turbolift
```
Jalleo, line 75: Okay thats NOT a suitable child there. Flat out no make it a child of ship not ship/public/glass/turbolift.
That makes life hell for mappers

nsv13/game/lazylift.dm
```diff
@@ -0,0 +1,394 @@
...
+/obj/machinery/lazylift/attack_hand(mob/user)
+	var/list/deckNums = list()
+	for(var/obj/machinery/lazylift/LL in master.decks)
+		deckNums += LL.deck
+	var/theDeck = input(user, "Which deck would you like to go to?", name, null) as null|anything in deckNums
+	if(!theDeck || !isnum(theDeck))
+		return FALSE
+	master.path_to(theDeck)
+
+/obj/machinery/lazylift/Hear(message, atom/movable/speaker, message_language, raw_message, radio_freq, list/spans, message_mode)
```
Jalleo, line 127: Have you thought about using signals and the voice analyser object for this to a degree?

nsv13/game/lazylift.dm
```diff
@@ -0,0 +1,394 @@
...
+/obj/machinery/lazylift/master/Initialize()
+	. = ..()
+	master = src
+	var/turf/T = get_turf(src)
+	turf_type = T.type
+	//Time to set up the wacky turbolift network!
+	platform_location = src //IMPORTANT! Platform starts at the bottom lift.
+	var/turf/last = get_turf(src)
+	decks += src //Ew. But necessary.
+	for(var/I = 0; I <= world.maxz; I++)
```
Jalleo, line 212: Have you checked to see if you can use a version of get z's of this type? Aka station_z's. This seems like a slightly inefficient method personally.

nsv13/game/lazylift.dm
```diff
@@ -0,0 +1,394 @@
...
+					to_chat(karmics_victim, "<span class='warning'>You hit your head as you're thrown about wildly!</span>")
+					karmics_victim.adjustBrainLoss(10)
+					playsound(karmics_victim.loc, 'nsv13/sound/effects/lift/bonk.ogg', 100)
+				if(HAS_TRAIT(karmics_victim, TRAIT_SEASICK)) //oh my god you poor soul
+					to_chat(karmics_victim, "<span class='warning'>You can feel your head start to swim...</span>")
+					if(prob(40))
+						karmics_victim.adjust_disgust(20)
+					else
+						karmics_victim.adjust_disgust(10)
+		sleep(1.4 SECONDS) //Sound time!
```
Jalleo, line 361: This is potentially a unsuitable sleep here have you thought about using a callback instead?

Jalleo:
Conflicts with the update_branch. It would of been better if you PR'd it directly to that branch.

Karmic-Skink:
Any chance we can have it pick a from a selection of elevatormusic.oggs so we can "start" the music at different sections of the track? Using the same initial note every time sounds strange.
ie have multiple cuts of the same track at different time segments.
Words hard.

nsv13/game/lazylift.dm
Kmc2000: Yeah unfortunately these two systems are totally incompatible, so extension wouldn't help here.

nsv13/game/lazylift.dm
Kmc2000: This was to inherit the icons n' stuff but sure

nsv13/game/lazylift.dm
Kmc2000: So this is actually from voice analyser code, using signals would make it more complicated as I'd have to register a speech signal on any mob that walks in to the elevator

nsv13/game/lazylift.dm
Kmc2000: To be honest this isn't that bad. It terminates as soon as it stops finding relevent turbolift bits. In terms of CPU operations this is:
get value of world.maxz
increment a singular number until it reaches this value, or we stop finding openspace turfs

nsv13/game/lazylift.dm
Kmc2000: Potentially, but sleeps are fine in this instance, as I've heard from multiple people. As long as they're not excessively long, they're on par with timers. This is the intended use case for them (short, animation related breaks). If I were to use timers, it'd also increase proc call overhead.

nsv13/game/lazylift.dm
Jalleo: That makes sense. We might want to make a Hear component then.

Jalleo:
Or you could of just PR'd it to the update_branch. Then you would of resolved your conflicts early.

PR #668, opened 2020-06-16
Hammerhead Monstermos Compliance part 2

Vasily2013:
This PR brings hammerhead up to full monstermos compliance, and establishes several beneficial mapping QOL changes such as window spawners spawning window firelocks, and directional variants of airlock controllers.


Jalleo:
We can find the issues and fix them with the actual tests.

PR #673, opened 2020-06-19
Window spawners

Cenrus:
adds seperate window spawners with window firelocks. No NSV tags because they're merged from upstream.


Vasily2013:
This PR is redundant because it has already been done to our spawners in NSVCode.

Cenrus:
If all windows need firelocks then this PR is redundant

Issue #665, opened 2020-06-15
Science has no experimental hardsuit, just a normal EVA suit.


DrSirKnight:

Cenrus:
Can be closed now

theboy6545:
@Vasily2013 can you close this? @Cenrus said its been fixed.

Vasily2013:
99% sure I fixed this.

Issue #652, opened 2020-06-06
XO Waypoints are fucked.


Corsaka:
## Round ID:
429, 430

## Testmerges:
Possibly: 607, 646

## Reproduction:
XO spawns in random locations duing roundstart, the spawning logic is incorrect it seems

francinum:
This is a serial issue related to the bullshit we did to the HoP job.

Issue #650, opened 2020-06-05
shield perma from radiation, at least a little bit.


Vasily2013:
title

francinum:
Sounds like security's job, really. Counts as 'Dangerous Environment'

Corsaka:
Who's gonna do that? No security officer would risk having the perma prisoner escape, which yes, IS their problem, but it's also unfair to the perma prisoners, who are EVEN MORE out of the round than they were previously.

francinum:
![](https://file.house/ERPT.png)

![](https://file.house/mR0z.png)

Issue #645, opened 2020-06-01
[Hammerhead] Some doors have wrong access settings


Bokkiewokkie:
## Round ID:
Not relevant
## Testmerges:
Not relevant
## Reproduction:

This door in Toxins has access that allows almost no one to open it (possibly set to require all accesses to open instead of only one)
![image](https://user-images.githubusercontent.com/43698041/83405549-02662480-a40d-11ea-9577-415b645b3f15.png)

This door near departures also never wants to open for anyone
![image](https://user-images.githubusercontent.com/43698041/83405733-77d1f500-a40d-11ea-9724-257988927248.png)

The arms dump shutter button is set to "one_access=0" so someone needs MaA, HoS and armory access to open it. (probably the same issue as the other two)
![image](https://user-images.githubusercontent.com/43698041/83406047-19594680-a40e-11ea-8e0c-7f1dac35cf24.png)

This toxins airlock has no access for scientists even though that's the whole reason it and the eva suit are there.
![image](https://user-images.githubusercontent.com/43698041/83407405-f5e3cb00-a410-11ea-883a-e95ff0f987b7.png)

PR #677, opened 2020-06-20
Fixes some map issues that cropped up during testing

Cenrus:
Fixes issues with the Enterprise and the Hammerhead.

Enterprise Tcomms has received a security upgrade too.

:cl:
tweak: Enterprise tcomms has received a security upgrade
/:cl:


Vasily2013:
While we're still at it by the by, please review the following issues and verify if they are either still issues or if they are resolved. If they are issues I would appreciate if you could clean them up while we're at it.

#683 
#665 
#652 
#650 
#645 
#636 
#641 
#598

PR #685, opened 2020-06-24
Fixes the wall related runtime errors during initialization

Crossedfall:
## About The Pull Request

When the reinforced walls were updated to have their `update_icon()` proc temporarily removed in https://github.com/BeeStation/NSV13/pull/629, the color value for the syndicate walls wasn't overridden to prevent runtimes. 

## Why It's Good For The Game

Less runtimes during startup. Also Travis should actually pass now when combined with my other PR #684 

## Changelog
:cl:
fix: Updates the syndicate r_wall's color value to be null. Resolving startup runtime errors
/:cl:


Crossedfall:
The ultimate solution here would be to re-implement the `update_icon()` proc in the file. But this should resolve the runtimes without actually changing anything; in other words, the walls still wont be smoothed but at least you wont have runtimes.

PR #708, opened 2020-07-04
Minor interstitial upstream pull

francinum:
Minor interstitial update.


francinum:
SQL stuff needs a check over, BSQL's removal still has a few pretty deep wounds.

Jalleo:
Remove the Do not Merge once you are a okay with it.

francinum:
After talking with Crossed, assuming that we have no custom SQL interactions (we don't), there should be no issues with it.

On the bright side we can use all of rust-g's stuff.

francinum:
We'll be here all fucking day if I wait for the map diff bot to wake up from it's eternal slumber

PR #711, opened 2020-07-06
Comments out wallstuns

Cenrus:
Rebalances tazers by removing wallstuns. People have complained about them and they cause bugs (Hulks/Xenos endlessly wallslamming).

:cl:
del: disabled wallstuns
/:cl:


francinum:
Define the bugs with hulks and such

PowerfulBacon:
Wallstuns removed entirely? Rest in peace space lube.

GoatMonarch:
> Define the bugs with hulks and such

I theorize that given hulks are stun immune, if they walk into a wall it'll just repeatedly cause them damage. Or it completely breaks the single cause as to why hulk is good, by stunning you anyways.

As for the PR, I am kind of on the mix. I think wallstuns from confusion are really dumb, but you ought to also re-enable simple stuns from sources like space lube (so it isn't a lethal and deadly tool like how it is currently, and actually returns to it being fun to use), and make the HoS' gun taser shot more than just a very short stun that you can't do anything with.

Kmc2000:
I think the core issue with this is confusion and this kinda just bandaids over the problem

PR #727, opened 2020-07-15
A potential fix to msay breaking

Corsaka:
## About The Pull Request
Attempts to fix msay. It only breaks when admins override OOC with a message, and it overrides with a `check_rights_for` proc. Including a third argument might just fuck it. I cannot think of any other reason why this would break.

## Why It's Good For The Game
Allows mentors to communicate once more

## Changelog
:cl:
fix: mentorsay, maybe.
/:cl:

### DIFF
--- a/code/modules/mentor/mentorsay.dm
+++ b/code/modules/mentor/mentorsay.dm
@@ -11,7 +11,7 @@
 
 	log_mentor("MSAY: [key_name(src)] : [msg]")
 	msg = keywords_lookup(msg)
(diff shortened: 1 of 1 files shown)


francinum:
No. That's not the problem.
The change you did just causes mentors to have their chat spammed by 'you don't have this right' messages on msay. The problem is that the mentor list isn't being properly cleaned and contains nulls that cause the proc to runtime.

francinum:
Resolved by #730

PR #770, opened 2020-07-26
Probably maybe fixes a possible memory leak with Aetherwhisp6

Kmc2000:
## About The Pull Request

I can't tell if this is it for SURE, but this is probably a far better way of doing things, anyways.

## Why It's Good For The Game

I really like whisp and want to be able to play it for longer than 2 seconds :))

## Changelog
:cl: Satan
fix: Elevators will now listen to you when you say "floor" as well as "deck". Rejoice.
/:cl:


Kmc2000:
I literally can't think of what else would be causing that cursed ship to crash so let's all hope this was it.

PR #771, opened 2020-07-26
Jeppison hangar mods

Kmc2000:
## About The Pull Request

Tapes a massive spoiler on the back of Jeppison's hangar to make it go faster

## Why It's Good For The Game

Probably voids your insurance idk

With this new setup. I've given the hangar a whole 2 tiles of extra space, but it now gives you suitable space to manage your flyboys. For the most part, jeppy's planes live out on the hull, where theyre nice and vulnerable to /RANDOM PDC BULLET EVENT/ but they can be brought into the hangar for repairs if you need, or you can just ignore it entirely and stick to the primary magpult.

ATC gets a neat camera view of the launch zone. We should do this on other maps COUGH COUGH

## Changelog
:cl: Kmc2000 + Not Satan
fix: Jeppison's hangar is usable now (don't hurt me sergei)
/:cl:

### DIFF
--- a/_maps/map_files/Jeppison/Jeppison1.dmm
+++ b/_maps/map_files/Jeppison/Jeppison1.dmm
@@ -178,15 +178,15 @@
 	dir = 8
 	},
 /obj/machinery/atmospherics/pipe/manifold/supply/hidden/layer1{
-	icon_state = "manifold-1";
-	dir = 8
+	dir = 8;
+	icon_state = "manifold-1"
 	},
 /turf/open/floor/circuit,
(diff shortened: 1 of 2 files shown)


Kmc2000:
> Please remove the hull plate, I want to keep it down to the two patches on the lower deck, and there's no easy external access point to that with basic engineering access. (See: Me not trusting pilots to let engineers use their airlock.)

Fair enough, it's 100% aesthetic anyway

Cenrus:
Looking swell but connect the airlock vent to the scrubbers or an exterior ejector port

PR #788, opened 2020-08-02
[Test Ready] Fire balance

Jalleo:
## About The Pull Request

This is a highly controversial pr that is working on removing the fireproof from most objects as they are unnecessary and likely redundant. I am going to investigate this system more thoroughly as to WHY this is added and then work towards making it suitable with however it should be.

## Why It's Good For The Game

Things that should get destroyed can get destroyed.

## Changelog
🆑 
balance: A lot more stuff gets destroyed via fires. Including NORMAL ID's, FIGHTERS and ALMOST ALL NON ANTAG ITEMS (Including the fireaxe)
/:cl:


Karmic-Skink:
That change to fighters won't actually change anything. Read line 30.

GoatMonarch:
So what this PR will change is:
Fire alarms will be destroyed by fires, so fixing a fire bomb is even more difficult than usual.
IDs will be burned in fires, very fun when the AI is burning the ship. Or when you need to head to the XO's office after a random fire anomaly.
Metal apparantly can burn into dust. Given crowbars, wrenches and other tools are successible.
Lets not forget the fun enjoyable part where backpacks will burn up, drop whatever important stuff on you, on the floor, for that to burn too.

_I dunno if I like this change, chief. Not everything is something you can recover by clicking a single vending machine some place._

WhyIsCaeciliusTaken:
Most things are fireproof for a reason. With fastmos making it really easy to flood large areas with burning gas, making it even more destructive is dumb.

Jalleo:
Okay the ID's and that are only for NORMAL. All other ID's are fireproof. As well by the fact that 90 armor on ids will not actually burn entirely unless its inside a TURBINE room on fire.

Vasily2013:
fire alarms should not burn, as they are meant to deal with high temperatures. I'd legitimately prefer it if the wall failed before fire alarms or APC's.

Jalleo:
@Vasily2013 that we would need to get fixed first.

PR #805, opened 2020-08-08
Removes most radiation spawners from maps

Vasily2013:
This PR removes most of the radiation spawners on all the currently active maps. We've discussed this at the highest levels, and agreed that we want a stormdrive meltdown to be less of a round-ending event. We think this is a good first step to do that.

🆑 
Remove: Most radiation spawners for meltdown events have been removed. The ship will be covered in a reduced amount of plutonium sludge.
balance: The above change should result in meltdowns being less of a round-ender.
Refactor: Atmospherics on Tycoon has been redesigned.
🆑


Jalleo (approved):
Can we force this in ASAP please @Kmc2000

PR #853, opened 2020-08-24
Resolves issues with bullets hitting each other and more

Kmc2000:
## About The Pull Request

Ok so yeah it turns out brain man not so good.

## Why It's Good For The Game

Resolves issues with bullets hitting each other and potentially causing crash inducing physics calc spam. Tries out a new way of deleting physics2d components to maybe make them a little easier to be rid of.

## Changelog
:cl:
fix: Fixes a few crashes
/:cl:


Jalleo (commented):
TEST FOR THE TEST GODS!

code/modules/projectiles/projectile.dm
```diff
@@ -234,7 +234,7 @@
 	if(!check_faction(A))
 		return FALSE 	 //Nsv13 - faction checking for overmaps. We're gonna just cut off real early and save some math if the IFF doesn't check out.
 
-	var/datum/point/pcache = trajectory.copy_to()
+	var/datum/point/pcache = trajectory?.copy_to()
```
covertcorvid, line 237: Will this error later in the function when other things try to use pcache?

Kmc2000: Probably but I don't want to think too hard about it

covertcorvid: IT'LL STILL RUNTIME THOUGH, BAD
Either make a default trajectory or return early, I think

Kmc2000: no please I don't want to work with byond

covertcorvid, line 237:
```suggestion
    if(!trajectory)
        var/turf/starting = get_turf(src)
        trajectory = new(starting.x, starting.y, starting.z, pixel_x, pixel_y, Angle, SSprojectiles.global_pixel_speed)
	var/datum/point/pcache = trajectory?.copy_to()
```

Jalleo: You can literally set it up with a semi colon or something to force a alternative instead I think.

Jalleo: Do that please.

Jalleo (changes requested):
It might be suitable to convert all of this to using the newer TG system of update overlays. Just to reduce the chance of things messing up badly.

Kmc2000:
Tl;dr I think bullets kept repeatedly spam colliding with each other and not moving, causing a shitload of expensive physics calc and whatever. The bullets hitting each other strikes me as a symptom rather than a cause.

Lemme know how this goes and if anything else is still fucked. Back to watching daytime tv for me.

nsv13/code/modules/overmap/physics2d.dm
```diff
@@ -74,15 +74,30 @@ PROCESSING_SUBSYSTEM_DEF(physics_processing)
...
 /datum/component/physics2d/proc/setup(list/hitbox, angle)
 	position = new /datum/vector2d(holder.x*32,holder.y*32)
 	collider2d = new /datum/shape(position, hitbox, angle) // -TORADIANS(src.angle-90)
 	last_registered_z = holder.z
 	START_PROCESSING(SSphysics_processing, src)
 
 /datum/component/physics2d/proc/update(x, y, angle)
-	collider2d.set_angle(angle) //Turn the box collider
-	collider2d._set(x, y)
+	collider2d?.set_angle(angle) //Turn the box collider
```
qwertyquerty, line 99: I believe two nullcheck sets is slower than one nullcheck and two sets

PR #868, opened 2020-08-31
Crew are now properly warned about boarders

Cenrus:
Fixes a span.

:cl:
fix: Crew are now warned when boarders spawn
/:cl:


GoatMonarch:
This is some Alien: Colonial Marines tier of code breaking from a typo.
I'd say it's a good idea that nobody has thought of, but it seems like *someone* already has.
HONK!

francinum:
Thanks KMC

PR #869, opened 2020-08-31
Adds overmap objects to the ghost orbit list

Cenrus:
Ghosts can now view and orbit all entities on the overmap.

## Why it's good for the game

Ghost spectator QOL

:cl:
add: Added overmap objects to the point of interest list.
/:cl:


francinum:
Note that overmap ships are not destroyed when a system is left.

You need to make sure that no ship gets placed in nullspace while still being a PoI

francinum:
Also this will add asteroids to the list.

That's completely unnecessary bloat.

PR #945, opened 2020-10-17
Tiny PR - Tiny Fans

Vexylius:
## About The Pull Request

This PR does two things:
-Replaces holofans on Pegasus with tiny fans
-Adds tiny fans to the CentCom ferry

## Why It's Good For The Game

Why would you put holofans on your map instead of tiny fans in the first place?
Also, CentCom ferry suffers from too many depressurisation accidents, when someone accidentally bumps into the doors, so tiny fans should fix that.

## Changelog
:cl:
add: Added tiny fans to CentCom ferry
del: Removed holofans from arrivals on Pegasus
add: Added tiny fans in place of those holofans in arrivals on Pegasus
/:cl:

### DIFF
--- a/_maps/map_files/Pegasus/pegasus2.dmm
+++ b/_maps/map_files/Pegasus/pegasus2.dmm
@@ -7502,7 +7502,7 @@
 	dir = 1
 	},
 /obj/effect/mapping_helpers/airlock/unres,
-/obj/structure/holosign/barrier/atmos,
+/obj/structure/fans/tiny,
 /turf/open/floor/plasteel/grid/mono,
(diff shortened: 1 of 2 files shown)


francinum:
The arrivals needs additional changes to make it viable, tiny fans aren't the solution here.

Vasily2013:
The reason why holofans are actually suitable for the Pegasus arrivals is it A: prevents stupidity and B: they're relocatable in round by players, so they're not a finite magic item that only exists at round start. Personally I think the better fix here would likely be figuring out how to implement a full airlock here.

PR #968, opened 2020-10-21
Allows teleportation on the mining ship

### LINKED ISSUE #457: Teleporting on the Nostromo (and possibly FOB) doesn't work (Missing Z-traits for mining)
## Round ID:
Today (technically 4-4-2020) at around 3:00 am (the round ended with a crash so idk the id)
## Testmerges:
none
## Information:
When you use an item which is supposed to teleport you while you are on the Nostromo (or maybe just the mining ship Z level in general) you are unable to use it. This looks like it works the same way as the anti-teleporting of the Central Command Z level.  
## Reproduction:
1 Go to the Nostromo
2 Activate an item that teleports you, for example: A bluespace crystal, Quantum spin inverter, bluespace pad, admantium armor, etc. (effect of conventional teleporting still unknown)
3 Wait for the normal time it would take you to teleport and realize you're in the exact same place without any teleporting having even occurred.

TheChaser212:
## About The Pull Request
Removes the noteleport variable from mining ship areas so that teleportation can be used.
Fixes #457

## Why It's Good For The Game
I don't see any reason why teleportation shouldn't be allowed

## Changelog
:cl:
fix: Teleportation can be used on the mining ship
/:cl:

### DIFF
--- a/nsv13/game/areas.dm
+++ b/nsv13/game/areas.dm
@@ -199,7 +199,6 @@
 /area/maintenance/nsv/mining_ship
 	has_gravity = TRUE
 	looping_ambience = 'nsv13/sound/ambience/maintenance.ogg'
-	noteleport = TRUE
 
 /area/maintenance/nsv/mining_ship/central
 	name = "Rocinante maintenance"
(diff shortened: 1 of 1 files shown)


Karmic-Skink:
I believe this was set intended to prevent antags bouncing back and forward off it, and also making the journey between the two be actually something significant.

Also rule of space 80s dictates personal long range teleporting to be Future Tech.

TheChaser212:
> I believe this was set intended to prevent antags bouncing back and forward off it

I think the amount of setup required to teleport there and back makes this a bit of a non-issue as you'd need to go there, create a way to teleport back and also setup a way to teleport there, which would be difficult to do without being noticed.

> Also rule of space 80s dictates personal long range teleporting to be Future Tech.

This doesn't really stop long range teleporting as you can just teleport to the space around the mining ship or make a new area on the mining z level to teleport back.

francinum (approved):
Improves consistency with player expectations.

PR #989, opened 2020-10-25
Pegasus atmos changes

MacBlaze1:
## About The Pull Request

Reworks the Pegasus atmos to be a bit more compact while also allowing for an n2o input line to the stormdrive
image

## Why It's Good For The Game

This has been requested by a veteran engineer IPC, A.I.M. and I believe it would allow for easier input of n2o into the engine, theres also some extra room for building.

## Changelog
:cl:

tweak: changed the Pegasus atmos to be a bit more compact and adds an n2o line
tweak: removed the useless plasma filter line for the magnetic constrictors 
/:cl:

### DIFF
--- a/_maps/map_files/Pegasus/pegasus1.dmm
+++ b/_maps/map_files/Pegasus/pegasus1.dmm
@@ -1350,10 +1350,6 @@
 	},
 /turf/open/floor/monotile/dark,
 /area/science/mixing)
-"aLI" = (
-/obj/machinery/atmospherics/pipe/simple/dark/visible/layer3,
-/turf/closed/wall/steel,
-/area/engine/atmos)
 "aLK" = (
 /obj/effect/turf_decal/stripes/line{
 	dir = 6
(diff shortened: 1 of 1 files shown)


Altangy:
- [ ] Plasma Constrictors are a type of machine similar to freezers. Putting them in walls in not only bad form but frowned upon
- [ ] Gas Mixers or Filters, even if they start on, should not be inside of a gas tank where they can not be reached. Especially if that gas tank will be pressurized with highly flammable gas
- [ ] Double pipe at X 190 Y 119 on the third layer. A green pipe and a passive pump. What is a passive pump doing on the filter line anyway?

Now that the critical issues that should block the PR from a stance of mapping standards, here's my problems with this prebuilt direct line of N2O as the guy Kmc asked to make this atmos setup for.

Firstly, putting the N2O mixer where it is changes the math slightly so instead of knowing exactly what percentage of fuel is going into the mix at a glance (which is most important since with too little the storm drive will starve) N2O is most accurately modulated.

Secondly, Kmc requested that I build a functional set up that didn't hand feed the players. I'd like to congratulate you for finding the layer adapter I left to make connecting the N2O to the set up easier as you are one of the few players to actually look. Now, one thing both of us know is that while all gasses in the tanks here have gas reactions with the SD not all of them are used by players. Layer adapters have been left to allow players to hook up inputs or they can just steal the feed pipe for their own gas.

At the very least the first three points need to be addressed. Double pipes can cause Out Of Memory errors which crash the server so in the future please be mindful of these really hard to spot issues in the future.

Altangy:
> 

> On top of the doublepipe mentioned above, there are also a few unusually exposed distro manifolds. Likely using /visible instead of /hidden type manifolds

Those are from the night before I sent it to Kmc when he told me he needed it in 6 hours at midnight my time. I've intended to fix them but they're non-critical.

MacBlaze1:
fixed 1,2 and 3 on local made plasma in front, fixed double pipes

MacBlaze1:
@Altangy check it out now

Altangy:
I just looked at it. No issues but it looks like a wall pipe was added. Unless someone cares about that minor detail it can be merged

Vasily2013 (changes requested):
The four way manifold should be replaced with a three way.
image

Overall I think this is a little bit more complex than it needs to be, but then again I designed the current Tycoon atmos, so who am I to talk. This looks to be good, but there are enough changes to merit a test merge first after that four way gets fixed.

MacBlaze1:
> The four way manifold should be replaced with a three way.
> image
> 
> Overall I think this is a little bit more complex than it needs to be, but then again I designed the current Tycoon atmos, so who am I to talk. This looks to be good, but there are enough changes to merit a test merge first after that four way gets fixed.

Fixed that

Vasily2013 (approved):
No observed issues.

PR #1006, opened 2020-10-30
Makes the gauss gun reload itself if its out of ammo.

MLGTASTICa:
## About The Pull Request
Makes the gauss gun try to reload itself when you try to fire without ammo.
## Why It's Good For The Game
More often than not , the gunner has to come down to raise the rack . This makes using the gauss tedious as you have to go down and up everytime,  because the loader did not raise the rack.
## Changelog
:cl:
add:Gauss guns will now raise their rack if you try to fire with no ammo.
/:cl:

### DIFF
--- a/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/gauss_gun.dm
+++ b/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/gauss_gun.dm
@@ -182,6 +182,9 @@
 	dir = WEST
 
 /obj/machinery/ship_weapon/gauss_gun/proc/onClick(atom/target)
+	if(ammo.len<1)
+		to_chat(gunner, "<span class='notice'>Attempting to load ammo!</span>")
+		src.raise_rack()
 	if(pdc_mode && world.time >= last_pdc_fire + 2 SECONDS)
 		linked.fire_weapon(target=target, mode=FIRE_MODE_PDC)
 		last_pdc_fire = world.time
--- a/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/gauss_gun.dm
+++ b/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/gauss_gun.dm
@@ -182,7 +182,7 @@
 	dir = WEST
 
 /obj/machinery/ship_weapon/gauss_gun/proc/onClick(atom/target)
-	if(ammo.len<1).
+	if(ammo.len<1)
 		to_chat(gunner, "<span class='notice'>Attempting to load ammo!</span>")
 		src.raise_rack()
 	if(pdc_mode && world.time >= last_pdc_fire + 2 SECONDS)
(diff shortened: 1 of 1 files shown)


ta3370:
Communication is important on a crewed weapon. This situation can easily be fixed by communication and people over time have already found several ways to increase communication such as spraypaint to label spots, which people can then call out on radio. Also add on that each Gauss bay is always the same, Top left, Top middle, Top right, Bottom left, Bottom middle, Bottom right.
"Ted, reload bottom middle!".
On maps such as Pegasus where you have two Gauss bays, if you have Ted running between both, you need Fred to get his lardbutt down from the cannon and start reloading the other bay. "Fred, bottom left!".

With spraypaint (Color and/or Number) or simply defining numbers as top left to bottom right, you can whittle this down to "Reload 4!", "Reload Blue!", or similar.

This is not meant to be a solo rambo weapon, it's meant to be a bonus weapon gained by a larger crew working together to add to the main battery. And a _major_ part that is critical to working _together_ is this wonderful thing known as _communication_.

MLGTASTICa:
> Communication is important on a crewed weapon. This situation can easily be fixed by communication and people over time have already found several ways to increase communication such as spraypaint to label spots, which people can then call out on radio. Also add on that each Gauss bay is always the same, Top left, Top middle, Top right, Bottom left, Bottom middle, Bottom right.
> "Ted, reload bottom middle!".
> On maps such as Pegasus where you have two Gauss bays, if you have Ted running between both, you need Fred to get his lardbutt down from the cannon and start reloading the other bay. "Fred, bottom left!".
> 
> With spraypaint (Color and/or Number) or simply defining numbers as top left to bottom right, you can whittle this down to "Reload 4!", "Reload Blue!", or similar.
> 
> This is not meant to be a solo rambo weapon, it's meant to be a bonus weapon gained by a larger crew working together to add to the main battery. And a _major_ part that is critical to working _together_ is this wonderful thing known as _communication_.

1.This just raises the rack , it doesn't load the slugs themselves onto the rack.
2.SS13 is a game where the chat is used often , there is no voice chat , meaning that communicating for reloads is very likely to get lost in the chat logs , or be tedious for the gunner.
3.This does not in any way make it a solo rambo weapon , it just makes them less dependant on the loaders bothering to load 4 UI's and  pressing load ammo at the right time.
4.Is this fun ? having to everytime you want a reload spend seconds typing in the chat ? what if you repeat it every 6 seconds because of how easy it is to waste ammo, doesn't it become annoying ?

Plotron:
I agree with this idea considering that you cannot easily tell the state of gauss guns without opening the UI. This has the added benefit of giving visual and aural feedback when a gunner does their job and needs a refill — you can easily see the rack going up and down and needing a refill afterwards.

Kmc2000 (approved):
Looks fine to me.

I think this is an acceptable QOL thing. The crews still have to coordinate loading etc. but this'll make their lives a bit less tedious if they've actually loaded the guns.

PR #1026, opened 2020-11-06
Fixes all active turfs found on mining ruins

IndusRobot:
## About The Pull Request

At first I just looked at mining11.dmm, ruined stormdrive. Previously, the asteroid ruin was mixed airless sand turfs and pressurized reinforced plating, monotiles and tileless plating. Now it's all airless. Also removes a tiny fan

Then I went ahead and just checked them all 

## Why It's Good For The Game

Lag bad


Vasily2013 (approved):
Active turfs bad! (Sometimes funny when they yeet miners into walls, but BAD!)

PR #1027, opened 2020-11-06
Armor increasing fighter max hp is no longer completely useless

DeltaFire15:
## About The Pull Request
As of now, armor (T2+) increases the max integrity of fighters, but.. resets it immediately upon the armor being destroyed, which will ALWAYS happen before the actual fighter integrity is influenced, making that part of them.. useless.

This PR adds an additional arg to remove from / remove_hardpoint, which determines if a component is being voluntarily removed (0), or if it is caused by damage (1). Currently only influences armor component removal.
This should make it so that the integrity increase actually has an effect, while preventing trying to exploit it via e.g. installing and then immediately removing armor. The max integrity will still reset upon installation of new armor.

I could alternatively just remove the armor integrity increase, and make armor more durable by that amount, but this seems like a better solution.

## Why It's Good For The Game
The integrity increase promised by armor actually working is a good thing, isn't it?

## Changelog
:cl:
fix: Fighter integrity increases caused by armor no longer immediately reset on armor breach.
/:cl:


Kmc2000 (changes requested):
Please use the TRUE and FALSE macros for clarity. Code looks fine though!

DeltaFire15:
I have quite literally no idea why I didn't use those in the first place, but either way, it's done now.

Kmc2000:
Please resolve the conflicts and I'll test-merge this @DeltaFire15

DeltaFire15:
Conflicts resolved.

Kmc2000 (approved):

PR #1028, opened 2020-11-06
Fixes RBMK valve control console exploitability

DeltaFire15:
## About The Pull Request
Turns out, as opposed to the code comment, valves do NOT sanitize the values provided by the signal, which means you were able to set the pumps to absurd values. This PR fixes that.

## Why It's Good For The Game
Being able to set inacceptably high pressures on pumps via these consoles ain't good.

## Changelog
:cl:
fix: RBMK valve control consoles can no longer set pumps to higher pressures than allowed.
/:cl:


francinum:
This should be fixing the sanitization on the pump, not on the console.

Kmc2000:
Pipes do actually sanitize this value already, hence why I didn't bother

Kmc2000 (approved):

Issue #666, opened 2020-06-15
Dying on the ship combat z-level, makes the auto cloner not work.


GoatMonarch:
## Round ID:
 470 (Also any round where a fighter pilot gets a pre-scan, and dies in space)

## Testmerges:
No test merges

## Reproduction:
Having the cloner upgraded so it's capable of the auto-proses.
Get yourself scanned with said auto cloner.
Die on the ship combat z-level. Easiest to do whilst a fighter pilot and undocked.
The auto cloner is now unable to clone you for some reason. Nor can someone manualy clone you with the records.

In acordance to Altnagy our seniormin, you require admin intervention to be clonable by doing some monkey brain frickery.

GoatMonarch:
No changes related to this has been made as far I am aware. So it's still an issue.

PR #1035, opened 2020-11-07
Makes pilots not get deleted (Should fix uncloneability)

### LINKED ISSUE #407: Fighter ejection pods qdel pilot on kill instead of ejecting them
## Round ID:
121

## Testmerges:
#386 

## Reproduction:
Put a pilot in a locked canopy fighter
Shoot them with small arms until the fighter breaks
put like 3 bullets into the escape pod
bam, pilot qdels with all their shit.

Bokkiewokkie:
## About The Pull Request
Stops pilots from getting deleted along with their fighters when they blow up, they are instead sent to (and murdered on) the empty space z level.
should fix: #666 and fix #407 

## Changelog
:cl:
fix: Stops fighter pilots' corpses from getting deleted upon ship death.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/fighters/fighters2.dm
+++ b/nsv13/code/modules/overmap/fighters/fighters2.dm
@@ -40,6 +40,10 @@ Repair
 	toggle_canopy()
 	forceMove(get_turf(locate(255, y, z)))
 
+/obj/structure/overmap/fighter/Destroy()
+	throw_pilot()
+	.=..()
+
 /obj/structure/overmap/fighter
 	name = "Space Fighter"
 	icon = 'nsv13/icons/overmap/nanotrasen/fighter.dmi'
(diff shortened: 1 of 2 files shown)


Karmic-Skink:
TBH I don't really like this as a solution.

Bokkiewokkie:
I agree this isn't the best solution but right now it's the only solution I'm seeing. If you can make a better option I'd be glad to just close this and let your PR get merged. So I think the best solution for now would just be (test)merging this and reverting it when a better solution is made.

francinum:
Currently this behavior is also applied to ships on valid dock z-levels.

It should not. It should drop them horribly dead on the spot.

Bokkiewokkie:
> Currently this behavior is also applied to ships on valid dock z-levels.
> 
> It should not. It should drop them horribly dead on the spot.

Yep, I tried to implement that but decided not to because I'm not a very experienced coder. And I was already happy that the main part with the yeeting into space worked because it'd still allow them to get cloned even if the ship does explode on the Z level.
I'm willing to take another shot at making it drop them when they're on a z level with the station trait though.

Vivalas:
I strongly believe that this PR is responsible for the bug earlier where a pilot had all his references to a deleted fighter intact and wasn't able to be interacted with at all.

If the fighter explodes, the references the pilot has to the fighter need to be handled. I think before that was being handle dby the pilot being deleted, but now without the pilot being deleted it's causing problems.

Bokkiewokkie:
Okay so I've managed to fix (remove) all of the references to the player except for the one that keeps the camera locked in place 😔

Kmc2000:
Use the unfuck_overmap() proc in a pinch

Kmc2000:
Travis is failing due to that call()() usage.

I'd just use unfuck_overmap()

Bokkiewokkie:
This should be completely finished now.

Kmc2000 (approved):

PR #1042, opened 2020-11-11
[Fix] Resolves an issue with physics component dereferencing

Kmc2000:
## About The Pull Request

* Resolves an issue where physics components wouldn't have their memory references released easily, relying on expensive del() calls.

## Why It's Good For The Game

Should free up some CPU.

## Changelog
:cl: Kmc2000
fix: Fixed an issue with physics component reference holding.
/:cl:


nsv13/code/modules/overmap/physics2d.dm
```diff
@@ -83,6 +83,11 @@ PROCESSING_SUBSYSTEM_DEF(physics_processing)
 		var/obj/item/projectile/P = holder
 		if(istype(P))
 			P.physics2d = null
+	for(var/I in SSphysics_processing.physics_levels)
+		var/list/za_warudo = SSphysics_processing.physics_levels[I]
+		for(var/datum/component/physicsbody in za_warudo)
```
francinum, line 88: I think `if (find(src))` works and is cheaper? Ask coderbus, but this feels horribly expensive and jank, and you're looping over all active physics objects regardless of if you find yourself in literally the first object of the first level.

francinum (dismissed):

PR #1047, opened 2020-11-14
Fix JSON verifier only verifying the very first JSON file

SpaceManiac:


.vscode/settings.json
```diff
@@ -1,3 +1,3 @@
 {
-	"dreammaker.extoolsDLL": "byond-extools.dll",
+	"dreammaker.extoolsDLL": "byond-extools.dll"
```
Kmc2000, line 2: What’s with removing this comma?

Kmc2000 (commented):
Looks fine to me, just want to double check that comma change was intended. Thanks for the contribution!

.vscode/settings.json
SpaceManiac: JSON verifier complains about it, because BYOND would, if it tried to load this file. VSCode handles it fine though.

Kmc2000 (approved):

PR #1095, opened 2020-11-28
Fixes some camera errors, fighter docking slightly cleaned up

Kmc2000:
## Changelog
:cl:
fix: Resolved some bugs with Galactic Conquest
/:cl:

PR #1096, opened 2020-11-28
Fix tuning fork atmos

Cenrus:
SpaceSHIP Map now has a functioning atmospherics compartment with actual air in the tanks. Previously they didn't use the right turf types.

Added nucleium filter to the reactor so that the coolant line doesn't get flooded

Added vent to space in atmos (filter was getting clogged, breaking airlocks)


Kmc2000:
You may need to redo this, #1095

Cenrus:
shouldnt conflict because the other PR only changes cameras

Vasily2013 (approved):
Defeat!

PR #1104, opened 2020-11-29
Brazil 2.5

DeltaFire15:
### What does this do:
Thats pretty easy to summarize:
It replaces the existing algorytm for sector three with a new one, which should make navigating it  easier, but still keeps it one hell of a uncharted sector. Basically operating on a tree-like structure and adding some random connections.
Also fixes some sector-generation related stuff, like the Rubicon not being taken into account for distance checks when placing systems.

### How does this work?
Thats gonna take a bit longer to explain.
Alright, so, first we create all the systems as usual, but don't create connections yet.
Now, we grab the system closest to Rubicon and use it as the Rubiconnector, that part stays the same.
Here's where the differences begin: 
Starting from the Rubiconnector as source node, it runs a slightly modified version of the Dijkstra algorytm, which gives every system a parent, and using that creates a tree guaranteering no system will have no way to reach the rubicon at all. There's a very important define for this:

*NONRELAXATION_PENALTY*: Effectively, this forces the generator to not just 'cheat' us and connect every system to the Rubiconnector directly, as obviously the shortest path there is the direct one. This incentivises using relaxations, or well, penalises not using them by multiplying the nonrelaxed distance by that amount.

Now, after thats done, we have a tree. Issue is, obviously, trees have no cycles, and therefore the map would be pretty boring - you just have one, and *exactly* one way from any system a to system b.
This is why we enter the second phase:
We go through each system, and give some of them additional connections, if they don't have too many already, aswell as trying to not have those connections be too long as jump lines going to the other side of the map take aaaages to jump (Though these are allowed during initial tree generation, but probably very uncommon especially with high system counts)
There also is a few defines for that part, though none of these have any influence on the initial tree generation.

*MAX_RANDOM_CONNECTION_LENGTH*: This determines how long the connection between two systems is allowed to be at most.

*MIN_RANDOM_CONNECTION_LENGTH*: Same, but for the minimum distance. Set to 0 by default, but hey, if someone wants to create wacky stuff I ain't gonna stop them.

*RNGSYSTEM_MAX_CONNECTIONS*: If systems have at least this many connections, they are ignored by the random jumpline generation when searching for targets. It also early aborts creating additional connections originating in systems with that amount or more connections.

*RANDOM_CONNECTION_BASE_CHANCE* and *RANDOM_CONNECTION_REPEAT_PENALTY*: During random gen, each system rolls repeatedly to add new jump lines to itself until either a) it becomes invalid, b) it has no more valid targets, or c), most common probably, it fails a roll with the probability of *RANDOM_CONNECTION_BASE_CHANCE* - *RANDOM_CONNECTION_REPEAT_PENALTY* * connections already created by the random connection generator during the handling of that specific system.

Example of an end result (obviously not the full one because screen too small, but this is a good chunk of the sector):
connection-example-1

## Why It's Good For The Game
Makes people able to actually navigate Brazil, even at, or well, especially at high amounts of systems aswell as fixing a bunch of issues with it (e.g. systems sometimes having no way to rubicon, tons of one-way routes, alot of issues with connections of the type
 a -> b -> c -> a without connections to other systems, etc.), while still keeping it a highly randomized sector with many hazards, and at least some difficulty of mapping.

## Changelog
:cl:
add: Brazil now uses a new algorytm for connection creation. Rejoice, Navigators.
fix: Fixes some things in the sector 3 generation misbehaving.
/:cl:


francinum:
Much much better.

Kmc2000 (approved):

PR #1111, opened 2020-12-02
Ramming asteroids with the player ship now destroys the asteroid and summons a meteor event. Disables meteor events on landing in a debris type system

IndusRobot:
## About The Pull Request

Changes how the meteor event is handled to tie it closer to the overmap, rather than triggering the meteor event at random

Preview:
https://drive.google.com/file/d/1oX2kkDyX-S7ie1-dmh81TEXmVE_0EI03/view?usp=sharing

There are several checks before summoning meteors by collision: 
The target must be an asteroid
The bump velocity is high enough 
The attacker must be a player ship, not ai_controlled
The attacker must not be a mining ship

The meteor event severity is based on the meteor size: 
Large: Threatening meteors 
Medium: Normal meteors
Small: Major space dust

Large and medium meteors will send a warning, but small meteors will not send a warning

## Why It's Good For The Game

Random meteors usually lead to shuttle calls and can end a perfectly good round. This takes the meteor event away from randy random and puts it in the players hands. Bad pilots are now to blame for meteors!

## IndusRobot
:cl:
tweak: Ramming asteroids with the player ship now summons meteors. Centcom reminds all bridge staff to fly responsibly, and not to bump the ship into things. 
tweak: Landing in a debris type system no longer has a chance to randomly summon meteors. 
/:cl:


nsv13/code/modules/overmap/physics.dm
```diff
@@ -494,7 +494,7 @@ The while loop runs at a programatic level and is thus separated from any thrott
 		velocity.x -= bump_impulse
 	return ..()
 
-/obj/structure/overmap/Bump(atom/movable/A, datum/collision_response/c_response)
+/obj/structure/overmap/Bump(atom/movable/A, datum/collision_response/c_response, atom/movable/B)
```
ondrej008, line 497: src and B should be equal, so you can get rid of the atom/movable/B and replace all B. things with either src. or just nothing.

nsv13/code/modules/overmap/physics.dm
```diff
@@ -512,6 +512,34 @@ The while loop runs at a programatic level and is thus separated from any thrott
 		return ..()
 	// if a bump is that fast then it's not a bump. It's a collision.
 	if(istype(A, /obj/structure/overmap) && c_response)
+		// If the target is an asteroid
+		if ( istype( A, /obj/structure/overmap/asteroid ) )
+			// If the attacker hits the rock hard enough
+			if ( bump_velocity >= 3 )
+				// If the attacker is a player ship, and not a mining ship
+				if ( istype( B, /obj/structure/overmap/nanotrasen ) && !istype( B, /obj/structure/overmap/nanotrasen/mining_cruiser ) )
```
ondrej008, line 520: AI ships are also /obj/structure/overmap/nanotrasen, so this would also affect them, maybe a better way would be checking if B.role == MAIN_OVERMAP

ondrej008: Which then could also get rid of the AI check on 523, though this is probably fine as it is.

Bastian0930:
Should make it not go through shields though, since yknow, shields

nsv13/code/modules/overmap/radar.dm
```diff
@@ -35,8 +35,8 @@
 	var/scanning_speed = 2 //Duration of each pulse.
 	var/last_scanning_speed = 2 //To update the sound loop
 	var/start_with_sound = FALSE //Used to stop fighters playing dradis sounds all at once and being annoying.
-	var/show_asteroids = FALSE //Used so that mining can track what they're supposed to be drilling.
-	var/mining_sensor_tier = 1
+	var/show_asteroids = TRUE //Used so that mining can track what they're supposed to be drilling.
+	var/mining_sensor_tier = 3 // Pilots need to see all asteroids so they don't bump into them
```
Karmic-Skink, line 39: Sounds like power gaming here. Let them locate them visually until upgraded.

PR #1113, opened 2020-12-02
Stop mining ship damage affecting the main overmap

Cenrus:
this didn't check for the main overmap object...

:cl:
fix: Mining ship FTLing when in SScrit will no longer cause the main ship to explode
/:cl:


Kmc2000 (approved):

Kmc2000:
Good catch

PR #1129, opened 2020-12-08
Adds an atmos holofan backpack

MacBlaze1:
## About The Pull Request

This PR adds an atmos holofan backpack that can deploy 12 holofans from a distance (viewrange). It works off the same principle as an atmos water tank, but instead of water it works off the principles of the atmos holofan.

## Why It's Good For The Game

It adds a new tool for engineers to use and can be used for massive breaches that so often happen. The normal holofan is often limited by range and number and mainly good for preventative measures, this one is more responsive to imminent danger and removes your backpack storage in exchange for breach stopping much like the atmos water tank removes back page storage for fire suppressing. 

## Changelog
:cl:
add:  New atmos holofan backpack that allows holofans to be placed from a distance 
/:cl:

### DIFF
--- a/code/datums/action.dm
+++ b/code/datums/action.dm
@@ -252,6 +252,9 @@
 /datum/action/item_action/toggle_mister
 	name = "Toggle Mister"
 
+/datum/action/item_action/toggle_powerpack
+	name = "Toggle Powerpack"
+
 /datum/action/item_action/activate_injector
 	name = "Activate Injector"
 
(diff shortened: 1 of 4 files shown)


Karmic-Skink:
This seems extremely powerful, what are you intentions for populating it into rounds?

Also, generally these things require proximity placement for balance reasons.

ta3370:
Holofans are already super overpowered magical "I win, atmos never matters anymore" devices that you can then ignore and walk through and around whenever you want, if anything I'd want more inflatable barriers around, so a PR porting bay's suitcase (or is it kit or placer?) method for using them easier and porting inflatable doors alongside them (That let atmos through when opened! Use logic and actual thinking and you can set them up in a way that loses less air!) would be much more preferable to me, but such a project would need to be passed through maintainers before starting.

No to magical "I click anywhere on my screen and suddenly atmos nolonger matters at no actual obstruction" device. I barely tolerate basic holofan devices.

MacBlaze1:
Ill put this on the backburner for when we have actual balancing methods

PR #1123, opened 2020-12-04
Allows the code to compile again

Kmc2000:
## About The Pull Request

https://github.com/BeeStation/NSV13/pull/1122/files introduced a fix that uses a subsystem that we don't have.

This botches it in temporarily, it somewhat works, but is dependent on a full TGUI 4 port which is out of scope for the time being.

## Why It's Good For The Game

Compiling the code good

## Changelog
:cl: Kmc2000
fix: Asset cache partial port
/:cl:

PR #1135, opened 2020-12-12
Allow linters to pass again

Kmc2000:
## About The Pull Request

This PR is a bodge. It leaves assets completely broken. It also allows linters to be useful again.

This is still trying to fix the fallout from #1123 and #1122 

I don't have the availability or distinct knowledge of the asset system to actually fix this.

If anyone can help with this...please do. This is really bad.

So. No assets will work in game, but your code can at least compile.

## Changelog
:cl:
fix: Temporarily disabled assets to allow code to compile
/:cl:


theboy6545:
Why is everything breaking?

PR #1146, opened 2020-12-19
Smarter FTL AI

DeltaFire15:
## About The Pull Request
As usual, foreword:
Testmerging this may be useful before a fullmerge, I localtested a bunch but it might have edgecases I didn't account for.

This PR adds a pathfinding algorytm for FTL / the starmap, with a bunch of variable paramaters.
It gets used by any fleets that have a goal_system set, which is:
- Syndi armadas
- Midround randomly spawned syndi fleets that don't start in setor 2 aka the neutral zone
- Fleets that get their goal_system varedited from null
- Interdiction fleets (more on those later)

Movement for fleets now works this way:
If there is no target system directly given by the move(), it checks if the fleet has a goal system. If yes, it generates a course if there isn't one already, then uses the next system in route as jump target, or, as fallback, if there is no route to their target, they use the old random jump technique. If they already arrived at their goal system, they'll just stay there and continue occupying it (no more Armadas leaving Sol alone after managing to arrive there).

Now, what is a Interdiction fleet?
There are a very **fun** new fleet type possible due to fleets now having access to pathfinding. Interdictors have drives with very low jump recharge times, and with a very clear goal: *You*.
They can and will chase down the main ship (or any other ship if admins varedit their hunted_ship var), and try to force it into an engagement, until they are destroyed, or they manage to destroy you. Very scary.
The strong versions of these are adminspawn only, though a weaker version can be spawned by the syndicate when they send a new fleet out randomly (not for mission systems, as they'd just leave), though that is *comperatively* rare to happen. I'm sure you'll manage to distinguish them from normal syndicate fleets and react accordingly... riight?

There's a bunch of vars for all this stuff that can be tweaked, so, if you don't want armadas to be able to use wormholes for example, yeah that is one singular var.
Also supports black- / whitelisting of specific allegiances for the pathfinding, though these are not used as of now.

## Why It's Good For The Game
Actual pathfinding being available for AI fleets might be pretty useful.
Also Interdictors are a fun new enemy type to get *accustomed* with that should add some variety.

## Changelog
:cl:
add: Some AI fleets have realised they can actually navigate instead of just jumping randomly.
add: Intercepted communications suggest the Syndicate has started to deploy so-called 'Interdiction Fleets' with advanced FTL tracking technology. Stay vigilant.
/:cl:


theboy6545:
I feel like the interdiction fleets sending out a ping of some form, like a submarines sonar, before it jumps would be interesting, so you can realise you are about to be jumped.

DeltaFire15:
> I feel like the interdiction fleets sending out a ping of some form, like a submarines sonar, before it jumps would be interesting, so you can realise you are about to be jumped.

If you or someone else can supply a ping sound that fits then yeah sure I could make them give an audible ping as warning if they finish a jump within ~2 jumps of the ship they're chasing.

Kmc2000 (approved):

PR #1174, opened 2021-01-18
Adds a NAV onto the Jepp

theboy6545:
adds a NAV, and the associated giblets

## About The Pull Request

Adds a NAV to the Jepp, and strips out the railgun and other junk that was there, I also added a small lounge area where the railgun was, and modified where the escape shuttle lands to make space for it (also added airlocks because they are good)
## Why It's Good For The Game

Oldnitions bad, newnitions good! (Screw you Ansome it's a good name)
## Changelog
:cl:

tweak: Tweaks munitions and the arrivals shuttle.

/:cl:


theboy6545:
I'm not acting as life support for a ship.

PR #1179, opened 2021-01-20
Gladius Atmos Changes

SerynEngi:
## About The Pull Request

This PR updates the Gladius atmos setup to reflect a standard to be implemented across all maps.

## Why It's Good For The Game

The changes will hopefully make it easier for people newer to atmos to understand and add in emergency vents to the reactor supply lines so that a mistake made by engineering in the first 20 minutes does not kill a round.

## Changelog
:cl:
tweak: Changed the piping layout of atmospherics on the Gladius.
Gladius Atmos Changes

/:cl:

### DIFF
--- a/_maps/map_files/Gladius/Gladius2.dmm
+++ b/_maps/map_files/Gladius/Gladius2.dmm
@@ -4,9 +4,15 @@
 /area/space)
 "aan" = (
 /obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer3,
-/obj/machinery/atmospherics/pipe/simple/supply/hidden/layer1,
+/obj/machinery/atmospherics/pipe/simple/supply/visible/layer1,
 /turf/open/floor/plasteel/grid/steel,
 /area/engine/atmos)
+"aar" = (
+/obj/machinery/atmospherics/pipe/simple/purple/visible{
+	dir = 4
+	},
+/turf/open/floor/plasteel/grid/lino,
+/area/engine/engineering/reactor_control)
 "aaD" = (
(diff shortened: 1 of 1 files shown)


Vasily2013:
You forgot to flip the filters around.

SerynEngi:
> You forgot to flip the filters around.
This is slightly unclear. The filters are designed in this map to pull from the right hand side, first filtering O2, all the way through plasma at the end on the left side. This is reversed from their original order.

SerynEngi:
> You forgot to flip the filters around.

Thank you for that. Fixed in the commit made shortly before this message posts. Also added in a vent for the GenMix tank as in the original design doc (though placed differently).
Capture2

Vasily2013:
This gas meter is on the wrong layer and doesn't connect to the pipe. 
image

Same thing on the distro loop.
image

Wrong layer on both the freezer and heater for distro. 
image

I'd suggest making an attachment between the pure line and the connectors so you don't have to go through the mix tank. Here's my suggested pipe layout. Things in grey have been changed. Make sure you change the direction of flow for the pump at the south towards the scrubber network.
image

Finally: the amount of time the air to distro pipe stays in space causes about 6 degrees of cooling overall. Consider moving it inside if possible, but if not that's easy enough to fix with the heater and freezer.

SerynEngi:
Issues corrected. The notable departures from the design included in your example is the pump being removed from the line on the return to GenMix tank side of that line, as it would block the flow in the direction included, and would be redundant if it was facing the opposite direction. All major changes highlighted in purple.
Capture3

Vasily2013 (approved):
image

PR #1191, opened 2021-01-25
Some breach fixing stuff

DeltaFire15:
## About The Pull Request
This PR does the following:
Firstly, it does a small QoL change allowing RCDs to be used on hull plating, which will instead target the turf below, in case there is a breach below hull plating and you don't want to have to alt-click the hull plating to be able to click the actual turf.

Secondly, this fixes smartfoam not fixing floor / 'ceiling' breaches inbetween z-levels, as it did only fix explicitly space tiles, not openspace.

## Why It's Good For The Game
Fix man good.
Some QoL is neat too, saving a few clicks / seconds isn't too needed, but it shouldn't cause issues either.

## Changelog
:cl:
fix: Using smartfoam grenades on breaches that cause open space (aka, on the top deck of a ship) works now.
tweak: Using a RCD on hull plating will now attempt to use it on the turf below instead.
/:cl:

### DIFF
--- a/code/game/objects/effects/effect_system/effects_foam.dm
+++ b/code/game/objects/effects/effect_system/effects_foam.dm
@@ -119,7 +119,7 @@
 	STOP_PROCESSING(SSfastprocess, src)
 	if(metal)
 		var/turf/T = get_turf(src)
-		if(isspaceturf(T)) //Block up any exposed space
+		if(isspaceturf(T) || istype(T, /turf/open/openspace)) //Block up any exposed space
 			T.PlaceOnTop(/turf/open/floor/plating/foam, flags = CHANGETURF_INHERIT_AIR)
 		for(var/direction in GLOB.cardinals)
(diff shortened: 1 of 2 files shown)


code/game/objects/effects/effect_system/effects_foam.dm
```diff
@@ -119,7 +119,7 @@
 	STOP_PROCESSING(SSfastprocess, src)
 	if(metal)
 		var/turf/T = get_turf(src)
-		if(isspaceturf(T)) //Block up any exposed space
+		if(isspaceturf(T) || istype(T, /turf/open/openspace)) //Block up any exposed space
```
covertcorvid, line 122: Please add an "NSV13" tag to this change since it's in the base code

DeltaFire15: done

Kmc2000 (approved):
The hero we need

PR #1199, opened 2021-01-26
Some AGCNR tweaks

DeltaFire15:
## About The Pull Request
While using the AGCNR & seeing it in operation quite some times, I noticed a few problems with it:
1. Frequent atmos lag can cause the reactor to slowly take damage due to it not having enough coolant for a single tick every now and then.
2. Some meltdown types can be absurdly fast from no danger to 'blowout time', with zero possible counterplay as you have ~5-10 seconds to react (Which is not enough even if by chance you are close to the reactor control room at that time)
3. One can chain doafters when fixing the reactor to fix it more than should be possible.

For the first issue, I made it so it require five ticks without enough moles to start accumulating damage due to low coolant. This number goes down twice as fast when there *is* enough coolant, so it takes half as much time with coolant as you were without to get it back to 0.

For the second, I capped the amount of integrity damage heat / pressure can cause, making it so you get at least 40 seconds when it is melting, or 45 if it is having critical pressure (assuming your reactor was at full integrity before). **These values are not final, if that seems like too much time, I can lower them.**

And lastly, for the third, I added a check preventing bypassing of the integrity threshold.

## Why It's Good For The Game
These tweaks should address a few issues the AGCNR has, which should make it better to work with.

## Changelog
:cl:
tweak: The AGCNR now only starts taking damage from having low coolant amounts after a few seconds have passed with that condition.
tweak: Added a cap to AGCNR meltdown speed, giving engineering some time to react.
fix: Repairing the AGCNR's vessel can no longer be cheesed by chaining doafters.
/:cl:

### DIFF
--- a/nsv13/code/modules/power/rbmk.dm
+++ b/nsv13/code/modules/power/rbmk.dm
@@ -395,7 +395,7 @@ The reactor CHEWS through moderator. It does not do this slowly. Be very careful
 	if(temperature >= RBMK_TEMPERATURE_CRITICAL)
 		alert = TRUE
 		if(temperature >= RBMK_TEMPERATURE_MELTDOWN)
-			var/temp_damage = min(pressure/100, initial(vessel_integrity)/40)	//40 seconds to meltdown from full integrity, worst-case. Bit less than blowout since it's harder to spike heat that much.
+			var/temp_damage = min(temperature/100, initial(vessel_integrity)/40)	//40 seconds to meltdown from full integrity, worst-case. Bit less than blowout since it's harder to spike heat that much.
 			vessel_integrity -= temp_damage
 			if(vessel_integrity <= temp_damage) //It wouldn't be able to tank another hit.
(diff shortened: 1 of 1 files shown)


nsv13/code/modules/power/rbmk.dm
```diff
@@ -190,6 +193,9 @@ The reactor CHEWS through moderator. It does not do this slowly. Be very careful
 			to_chat(user, "<span class='notice'>[src]'s reactor vessel is cracked and worn, you need to repair the cracks with a welder before you can repair the seals.</span>")
 			return FALSE
 		if(do_after(user, 5 SECONDS, target=src))
+			if(vessel_integrity >= 350)	//They might've stacked doafters
```
Jalleo, line 196: Nice catch.

Jalleo (commented):
Seems good to testmerge.

Kmc2000 (approved):

PR #1219, opened 2021-01-31
Rewards antags with torps even if they redtext, disables greentext torps

DrSirKnight:
## About The Pull Request

Makes it so antags get 200 torps on roundend. Disables them from getting torps for greentexting.

First time actually coding something for byond so it's probably shitcode.

## Why It's Good For The Game

Only giving torps for greentexting encourages rushing objectives and disencourages creativity. By disabling greentext torps you are more encouraged to do objectives in creative ways even if they lead to death.

## Changelog
:cl:
tweak: Replaced greentext torps with torps for being a antagonist that are awarded at roundend
/:cl:

### DIFF
--- a/code/__DEFINES/metacoin.dm
+++ b/code/__DEFINES/metacoin.dm
@@ -12,3 +12,5 @@
 #define METACOIN_NOTSURVIVE_REWARD       10
 /// Rewarded when you are alive and active for 10 minutes
 #define METACOIN_TENMINUTELIVING_REWARD  4
+/// Rewarded for being a antagonist - NSV13
+#define METACOIN_ANTAGONIST_REWARD 200
--- a/code/__HELPERS/roundend.dm
+++ b/code/__HELPERS/roundend.dm
@@ -90,6 +90,7 @@
 	var/list/team_ids = list()
 
 	var/list/greentexters = list()
+	var/list/antagonist_list = list() // NSV13 - List of antags for metacoins
 
 	for(var/datum/antagonist/A in GLOB.antagonists)
 		if(!A.owner)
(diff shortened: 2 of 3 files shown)


Jalleo (changes requested):
Code looks good but Kmc needs to review and to decide if its too much coin for the antags and choose what amount.
It does need a few extra //NSV13 code here statements just to help reduce the chance of issues when upstream merging.
Edit: And if this is a good idea or not.

Bokkiewokkie:
I think just giving torps for completing the "survive the shift" or "die a glorious death" objectives would be better instead of just letting people roll for traitor and then suicide for free torps.

DrSirKnight:
> 
> 
> I think just giving torps for completing the "survive the shift" or "die a glorious death" objectives would be better instead of just letting people roll for traitor and then suicide for free torps.

If you suicide as a antag you'll probably end up in trouble with the admins anyways

> 
> 
> Code looks good but Kmc needs to review and to decide if its too much coin for the antags and choose what amount.

The default amount is also 200, but we'll see what KMC thinks

Omppusolttu:
> The default amount is also 200, but we'll see what KMC thinks

Note that not everyone gets torps, and that is a fact the system might be balanced around.

Vasily2013:
I'm pretty sure normal antags already get torps for being antag. I could be wrong about that, but in any sense, torps for greentext are your motivation to greentext other than "it's greentext." In my opinion this is also a contradiction to antag policy.

DrSirKnight:
> 
> 
> I'm pretty sure normal antags already get torps for being antag. I could be wrong about that, but in any sense, torps for greentext are your motivation to greentext other than "it's greentext." In my opinion this is also a contradiction to antag policy.

You don't get torps for being a antag. The reason I've done this is because being a antag shouldn't entirely be about greentexting. It should be about making the round fun! And by awarding greentexting, you're only telling people that the way to win is doing the generic meta traitor strats. If people weren't awarded for greentext they'd be far more open to going with creative approaches towards their objectives, even if it may result in failure.

Jalleo:
On further reflection and due to Kmc not looking at this PR and the lackluster and reported aspects of people saying that just by surviving they would get torps. This PR is redundant and gets closed. Its cosmetics. If there were non cosmetic items torps would then be power gamed and we would be having a fair amount of people just flat out green-texting for specific items. We dont have that happening.
People choose how to play as antags. Not the mechanics. Its a simple gimmick.
The only thing that isnt is a antag token which is a LOT of points for a reason. 200 is a hefty amount to give out alongside surviving a round.

PR #1227, opened 2021-02-05
(Probably) fixes Search & Destroy AI goals

DeltaFire15:
## About The Pull Request
Testmerge this first. It worked on local but might have overlooked something yadda yadda.

Anyway, the main things this PR does are
a) Actually removing ships from their fleet taskforce once they get destroyed, as that didn't work before due to byond things.
b) Making the 'Defend' aka stand around goal not runtime without targets.
c) Tweaking the Defend goal to also get priority 0 if all the supply ships got destroyed, so the ships go hunting down the guys that destroyed their supply line instead.

Additionally, it adds a small thing to the overmap_explode proc (just a early return in case there are no areas), aswell as compressing the Destroy() proc of overmap ships from THREE procs in THREE seperate files down into one in overmap.dm

## Why It's Good For The Game
Search and destroy working as intended is probably a good thing.
The other stuff's just small changes that shouldn't be negative.

## Changelog
:cl:
code: overmap structures no longer have their Destroy() split into three seperate procs.
fix: Ships should now behave as intended in regards to their search & destroy AI goal.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/ai-skynet2.dm
+++ b/nsv13/code/modules/overmap/ai-skynet2.dm
@@ -717,20 +717,28 @@ GLOBAL_LIST_EMPTY(ai_goals)
 	score = AI_SCORE_DEFAULT
 
 /datum/ai_goal/seek/check_score(obj/structure/overmap/OM)
+	message_admins("Running checks for search & destroy, for object [OM]")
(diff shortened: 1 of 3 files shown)


Kmc2000 (approved):
Didn't see any issues at testmerge phase

PR #1239, opened 2021-02-09
Stops Suit Storage Units deleting (traitor objective) items when hacked or destroyed

Bokkiewokkie:
## About The Pull Request
 
Adds EMP and Emag interactions for the Suit Storage Unit.
Makes it so SSU's are no longer IC versions of TG.
The code diff looks like a mess too so that could be considered a feature 👍 

## Why It's Good For The Game

Delete important item bad, deep fried suit good.

## Changelog
:cl:
add: You can now E-mag and EMP Suit Storage Units for fun times!
fix: Fixes Suit Storage Units deleting traitor objective items.
/:cl:

### DIFF
--- a/code/game/machinery/suit_storage_unit.dm
+++ b/code/game/machinery/suit_storage_unit.dm
@@ -40,9 +40,11 @@
 	* If the hack wire is cut/pulsed.
 	* Modifies effects of cook()
 	* * If FALSE, decontamination sequence will clear radiation for all atoms (and their contents) contained inside the unit, and burn any mobs inside.
-	* * If TRUE, decontamination sequence will delete all items contained within, and if occupied by a mob, intensifies burn damage delt. All wires will be cut at the end.
+	* * If TRUE, decontamination sequence will burn and decontaminate all items contained within, and if occupied by a mob, intensifies burn damage delt. All wires will be cut at the end.
 	*/
 	var/uv_super = FALSE
+	/// NSV13 for managing the messages sent back when the machine was hacked
+	var/toasted = FALSE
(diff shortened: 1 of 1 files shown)


code/game/machinery/suit_storage_unit.dm
```diff
@@ -261,47 +271,52 @@
...
 			if(occupant)
-				things_to_clear += occupant
-				things_to_clear += occupant.GetAllContents()
-			for(var/atom/movable/AM in things_to_clear) //Scorches away blood and forensic evidence, although the SSU itself is unaffected
-				SEND_SIGNAL(AM, COMSIG_COMPONENT_CLEAN_ACT, CLEAN_STRONG)
-				var/datum/component/radioactive/contamination = AM.GetComponent(/datum/component/radioactive)
-				if(contamination)
-					qdel(contamination)
+				visible_message("<span class='warning'>[src]'s door slides open, barraging you with the nauseating smell of charred flesh.</span>")
+				mob_occupant.radiation = 0
```
Jalleo, line 294: Uh you want the mob to not be iiradiated but burned badly instead for being killed in it?

Jalleo (commented):
One question then I will probably review more fully tomorrow.

code/game/machinery/suit_storage_unit.dm
Bokkiewokkie: This is already part of the original code for the SSU, it removes radioactive contamination from the occupant like it does with clothing because they just got cleaned VERY EFFECIENTLY.

Kmc2000 (approved):

PR #1264, opened 2021-03-02
Even more fleet stuff

DeltaFire15:
## About The Pull Request
Did you know that fleets right now quite literally get regenerated every time a player encounters them and leave their already existing ships in nullspace forever, yet in the fleet, so the fleet can't be defeated if it was encountered before?
Yeah, me neither.

This PR improves some more stuff for fleets, namely
Fleets now get assembled as soon as they have a valid system, creating their ships. Further assembles will merely attemps to 'unstuck' ships that are still in nullspace when assemble is called in a loaded system (since that usually happens after the main restore_contents proc / simillar are handled). They also now store / deploy their stored ships correctly (least on local), so things should be less jank.

Additionally, fleets now will only change faction influence if there is a player ship in the system, since why'd NT give you credit for kills you weren't even in the system for it (This is mostly in preperation for some *fun* stuff but for now it's just helpful)

Oh and I also fixed the security level being reduced back to GQ if you encounter an enemy while over GQ (Zebra, or Delta). Rejoice.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Fleets should no longer sometimes be undefeatable due to ships stuck in nullspace.
fix: Encountering an enemy will no longer force your alert level down to general quarters if it is higher.
code: Somewhat changed how fleets assemble themselves aswell as how they move.
balance: Fleets that are defeated outside of systems with players in them no longer modify influence (aka patrol completion)
balance: Pirate fleets are now worth less influence due to them being very easy to defeat.
/:cl:

### DIFF
--- a/nsv13/code/controllers/subsystem/starsystem.dm
+++ b/nsv13/code/controllers/subsystem/starsystem.dm
@@ -352,6 +352,7 @@ Returns a faction datum by its name (case insensitive!)
 		var/datum/fleet/fleet = new fleet_type(src)
 		fleet.current_system = src
 		fleets += fleet
+		fleet.assemble(src)
 	if(preset_trader)
 		trader = new preset_trader
 		//We need to instantiate the trader's shop now and give it info, so unfortunately these'll always load in.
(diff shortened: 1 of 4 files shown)


Kmc2000 (approved):
No issues have cropped up thus far

PR #1279, opened 2021-03-07
Munitions QOL pass!

Kmc2000:
## About The Pull Request

image

I had a play around with munitions and noticed a few things that could be improved upon, sooo here we go!

My notes from playing:

* Gauss is incredibly underpowered, and as an MT I was actively told to get off gauss. 
- To address this, Gauss has had its maximum ammo buffed, additionally, you can now call up the gauss rack as a gunner!
- Gauss racks can now be upgraded to support AmBer style loading with conveyors, you can find the chip in advanced munitions
- Gauss loading / Animation length has been slashed to let you get in more quickly, it's also received some code touchups.

* The munitions trolley is never used
- People use crates instead because of their lack of delay, and superior mobility. To this effect, I've made the munitions trolley loadable via conveyor belts. (It's also had a much needed UI revamp)

* People disregard the way I originally intended them to load the NAC.
- Instead of using conveyor systems like the ones we laid out on our maps, they just leave the powder laying about. 
- To fix this, I've given munitions ammunition sorters! Theyre all controlled via one central console, and can be tweaked / renamed to your liking. These should allow for more creative ammo loading setups. The boards are again, researchable. 
- Ammo racks are a major part of this update. I'm hoping to see more MTs involved in cannon loading, perhaps with one dedicated "preloader" who racks the right shells and powder onto the conveyor and so on.

## Why It's Good For The Game

Munitions is pretty fun, but I feel like they could use a little more room for creativity without exploiting janky game mechanics.

## Changelog
:cl: Kmc2000
add: Added munitions ammo racks! Store your ammo in a more organised fashion today.
tweak: Gauss has been significantly buffed to become more viable, check out the autoloading upgrade in the tech tree.
tweak: Gauss gunners can now raise the gun rack, assuming it's loaded. No more confusion about whether the gun's got ammo.
tweak: The munitions trolley has been modernised, and can now be loaded via conveyor.
/:cl:


covertcorvid:
Can you add deconstruction for the VLS while you're at it?

PR #1294, opened 2021-03-15
Fleet combat - Tabletop edition

DeltaFire15:
## About The Pull Request
You know how it's weird that fleets don't actually fight eachother if you aren't in the system?
Turns out they are playing dice games instead!

Every *3* minutes (subject to change probably, depending on if it seems too fast / slow), the starsystem controller goes through all systems with conflict (aka, systems with fleets of at least 2 different factions) and lets them handle combat.
Each ship of every fleet fires at one ship of a hostile fleet, using dice rolls (Target / Evade / Armor / Damage)
The final damage value then gets multiplied by DICE_DAMAGE_MULTIPLIER to convert it into actual integrity damage. Value can be tweaked for if npc fleets are supposed to beat eachother's heads in faster or slower.

Oh also the Syndicate armada can now actually siege Sol if it ever gets there, giving the players a FTL-esque timer before the system falls. Can be halted via being in the system, or reset if the armada is destroyed.

For admins: The star_system controller now contains a enable_npc_combat var, which when set to FALSE / 0 stops that sort of combat from happening. Useful if you are doing some stuff and don't want things to happen 'offscreen'.

Tested all of it on local and works fine there. Threw a bunch of fleets at eachother a few times, watched Sol getting sieged and simillar. Should probably be testmerged first regardless.

## Why It's Good For The Game
Having fleets fight eachother even without players around should bring a bit of life to the galaxy, especially considering you can now, say, find a Syndicate fleet that's already heavily damaged after having fought a NT fleet and is down its carrier, or simillar things that can liven up things a bit

## Changelog
:cl:
add: Fleets now also fight when no player ships are in the system.
add: The Syndicate Armada can now actually siege Sol.
/:cl:


Pontenerd:
Let ghosts gamble on the results of fights

DeltaFire15:
> > Every 3 minutes, the starsystem controller goes through all systems
> 
> Is this going to cause lag? Potentially in the middle of combat?

Shouldn't cause any noticable lag, considering it iterates through a maintained list (so just a few systems instead of all of them), plus the combat itself is just a few operations for each ship in battle.
If against my assumptions, it does cause lag (aka, processing use / overtime of the starsystem controller spikes during battleticks), could thrown in a invoke async aswell which'd make it at its own timing (a bit vulnerable to race conditions in very rare cases, but still not much of a issue regardless)

Karmic-Skink:
Are we checking points off when fleets get defeated from this? Really don't want patrols completing from the tabletop simulation.

Kmc2000 (approved):

PR #1300, opened 2021-03-18
Allows the Ion sabre to actually hit things that aren't humans

Bokkiewokkie:
## About The Pull Request

The Ion sabre was missing a line of code that allowed it to function as a normal weapon against things that aren't humans.
Also made it toggle on/off better and made it sharp, it's a sabre for a reason.
Oh and it has a material value now, so that's cool.

## Why It's Good For The Game

Ided from carp because sword would no interact when do the clicky

## Changelog
:cl:
fix: Fixed the ion sabre not being able to hit non-human targets.
/:cl:


Bokkiewokkie:
@Kmc2000 Can you like... actually testmerge it though?

Kmc2000 (approved):

PR #1297, opened 2021-03-18
The Beebasening 2021

covertcorvid:
## About The Pull Request
Merges BeeStation-Hornet as of March 19th into our codebase

## Why It's Good For The Game
Up to date code?

## Changelog
:cl:
code: Merged current Bee code
/:cl:

PR #1322, opened 2021-04-01
[WIP]Genpop 

GoatMonarch:
## About The Pull Request
This PR is going to add genpop to our maps. Basically, refurbishing perma into a general population cell. Where you are given an ID with a set timer. Code is yoinked and cherrypicked from [Oracle](https://github.com/OracleStation/OracleStation/pull/419).

Todo:

- [x] Get it to compile
- [ ] Fix it so our officers no longer has zero access
- [ ] Make the turnstill actually function
- [ ] More bug testing
- [x] Wait for #1297 to be merged 
- [ ] Profit

## Why It's Good For The Game
This is arguably one of the few good things from Oraclestation. Being thrown into brig is less of a hazzle for you and the security officers.
You: Can drop the soap in the cell showers, and talk about the weather to the mime stuck there with you.
Officers: Don't have to babysit your ass untill the timer is over, for you can just walk out yourself.

## Changelog
:cl:
add: Genpop is now a thing! 
imageadd: Turnstils and security closets.
tweak: Prisoner IDs renamed to Convict IDs.
add: They got a timer function too!
/:cl:


code/__DEFINES/access.dm
```diff
@@ -70,6 +70,8 @@
 #define ACCESS_FL 73 //NSV13 - FL tagged decorated fighter
 #define ACCESS_MINING_ENGINEERING 74
 #define ACCESS_MINING_BRIDGE 75
+#define ACCESS_ENTER_GENPOP 76 //NSV13 - Access to enter genpop
```
Kmc2000, line 73: Can you not use one-way door access helpers here? like in medbay

GoatMonarch: Dunno if the one-way door helper will work here. Since the turnstill technically is not an airlock.
I'll give this some testing though once I got the turnstills working.

tgui/src/interfaces/genpop_timer.ract
```diff
@@ -0,0 +1,17 @@
+<ui-display title='Genpop Timer' button>
```
Kmc2000, line 1: Let me know when you're mostly done with this PR, this is a tgui 1.0 interface, I'll help you update it.

GoatMonarch:
> Let me know when this gets approved and I'll turn Aether security upside down

It's actually #1367 you should keep your eye on. As Kmc got this and I've moved onwards to Lawyer additions Soon™. Kmc is gonna create an example settup on one of our ships that you can copy from.

PR #1324, opened 2021-04-02
April fools is literally bricking the job selection this is funny but also really isn't

DeltaFire15:
## About The Pull Request
The april fools holiday sets Clown as overflow job. Apparently, this bricks everything and causes every single job to runtime.
So. This removes the line doing that.

It-works-on-localTM

Attendum: The main cause for this is likely a race condition between the holiday triggering, and the mapping subsystem not having finished initing, leading to errors.

## Why It's Good For The Game
180 Job runtimes, after init.

## Changelog
:cl:
fix: The april fools holiday no longer tries to do something that bricks every single job.
/:cl:


DeltaFire15:
CI failing due to unrelated issues.

Kmc2000 (approved):

PR #1327, opened 2021-04-03
Pipe autism unleashed

Cyaneer:
## About The Pull Request

Comments out one-per-tile for some pipe components. This still leaves the ones intended for that intact, so no connector/machine funtimes.

## Why It's Good For The Game

Four **/** to make atmos compact.
Autistic atmos setup limits have been unleashed and increased by at least a factor of six. Also allows the building of compact one-tile airlocks.

## Changelog
:cl:
balance: Vents/scrubbers/passives/filters/mixers can now be built one-per-layer.
/:cl:

### DIFF
--- a/code/modules/atmospherics/machinery/components/trinary_devices/trinary_devices.dm
+++ b/code/modules/atmospherics/machinery/components/trinary_devices/trinary_devices.dm
@@ -5,7 +5,7 @@
 	use_power = IDLE_POWER_USE
 	device_type = TRINARY
 	layer = GAS_FILTER_LAYER
-	pipe_flags = PIPING_ONE_PER_TURF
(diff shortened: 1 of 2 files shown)


Bokkiewokkie:
I already think it's better to keep these restrictions _because_ encouraging more engie autism setups is bad, but right now putting a mixer or filter on top of another one stops the AI from being able to interact with it which is a very bad thing.

Karmic-Skink:
Yeah these restrictions are in place for valid reasons.

ta3370:
I have watched the Master Controller cry itself into a pit because somebody accidentally dual-layered *ONE* pipe on their map. This screams bad idea.

Cyaneer:
While the filter/mixers could be changed back to one per tile, if no-one else wants this I'll just close it.

Issue #303, opened 2020-01-30
Supermatter crystal starts at 1 EER and thus delams without attention


Vasily2013:
Saw this on other servers (Either TG or Bee, forgot which one) and it's an issue here. Reproduction: Load valk, sm monitor, and then wait for the kaboom.

francinum:
Must be verified before next grace period.

IndusRobot:
This issue has been inactive for long enough to be automatically marked as stale. If this was a bug report and hasn't been addressed yet, and is still a probelm, please don't hesitate to notify a maintainer.

PR #1329, opened 2021-04-04
Fixes supermatter starting at 1 power

covertcorvid:
## About The Pull Request
Makes the supermatter crystal have 0 power if it hasn't been activated yet. Also makes it not take damage on airless turfs. (Space turfs still hurt it.)
Fixes issue #303 (marked as an upstream issue but I don't feel like going through voting for this)

## Why It's Good For The Game
Supermatter crystals can just sit around without delaming if no one has done anything to them

## Changelog
:cl:
fix: Supermatter is no longer at 1 power after the first process_atmos
tweak: Supermatter does not take damage on airless non-space turfs
/:cl:


covertcorvid:
Update: warm air still causes damage

covertcorvid:
Further update: I don't know what was going on yesterday but no it does not. But I've made totally airless rooms stop the damage as well.

> 
> 
> Update: warm air still causes damage

code/modules/power/supermatter/supermatter.dm
```diff
@@ -355,10 +355,13 @@ GLOBAL_DATUM(main_supermatter_engine, /obj/machinery/power/supermatter_crystal)
 		removed = new()
 
 	damage_archived = damage
-	if(!removed || !removed.total_moles() || isspaceturf(T)) //we're in space or there is no gas to process
+
+	//NSV13 - airless areas don't cause damage
+	if(isspaceturf(T)) //we're in space
 		if(takes_damage)
```
Kmc2000, line 361: Could you not just do something like...

takes_damage = (isspaceturf(T) && removed && removed.total_moles()) ? takes_damage : FALSE

covertcorvid: The takes_damage flag is only there for the "hugbox" and fake crystals, if I go flipping it on and off I'll probably break those.

Kmc2000 (approved):

PR #1332, opened 2021-04-08
Resolves some airlock and parallax issues from the rebase

Kmc2000:
## About The Pull Request

* NSV airlock construction no longer spits out infinite default doors
* Removes floyd's stupid lavaland planet (except when youre not on a ship)
* Please don't look too closely at the code :))

## Why It's Good For The Game

Jank bad

## Changelog
:cl: Kmc2000
fix: NSV-specific airlocks can now be rebuilt!
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock.dm
+++ b/code/game/machinery/doors/airlock.dm
@@ -1407,6 +1407,8 @@
 		A.previous_assembly = previous_airlock
 		A.update_name()
 		A.update_icon()
+		//NSV13 - I'm lazy okay?
+		A.nsv_ify(src)
 
 		if(!disassembled)
 			if(A)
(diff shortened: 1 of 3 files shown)


code/game/machinery/doors/airlock.dm
```diff
@@ -1407,6 +1407,8 @@
 		A.previous_assembly = previous_airlock
 		A.update_name()
 		A.update_icon()
+		//NSV13 - I'm lazy okay?
+		A.nsv_ify(src)
```
GoatMonarch, line 1411: Did you just add a proc to convert a vanilla airlock to our airlocks?
Deadpan outlined

Kmc2000: :)))

covertcorvid: What if someone actually wants an old type of airlock?

Kmc2000: It has a type check to only apply to ours

PR #1333, opened 2021-04-09
Adds a hull foam extinguisher for mechs

benbot16:
## About The Pull Request

Adds a fairly simple mech-mounted hull foam extinguisher, similar to the portable and borg-mounted ones.

I also changed a bit of mech ext code so that you could theoretically make extinguishers that spray other stuff that's not water or hull foam, as well as supporting some other small stuff.

Although I did test this a few times, odds are since I'm rather new at this I'm missing something incredibly obvious, so if so please let me know. 

## Why It's Good For The Game

Allows utility mechs to be a bit more useful for external repairs (as well as spraying foam everywhere). 

## Changelog
:cl: Benbot16
add: Adds a hull foam extinguisher for utility mechs
code: Tweaked mecha fire extinguisher code a bit, so now it can theoretically support reagents other than water.
/:cl:


TerryTehCoder:
One step closer to loadable Exosuit Sprayers. I want to live in a world where you can get sprayed down with Chlorine Triflouride by someone riding a Ripley.

code/modules/research/designs/mechfabricator_designs.dm
```diff
@@ -520,6 +520,16 @@
 	construction_time = 100
 	category = list("Exosuit Equipment")
 
+//NSV-related change
```
Kmc2000, line 523: This is okay for now, but in future we'd usually make our own copy of the file under our subdirectory for ease of use

TheFakeElon:
There's already a modular file for mech designs, you shouldn't need to put it in the main

code/modules/research/designs/mechfabricator_designs.dm
benbot16: Think this is what you meant, if not let me know

Kmc2000 (approved):
This is a lot of code changes, consider doing this on upstream too to save hassle later down the line

PR #1341, opened 2021-04-10
Adds jetpack ugrade back to the mining vendor + Plasma cutter tweaks

bluey89962:
## About The Pull Request

Beebase merge removed the jetpack upgrade from the mining vendor, this PR adds it back.
Also reverts some plasma cutter nerfs, making them a tad speedier and use less plasma to recharge. They also come fully charged.

## Why It's Good For The Game

Bee mining is very different from NSV mining, the rebase was bound to have some changes that are unecessary in here. Jetpacks, while useless in Bee's Lavaland, are incredibly useful to fly around asteroids.
It also makes so Plasma Cutters are back to how they were. While Bee miners can easily find plasma in Lavaland, NSV miners rely heavily on luck to find materials for the arrestor and sensor upgrades so they can have a steady supply of plasma. If a miner has an unlucky day and doesn't manage to find them, plasma cutters end up being more of a liability than an useful tool by consuming more plasma than they can mine. 
In short, Bee has to worry about miners going for gamer loot and ignoring their job, this doesn't happen in NSV, thus some changes aren't needed. 

## Changelog
:cl:
add: Jetpack upgrade is back on the mining vendor
balance: Plasma cutters shoot a tad faster
balance: Plasma cutters don't consume an obscene ammount of plasma anymore
balance: Plasma cutters are charged when created
/:cl:

### DIFF
--- a/code/modules/mining/machine_vending.dm
+++ b/code/modules/mining/machine_vending.dm
@@ -47,6 +47,7 @@
 		new /datum/data/mining_equipment("P-KA Cosmetic Hyper Chassis",	/obj/item/borg/upgrade/modkit/chassis_mod/orange,					200),
 		new /datum/data/mining_equipment("Mining Hardsuit",				/obj/item/clothing/suit/space/hardsuit/mining,						2000),
 		new /datum/data/mining_equipment("Expanded E. Oxygen Tank",		/obj/item/tank/internals/emergency_oxygen/engi,						1000),
+		new /datum/data/mining_equipment("Hardsuit Jetpack Upgrade", 	/obj/item/tank/jetpack/suit,										2000),
 		new /datum/data/mining_equipment("Mining Conscription Kit",		/obj/item/storage/backpack/duffelbag/mining_conscript,				1000),
 		new /datum/data/mining_equipment("Survival Knife",				/obj/item/kitchen/knife/combat/survival,							1000),
 		new /datum/data/mining_equipment("Tracking Implant Kit", 		/obj/item/storage/box/minertracker,									1000),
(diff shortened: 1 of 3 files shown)


GoatMonarch:
Good stuff, but you'll want to add a //NSV13 (What change you did) within your code. Otherwise you'll loose this next time we pull from upstream!

bluey89962:
> 
> 
> Good stuff, but you'll want to add a //NSV13 (What change you did) within your code. Otherwise you'll loose this next time we pull from upstream!

would I add that in front of the lines that I changed?

GoatMonarch:
> would I add that in front of the lines that I changed?
If able to, right by the change. 
Though there's of course lenghty lists you can't exactly comment inside of, so just comment by the start of said paragraph. As long it's commented, the better.

bluey89962:
> 
> 
> > would I add that in front of the lines that I changed?
> 
> If able to, right by the change.
> Though there's of course lenghty lists you can't exactly comment inside of, so just comment by the start of said paragraph. As long it's commented, the better.

alright hopefully I did it correctly

Karmic-Skink:
> Personally I think these buffs are unnecessary. You can still charge these guns with plasma, the plasma you do for doing your job. And miners already double wield the plasma cutters. Lowering the delay even further borders a bit on powergaming

This.
10x the charge rate is excessive.
Also starting charged just reeks of utter laziness.

Kmc2000 (approved):

Kmc2000:
Code looks alright but I'll want to speak to some more people on the balance side first

bluey89962:
> > Personally I think these buffs are unnecessary. You can still charge these guns with plasma, the plasma you do for doing your job. And miners already double wield the plasma cutters. Lowering the delay even further borders a bit on powergaming
> 
> 
> 
> This.
> 
> 10x the charge rate is excessive.
> 
> Also starting charged just reeks of utter laziness.

I can definitely see the charge rate and shooting rate being changed considering all I did was revert it to its original value without giving them much thought, however miners On NSV can only dream of having as much plasma as Bee from the get-go. Plasma cutters are basically useless until they get lucky enough to find the materials needed for the Roci upgrades. While in some rounds you manage to find everything you need in the first asteroid, that isn't always the case. I've personally had many rounds where due to structures spawning in ferrous asteroids you don't get anything other than the main cluster (which is another thing I believe could have some tweaks, although I'm not experienced enough in coding BYOND at this point in time).

Some people already complain about miners not getting resources sometimes, I do not see how slowing them down because of a common issue on bee - that being miners ignoring their job to hunt for loot - would be fair, especially now that the BS Miner is gone and small crews have to rely on someone volunteering to do that job.

PR #1355, opened 2021-04-17
Fixes Radstorms triggering like crazy

DeltaFire15:
## About The Pull Request
What if I told you someone forgot about inheritance when adding the radstorm system event, so there was two radstorm events in the event control, both of which had weight ten, except with one being able to occur alot of times?
Well, this sets the weight of the event that should only be triggered by radioactive systems to 0, resolving the madness that is radstorm-mania.

## Why It's Good For The Game
The air begins to grow warm!

## Changelog
:cl:
fix: Radstorms should no longer be occuring FAR more often than intended.
/:cl:


Kmc2000:
feck.

Kmc2000 (approved):

PR #1363, opened 2021-04-20
Deckgun QOL improvement

TheFakeElon:
## About The Pull Request

TGUI interface is now a lot more reactive as to what's going on. You can no longer load the deck gun with a shell that doesn't exist, I'm guessing this was causing runtimes too.

The "Feed" button renames itself to "Eject" once a shell is inside of it and the loader is greyed out if there's no shell fed into it

## Why It's Good For The Game

QOL + removes unused variables

## Changelog
:cl:
tweak: deckgun UI is now more representative as to what's going on 
fix: fixes empty deckgun loading
/:cl:


TheFakeElon:
Why have tgui bundle conflict detection when it's rebuilt on merge 🤔

Kmc2000 (approved):

PR #1368, opened 2021-04-21
[Experimental] Lowpop PVP!

Kmc2000:
## About The Pull Request

This PR adds in support for PVP at a lower playercount than previously possible! Also included is PVP map rotations (with pop settings). The original iteration of PVP was designed around 30pop, so this should still hold.

image

## Why It's Good For The Game

PVP's definitely fun, but getting 50 people on all at once to run it is a real task. The intent of this is to have a less intense lowpop option for PVP so it can be transitioned into secret properly, as originally intended. We'll still keep the 50pop PVP madness in our back pocket though...

## Changelog
:cl: Kmc2000
fix: Re-added the dark-astraeus
add: PVP is now supported for 30 players and up!
/:cl:


Pontenerd:
> hammerbird bad


lies

PR #1369, opened 2021-04-22
[Ready] Gladius Update

SerynEngi:
## About The Pull Request

This fixes the cargo shuttle, the vertical double doors and a few other minor issues located throughout the map.

## Why It's Good For The Game

There are people who like the map, it should now be able to be added back into the rotation.

## Changelog
:cl:
fix: Gladius not broken.
/:cl:


Karmic-Skink:
CL should say fix, not tweak.

SerynEngi:
I am not at all sure why it doesn't like this, but from the mapping perspective here, there is nothing that I can see that should be causing any sort of runtime. 

This is ready from my perspective and ability to fix, however, as the core issues preventing the map from being used have been fixed. 

If there is another mapping thing that could be done to bypass the issue, I will do so.

Kmc2000 (approved):

PR #1375, opened 2021-04-24
Free mission points

IndusRobot:
## About The Pull Request

Completing missions now awards a small amount of faction points

## Why It's Good For The Game

Ties missions into patrols and encourages doing more missions for that sweet gameplay variety 

I was going to add more missions but cargo freight torpedoes stopped working a long time ago apparently and I didn't feel like also fixing torp collisions 

## Changelog
:cl:
add: Completing missions now awards 100 faction points, which count toward patrol completion 
tweak: Ships may only have 3 uncompleted missions at a time
/:cl:

### DIFF
--- a/nsv13/code/datums/missions.dm
+++ b/nsv13/code/datums/missions.dm
@@ -69,9 +69,9 @@ NOTE:
 
 /datum/nsv_mission/proc/payout()
   var/ownerFaction = owner.faction
-  for(var/datum/faction/NT in SSstar_system.factions)
-    if( ownerFaction == lowertext( NT.name ) )
-      NT.gain_influence(ticket_bounty)
+  for(var/datum/faction/F in SSstar_system.factions)
+    if( ownerFaction == lowertext( F.name ) )
+      F.gain_influence(ticket_bounty)
   var/obj/item/holochip/money = owner.send_supplypod(/obj/item/holochip)
   money.credits = payout
   if(LAZYLEN(rewards))
--- a/nsv13/code/datums/missions.dm
+++ b/nsv13/code/datums/missions.dm
@@ -18,6 +18,7 @@ NOTE:
   var/list/valid_factions = list("nanotrasen")
 
   var/payout = 4000
+  var/ticket_bounty = 100
   var/list/rewards // List of items to be offered as a reward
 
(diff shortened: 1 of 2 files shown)


Karmic-Skink:
-rep
Because he asked me to.

nsv13/code/modules/overmap/traders.dm
```diff
@@ -242,7 +242,22 @@
 			return
 		attempt_purchase(target, usr)
 	if(action == "mission")
-		give_mission(usr)
+		var/list/currentMissions = list()
+		for ( var/datum/nsv_mission/M in SSstar_system.all_missions )
```
Kmc2000, line 246: This will be weird on PVP. If, say, the astraeus crew cops all the 3 missions globally before NT

Cool idea though!

nsv13/code/datums/missions.dm
```diff
@@ -67,6 +68,10 @@ NOTE:
     return FALSE // If the mission is not active, it can not be completed
 
 /datum/nsv_mission/proc/payout()
+  var/ownerFaction = owner.faction
+  for(var/datum/faction/NT in SSstar_system.factions)
+    if( ownerFaction == lowertext( NT.name ) )
```
Kmc2000, line 73: This would mean that on PVP, syndicates completing missions would benefit Nanotrasen. More stringent checking's required here, why not look at the valid_factions list for inspo?

Kmc2000 (changes requested):
Thanks for the submission! Just a few things...

nsv13/code/datums/missions.dm
IndusRobot: I think `var/datum/faction/NT` is a leftover typo, it should be pulling the faction of the overmap ship that it belongs to, then awarding those points. Although I never actually tested if that's the case, so I'll do that

nsv13/code/modules/overmap/traders.dm
IndusRobot: I'll add an additional check to also verify the active mission's owner, and to add to the `currentMissions` list only if the mission owner is the ship trying to get new missions

Kmc2000 (approved):

PR #1401, opened 2021-05-09
Makes unanchored windows and falsewalls show up

### LINKED ISSUE #784: Non anchored windows and fake walls are invisible
Fake walls are only invisible when closed

## Reproduction:
Windows:
1. Use glass in hand to make a window
2. See that it's invisible
Fake walls:
1. Place metal on unanchored grider
2. See that it is invisible when closed

covertcorvid:
## About The Pull Request
When you start building a window, it's actually visible before you screw it down. Falsewalls are visible when built and smooth with normal walls.
Fixes #784

## Why It's Good For The Game
They're not supposed to be invisible

## Changelog
:cl:
fix: windows no longer invisible when first built
fix: falsewalls no longer invisible and smooth with regular walls
code: moved a bunch of stuff from custom_turfs into their own files
code: deletes the old railings that we're not using
/:cl:


Kmc2000 (approved):
Entrapped

Kmc2000 (approved):

PR #1408, opened 2021-05-15
Probably fixes munitions trolleys

DeltaFire15:
## About The Pull Request
Title. A != A.type which is what should be used here.

## Why It's Good For The Game
Trolley man.. good?

## Changelog
:cl:
fix: Munitions trolleys are no longer one-use.
/:cl:


covertcorvid (approved):
mirrors how stuff gets added, probably works

PR #1418, opened 2021-05-21
[TESTMERGE-READY] Multi-Z improvements

covertcorvid:
## About The Pull Request
Trying to make stairs work better
https://streamable.com/0o0a3j

## Why It's Good For The Game
You can bring things with you when you use stairs

## Changelog
:cl:
fix: various interactions with stairs, including pulling, pushing, riding, and carrying
fix: falling from the ceiling while buckled to something no longer stops you from taking damage
fix: fixed a runtime when something fell onto a tile with no objects in it
/:cl:


GoatMonarch:
Great stuff, though I'm wondering how clown cars will interact with this? And when you fall down a floor with a vechicle, will the vechicle also take damage to it's integrity?

Kmc2000:
earth shatteringly based, thanks!

covertcorvid:
> Great stuff, though I'm wondering how clown cars will interact with this? And when you fall down a floor with a vechicle, will the vechicle also take damage to it's integrity?

I haven't done any testing with cars yet, but it is on my agenda. If the vehicle took fall damage before, it still will. Previously the vehicle absorbed the impact and the rider was unharmed, but I only changed that for "ridden" vehicles like wheelchairs and scooters, so the car probably still protects its occupants.

covertcorvid:
> 
> 
> Great stuff, though I'm wondering how clown cars will interact with this? And when you fall down a floor with a vechicle, will the vechicle also take damage to it's integrity?

This does allow clown cars to travel up and down the stairs, and they seem to keep all the mobs in them, but they always crash when going up

Kmc2000 (approved):

PR #1428, opened 2021-05-30
Updates fleet youtube-dl to match regular youtube-dl

Kmc2000:
## Why It's Good For The Game

Duplicate ytdl instancing

## Changelog
:cl: Kmc2000
fix: Fleet combat music should work better now
/:cl:


francinum:
This shouldn't be the cause of it.

Vivalas:
I'm somewhat convinced that the issues with the internet sound stuff have to do with copyright-- I think youtube might screw with the streaming in a way for copyrighted stuff if it's classified as music or has a certain number of views, for legal reasons.

I say this because I've 'fixed' the issue on a few occasions by just searching by views ascending and choosing versions of the music I want to play that have less views

PR #1429, opened 2021-06-01
Replaces most mentions of HoP with XO

BootlegBow:
## About The Pull Request

Title. This affects things like names and descs for clothing and stuff. Also looking at the code, it seems like the XO could have previously rolled antags like cult, but I'm not marking this as a fix because I'm not sure.
Oh by the way, the MAA might be able to roll quite a few antags they shouldn't be able to roll. Might wanna check that.

## Why It's Good For The Game

How the hell can he be the HoP's beloved corgi when there is no HoP?

## Changelog
:cl:
spellcheck: replaced most mentions of HoP with XO
/:cl:


Karmic-Skink (changes requested):
Every single instance where you touch a base file with a change must include an //NSV13 tag for tracking.

BootlegBow:
gotta love it when an entirely seperate line of code breaks for no fucking reason

PR #1442, opened 2021-06-09
FTL drive rework

TheFakeElon:
## About The Pull Request

FTL drives are now a lot more complex than a magic computer with a funny dial. The drive now consists of three different main components

<strike>**Maintainer note:**
Changes extools to use dynamic gas limit, the source is on my extools fork, might be worth making an NSV clone as we now differ from bee. I tried to get it merged there but no success. Auxtools should be coming soon anyway</strike>

### FTL silo
aka "Nucleium-Plasma Frame Refiner"
This large refinery converts plasma directly into nucleium, conversion efficiency is determined by the amount of power currently being input into the silo. Make sure you manage the internal pressure, containment failure is devastating. This is a more convenient alternative to using the storm drive for FTL fuel production, albeit at a lower efficiency

### FTL pylon
aka "Drive pylon"
Has fast spinny gyros that is spooled with nucleium and a moderate amount of power, drive manifolds can have multiple connected pylons, each one increasing the total FTL charge speed. 
Once the pylon is fully spun up, it will no longer consume nucelium (subject to change) and will gradually (but exponentially) require more power to keep spooled, so make sure to only have these spun up when you need them.

### FTL drive
aka "FTL manifold"
The heart of it all, powered by drive manifolds. This also acts as the drive pylon control computer.

## Why It's Good For The Game

The current FTL computer is boring and a dent in engineer masochism.

## Changelog
:cl:
add: Adds a new FTL drive and it's respective components
code: cleans up legacy FTL drive file
code: Modularizes NSV gases 
/:cl:

- [x] FTL Silo
- [x] FTL Pylon
- [x] FTL Manifold
- [x] Frameshifted Plasma
- [x] New manifold UI
- [x] Silo UI 
- [ ] Test everything
- [ ] Make legacy compatibility not bodge ((optional))
- [x] Clean up unused procs/vars


MacBlaze1:
As an engineer who doesn't play, this sounds fantastic. Thinking of an engineer who _does_ play I would have to say engineers might not be able to handle all of this with the current pop. Anyway keep up the good work.

TheFakeElon:
> What happens if frameshifted plasma goes back into the stormdrive or funny chernobyl reactor?

Oo, Totally forgot about those. I'll add something fun

TheFakeElon:
> Thinking of an engineer who does play I would have to say engineers might not be able to handle all of this with the current pop

It shouldn't be too hard, the only thing they have to keep an eye on is the silo pressure, which _shouldn't_ be tedious unless someone accidently left it on or they're pumping spicy gasses in

FLME:
I feel like this may increase FTL spooling times - which isn't that bad except on Aether where you start at _sol_.

Also wondering how this will effect Eclipse?

code/modules/atmospherics/gasmixtures/gas_types.dm
```diff
@@ -147,15 +147,13 @@ GLOBAL_LIST_INIT(hardcoded_gases, list(/datum/gas/oxygen, /datum/gas/nitrogen, /
 	fusion_power = -10
 	rarity = 200
 
-/* NSV13 - I NEED THAT DATUM SLOT, NOT SORRY
```
Karmic-Skink, line 150: We are keeping miasma disabled.

Karmic-Skink (changes requested):
I'm aware that we no longer need the slot. But we need the performance.

Bokkiewokkie:
> I feel like this may increase FTL spooling times - which isn't that bad except on Aether where you start at _sol_.

A PR to move the Aetherwhisp to Argo is currently open, so that'll likely not be the case anymore when this gets added.

ilikesaying:
what about the miner ship? miners may not know how to use the new ftl drive or even watch the drive and depending on the power consumption, solars and pacmans may not be enough for long trips.

TheFakeElon:
The old FTL drive is not being removed. Not yet at least, it's a royal pain having to make all systems compatible with both

Karmic-Skink:
Honestly at this stage, its definitely a -1 from me.

I feel it is unnecessarily complicating the FTL drive and adding an additional knowledge/skill floor to an already reasonably demanding job and very necessary job. Engineering in general can be daunting as we already heap on extra requirements for overmap related gameplay in addition to potentially round ending reactors.

The inclusion of an additional gas on top of Nucleium isn't really required, given it just adds extra overhead as we now need to consider and support that gas with other affecting devices and mobs (Note: I very much disagree with the proposed SD interaction). 
Nucleium already covers the role of special NSV13 derivative gas, while Constricted Plasma covers the non-gas interaction plasma side, we don't really need an extra one.

nsv13/code/modules/overmap/starmap.dm
```diff
@@ -74,10 +74,17 @@
 	var/datum/star_system/current_system = info["current_system"]
 	SSstar_system.update_pos(linked)
 	if(linked.ftl_drive)
-		data["ftl_progress"] = linked.ftl_drive.progress
-		if(linked.ftl_drive.ftl_state == FTL_STATE_READY)
-			data["ftl_progress"] = linked.ftl_drive.spoolup_time
-		data["ftl_goal"] = linked.ftl_drive.spoolup_time //TODO
+		if(istype(linked.ftl_drive))
```
Kmc2000, line 77: This is actually unecessary, as long as both objects have the same vars, you don't need to worry about typecasting at all. Look at my new shield component in boarding for an example of this beautiful hot garbage.

TheFakeElon: @Kmc2000 I'm aware, I've done that for most ftl drive interactions but I needed an exception for this because they both use a different spool system. I've changed it up a bit since then so I might be able to remove this line, we'll see 😆

TheFakeElon:
image

TheFakeElon:
Dropping dynamic gas extools for 514 compatibility. I'll see if I can integrate it in another PR

nsv13/code/modules/atmospherics/gasmixtures/gas_types.dm
```diff
@@ -0,0 +1,26 @@
...
+/datum/gas/nucleium
+	id = "nucleium"
+	specific_heat = 850
+	name = "Nucleium"
+	gas_overlay = "nucleium"
+	moles_visible = MOLES_GAS_VISIBLE
+	dangerous = TRUE
+	rarity = 2000
+
+/datum/gas/frameshifted_plasma
```
Karmic-Skink, line 19: As previously discussed, please do not add an additional gas datum purely for this device.

We have Nucleium for this express purpose.

Karmic-Skink (changes requested):

TheFakeElon:
"I should have it finished by the end of the month"
- TheFakeElon, June 2021

TheFakeElon:
I'll re-open when it's finished:tm:

PR #1443, opened 2021-06-11
"fastmos"

Kmc2000:
## Why It's Good For The Game

What's more important, atmos sim or being able to move?

## Changelog
:cl:
tweak: Atmospherics processing slows down when the ship is in heavy combat.
/:cl:

### DIFF
--- a/code/controllers/subsystem/air.dm
+++ b/code/controllers/subsystem/air.dm
@@ -84,6 +84,18 @@ SUBSYSTEM_DEF(air)
 /datum/controller/subsystem/air/fire(resumed = 0)
 	var/timer = TICK_USAGE_REAL
 
+	var/TiDi = round((Master.tickdrift/(world.time/world.tick_lag))*100,0.1)
+	switch(TiDi)
+		if(0 to 15)
+			wait = 5 //This isn't an initial() call because that could get nasty with how much this thing runs.
+		//Fuck you I want my ships to fly
+		if(15 to 20)
+			wait = 2 SECONDS
+		if(20 to 30)
+			wait = 5 SECONDS
+		if(30 to INFINITY)
+			wait = 30 SECONDS
+
 	if(currentpart == SSAIR_REBUILD_PIPENETS)
 		var/list/pipenet_rebuilds = pipenets_needing_rebuilt
 		for(var/obj/machinery/atmospherics/AT as() in pipenet_rebuilds)


DeltaFire15:
I think one of the main issues with atmos breaches is upper deck breaches causing a chain of 'thing gets sucked out through to the top deck -> thing falls down since it's openspace -> repeat forever' (aka, spinny tile of doom)
Might be worth to try disabling cross-z pulling and see if that helps.

Kmc2000:
> I think one of the main issues with atmos breaches is upper deck breaches causing a chain of 'thing gets sucked out through to the top deck -> thing falls down since it's openspace -> repeat forever' (aka, spinny tile of doom)
> Might be worth to try disabling cross-z atmos pulling and see if that helps.

Oh yeah this is just the heavy handed bruteforce way of doing that

SerynEngi:
> I think one of the main issues with atmos breaches is upper deck breaches causing a chain of 'thing gets sucked out through to the top deck -> thing falls down since it's openspace -> repeat forever' (aka, spinny tile of doom)
> Might be worth to try disabling cross-z atmos pulling and see if that helps.

This would also have the perk of making the job of engineers much simpler on maps with huge topdecks like the Galactica or Tycoon.

DeltaFire15:
Yeah, 'not dragging objects up' is what I meant, not 'ceiling holes are no more'

TerryTehCoder:
> Yeah, 'not dragging objects up' is what I meant, not 'ceiling holes are no more'

I did always find this weird, that when objects are sucked out in decompression, they just stay there at the top(and then fall down, since like openspace. But mechanically, there's nowhere else for them to go. 
But, shouldn't decompression launch stuff off into space if it's strong enough? Not just on top of the hull. 

It's just sort of the way Z-levels are done, and excessive work for something tiny like this, but I think being forcefully ejected by decompression up a deck and into space has intense comedic value. But not something that ever works in a proper gameplay loop.

DeltaFire15:
Also this has been mentioned in discord but to leave it here for easier access, but
Just slowing down SSair like this PR does would cause major issues with a bunch of things like both the main engines (SD and AGCNR), due to them ticking on SSmachinery and worsifying the phenomenon of 'hey the engine stopped working because the gas doesn't move in its pipes anymore

TheFakeElon:
prob(80)

code/controllers/subsystem/air.dm
```diff
@@ -84,6 +84,18 @@ SUBSYSTEM_DEF(air)
 /datum/controller/subsystem/air/fire(resumed = 0)
 	var/timer = TICK_USAGE_REAL
 
+	var/TiDi = round((Master.tickdrift/(world.time/world.tick_lag))*100,0.1)
+	switch(TiDi)
+		if(0 to 15)
+			wait = 5 //This isn't an initial() call because that could get nasty with how much this thing runs.
```
TheFakeElon, line 90: Might be worth making this a define. Could even have the other wait times as multiplications of the define if you're a gamer 😎

covertcorvid:
Closing due to engine concerns.

PR #1448, opened 2021-06-14
[WIP] system gas miners

DeltaFire15:
## About The Pull Request
This adds system gas miners, which when added to something with an overmap (reads: ships) will harvest gasses present in their system. Also, systems can now have harvestable gasses, depending on their system type, randomness and even more rng.
Intended to replace 'infinite' gas miners, being balanced enough to be added to ships without close-to-infinite power stuff like infinite max throughput burnchambers.

This is marked as WIP for now since I'll be waiting for #1438 before I'm touching that UI to show gas amounts after scanning.

!!This is merely the code part, actually adding them to ships is at mappers discretion after this is merged, I don't want to conflict with map PRs that may possibly be worked on.

- [x] System gas miner code in general
- [x] Systems get randomly set minable gas amounts
- - [ ] Look at some specific systems and hardset their gasses, e.g. Sol always gets oxygen / nitrogen
- [ ] Astrometrics console scanning shows gas amounts in systems
- [ ] Configurable miner that can be constructed (major power drain and slower extraction), gas being extracted can be switched  (gas harvester roci?!)
-  [ ] Maybe: Destroyed ships leave trace amounts of plasma / oxygen / nitrogen 

## Why It's Good For The Game
Compromise between the 'gas miners that are infinite and ruin any kind of attempted gas balance' and 'oh boy this tank got breached, time to order 100 cans from cargo' extremes. Might also give some sense to exploring to refuel whilst doing other things.

## Changelog
:cl:
add: System gas miners, which harvest gasses present in the system the ship they're installed on is in.

/:cl:

### DIFF
--- a/nsv13/code/modules/atmospherics/system_miners.dm
+++ b/nsv13/code/modules/atmospherics/system_miners.dm
@@ -1,0 +1,95 @@
+/*
+ * File for system miners. As opposed to being infinite, these get their gasses from accessing a star system's gas resources.
+ * systems outside the commonly traversed areas generally are the main source for these, as they are usually pretty mined out otherwise.
+*/
+
+/obj/machinery/atmospherics/miner/system
+    name = "long-range gas collector"
+    desc = "A marvel of engineering, these devices use highly complicated processes to collect gasses over big distances."
+    power_draw = GASMINER_POWER_STATIC
+    var/obj/structure/overmap/attached_overmap = null
+    spawn_mol = 200 //200 moles mined per tick, if there is enough.
+
+/obj/machinery/atmospherics/miner/system/Initialize()
+    . = ..()
+    return INITIALIZE_HINT_LATELOAD
+
+/obj/machinery/atmospherics/miner/system/LateInitialize()
+    . = ..()
+    attached_overmap = get_overmap()
+
+/obj/machinery/atmospherics/miner/system/check_operation()
+    if(!attached_overmap)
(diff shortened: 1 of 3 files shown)


BootlegBow:
finally we can repressurize the eclipse

Karmic-Skink:
As discussed briefly via DMs; consider holding off on this for the upcoming mining rework as gas mining via Sabres could definitely be a thing.

DeltaFire15:
> As discussed briefly via DMs; consider holding off on this for the upcoming mining rework as gas mining via Sabres could definitely be a thing.

This'll merely be the code part when done.
Placing these, whether on sabres, the mining vessel or the main ship should all be doable, as is tweaking values like speed. Same applies for the "Configurable Miner" part, whether it'll be possible to make via techweb or merely preplaced in sabres or w/e.

nsv13/code/modules/atmospherics/system_miners.dm
```diff
@@ -0,0 +1,95 @@
...
+    merger.set_moles(spawn_id, extracting)
+    merger.set_temperature(spawn_temp)
+    O.assume_air(merger)
+    O.air_update_turf(TRUE)
+
+/obj/machinery/atmospherics/miner/system/examine(mob/user)
+    . = ..()
+    if(!spawn_id)
+        return
+    if(!attached_overmap || !attached_overmap.current_system)
```
TheFakeElon, line 56:
```suggestion
    if(!attached_overmap?.current_system)
```

PR #1458, opened 2021-06-20
Optionally ID-based lift buttons

covertcorvid:
## About The Pull Request
Allows mappers to optionally assign a unique ID to a lift button and lift, which will still link if the two are in different areas. Defaults to the standard area linking if the id var is not set.
The mapper is responsible for making sure the IDs are unique and matching.

## Why It's Good For The Game
More flexible button placement

## Changelog
:cl:
code: Let mappers link lazylift buttons by ID instead of area if desired
/:cl:


nsv13/code/modules/turbolift/lazylift.dm
```diff
@@ -54,8 +55,25 @@ That's it, ok bye!
...
+				continue
+			else if(LL.id == lift_id)
+				lift = LL
+	else
+		for(var/turf/T in get_area(src))
+			if(T.z != z)
+				continue
+			var/obj/machinery/lazylift/LL = locate(/obj/machinery/lazylift) in T
+			if(LL)
+				lift = LL
```
TheFakeElon, line 75: You can save a bunch of performance by locating the lazylift in the area, rather than each turf

covertcorvid: Any way to locate multiple items? There should be 2 of these in the area, and I need the one that's on the same z level as the button.

TheFakeElon: There sure is :)
```suggestion
		for(var/obj/machinery/lazylift/LL in get_area(src))
			if(LL.z == z)
				lift = LL
```

Karmic-Skink (approved):

PR #1481, opened 2021-06-24
Fixes infinite money caused by cheap crates

RemervKeberb:
## About The Pull Request

Tweaks the cost of crates to prevent people from ordering cheap crates and returning them to CC with a stamped manifest. Giving them more money than they started with on top of the ordered items. Also adds extra items to tweaked price crates where it made sense to

## Why It's Good For The Game

Infinite money punjabi 360p no virus glitch is not epic to have. There's always going to be "that guy" who tries to game the system
## Changelog

tweak: tweaked cargo prices and contents
fixes: #1461

### DIFF
--- a/code/modules/cargo/packs.dm
+++ b/code/modules/cargo/packs.dm
@@ -126,17 +126,23 @@
 /datum/supply_pack/emergency/firefighting
 	name = "Firefighting Crate"
 	desc = "Only you can prevent station fires. Partner up with two firefighter suits, gas masks, flashlights, large oxygen tanks, extinguishers, and hardhats!"
-	cost = 600
+	cost = 800 //NSV edit, upped price to avoid infinite money. Added one extra of each as compensation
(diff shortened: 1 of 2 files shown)


Bokkiewokkie:
Can you also make sure this is fixed on upstream (bee)?

RemervKeberb:
Added the fix thing and uh, how would I go about checking that first part? Still very very new to this.

Karmic-Skink:
Alright, the //NSV13 tags are a little excessive but whatever.

Changing the price is fine, adding all the extra stuff isn't really needed.

RemervKeberb:
Just added all the extra stuff because I didn't wanna change anything up too much. All the crates should still have the same cost per item. (At least that was my intention while making it.)

RemervKeberb:
Yup yup, did that already.

covertcorvid (approved):

PR #1482, opened 2021-06-24
Fixes armour pumps breaking if multitooled with an invalid buffer

### LINKED ISSUE #1306: Armour pump to well linkage doesn't work sometimes
## Round ID:
NSV Round 2080 on Atlas map

## Reproduction:

This issue most likely occurs when switching armour pumps linked to armour wells around.
Normally armour pumps can be linked to the Armour well with a multitool, however in this round that didn't work after while. Instead you only hit them even with a buffer of the well loaded.  One still worked normally but the other three didn't.
Purging the well or opening the pumps open and closed didn't do anything either. 
Newly bought ones worked as expected though.

DeltaFire15:
## About The Pull Request
Title. If the buffer isn't a nanowell, it should just do nothing and not set it to said invalid thing and runtime & cry, This fixes that.
Fixes #1306 (probably)

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Multitooling an armour pump no longer breaks it if the buffer isn't an armour well.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/armour/nano_pump.dm
+++ b/nsv13/code/modules/overmap/armour/nano_pump.dm
@@ -125,12 +125,12 @@
 	. = FALSE
 	if(!multitool_check_buffer(user, tool))
 		return
-	apnw?.apnp -= src
 	var/obj/item/multitool/M = tool
-	if(!isnull(M.buffer))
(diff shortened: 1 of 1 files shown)


FLME:
fix man good

covertcorvid (approved):

PR #1485, opened 2021-06-25
Fixes the starsystem manager bluescreening

DeltaFire15:
## About The Pull Request
Title. Its system_type data needed to be adjusted to work correctly. Also makes the system_type var of starsystems a tiiny bit safer.

## Why It's Good For The Game
Fix man good

## Changelog
:cl:
fix: The starsystem management UI should no longer bluescreen.
/:cl:


TheFakeElon:
Wait how does this happen 👀. Systems should always be loaded because they're made during subsystem initialization right?

Bokkiewokkie:
> Wait how does this happen 👀. Systems should always be loaded because they're made during subsystem initialization right?

If an admin decides to press the button during setup in the lobby, at which point it hasn't initialized yet.

covertcorvid (approved):

PR #1488, opened 2021-06-25
Changes the character slot limit from 3 to 14

RemervKeberb:
## About The Pull Request

Does what it says on the tin, 
More character slots

## Why It's Good For The Game

Having the ability to play more characters is nice. Also would be pretty necessary in the event we decide to mess around with FTL like races. 

## Changelog

tweak: Tweaks the preferences.dm file to allow for more character slots

### DIFF
--- a/code/modules/client/preferences.dm
+++ b/code/modules/client/preferences.dm
@@ -5,7 +5,7 @@ GLOBAL_LIST_EMPTY(preferences_datums)
 	//doohickeys for savefiles
 	var/path
 	var/default_slot = 1				//Holder so it doesn't default to slot 1, rather the last one used
-	var/max_save_slots = 3
+	var/max_save_slots = 14
 
 	//non-preference stuff
 	var/muted = 0


Karmic-Skink:
There are currently ingame methods for acquiring more slots, etc

DeltaFire15:
Remember to also update (or remove) the byond member slot count modifier (line 154), because while it'd be sort of funny to have byond members have less slots than normal players, it'd definitely not be a good thing.

RemervKeberb:
Ah nah don't worry about that this is just me getting back at byond premium members for flexing their fancy ghosts and coloured names. All perfectly intentional and not at all fixed in the latest commit. I am definitely NOT thanking you for pointing it out in act I am cursing you for outing my plans!

covertcorvid:
4 or 5 might be reasonable, but 14 is kind of excessive. We don't plan on adding a ton of races, since even the ones we already have are mostly weird flavors of human-equivalent.

RemervKeberb:
When I made it I just chose the number that looked the best with the TGUI

covertcorvid:
Not currently planning on a huge amount of races, and extra slots can be purchased.

PR #1422, opened 2021-05-26
Skynet 3 - Boarder ai

Kmc2000:
## About The Pull Request

This is a mistake

## Changelog
:cl: Kmc2000
add: Added elite AI controlled boarder units, capable of intelligently reacting to combat!
/:cl:

PR #1503, opened 2021-06-28
Skynet 3.5 Boarder AI but it's updated with our latest code 

IndusRobot:
## About The Pull Request

This is a new branch off of #1422, the only change was moving a small chunk of code out of `ai-skynet2.dm` (now removed), and adding a previously missing variable `required_trait` to `/datum/ai_goal`

## Why It's Good For The Game

Keeping sec engaged and crew on their toes during deadpop 

🆑 Kmc2000
add: Added elite AI controlled boarder units, capable of intelligently reacting to combat!
/🆑


DeltaFire15:
This readds skynet2.dm which was entirely moved to skynet.dm, so probably delete that again.
Additionally, considering none of the AI goals of knpcs actually use AI traits (the var is null for all of them), may aswell leave it out, given that the functionality of it got moved to ai_flags (/ replace the trait check in the human AI goal check_score with the new version)

DeltaFire15:
> @DeltaFire15 I haven't used bitflags before, is this setup right?

Honestly, the more I look at the code for knpc ai goals the more I think they shouldn't be a subtype of normal ai goals given that they both don't land in the same global list aswell as them using completely different traits.

covertcorvid (approved):

PR #1515, opened 2021-07-03
Makes trader stations immediately stop processing if they are created in a non-loaded system

DeltaFire15:
## About The Pull Request
Title. Basically, traders get created processing. If you, say, enter their system, later leave it, the handler for leaving systems sees the trader and makes them stop processing since they no longer have a z. Buut, if you never even enter their z, they'll just keep wasting processing trying to figure out what they're doing moving in nullspace.
This makes them terminate their processing if they are created in a non-loaded system, as other cases should already be handled by jump / starsystem code.

## Why It's Good For The Game
Free microoptimisations good.

## Changelog
:cl:
code: Traders now immediately stop processing if in a unloaded system instead of waiting for a z-level possessing ship to enter and later leave.
/:cl:


DeltaFire15:
CI failing due to unrelated error that is fixed in a saperate PR of mine.

Bokkiewokkie (dismissed):

nsv13/code/modules/overmap/traders_items.dm
```diff
@@ -49,6 +49,11 @@
 	qdel(inhabited_trader)
 	. = ..()
 
+/obj/structure/overmap/trader/process()
```
TheFakeElon, line 52: Surely there's a better way to check without firing the check every process tick

DeltaFire15: minor but fair enough. Moved to LateInitialize which won't catch admins doing stuff but works otherwise.

Bokkiewokkie (approved):

PR #1523, opened 2021-07-07
Bambi 2: Ultimatum 

Karmic-Skink:
## About The Pull Request
This PR introduces a third OM armour type.
All ships (excluding fighters) are immune to anti-projectile weaponry.

## Why It's Good For The Game
Better refinement of the overmap gameplay.

## Changelog
:cl:
add: Added "overmap_medium" armour type
balance: Re-balanced all overmap entity armour values
balance: Moved a few weapon types into "overmap_medium"
add: Added consequences for ignoring nukes
/:cl:


Bokkiewokkie (dismissed):

DeltaFire15:
Normal Fighters being quite literally useless against 90% of ships seems kinda a bit overkill..

Also you don't think that losing a *ton* of integrity and getting a maxcap explosion in your face is enough of a consequence for being hit by a torp?

Karmic-Skink:
> Also you don't think that losing a _ton_ of integrity and getting a maxcap explosion in your face is enough of a consequence for being hit by a torp?

It isn't on the torp code, look closer. And max caps are boring.

Demoxide:
> Normal Fighters being quite literally useless against 90% of ships seems kinda a bit overkill..
Absolutely agree here, if you're gonna nerf fighters and make then even more boring and unfun don't bother "rebalancing" the light values to make them impossible to be killed by lights, heavy fighters are already constantly mauled by PDC and other weapons while strafing due to their slowness and stubborness to manuever that the only reliable way to deal with bigger ships is a light fighter, not only that but if you're nerfing light fighters, you're also locking the fun to 90% of fighter pilots due to the light to heavy fighter ratio on most maps, to add ontop of that, theres barely any "soft" targets such as enemy fighters or bombers that make it worth while actually launching light fighters to intercept, as putting more than 5 fighters or bombers is likely to lag the game due to ship count (?). All in all I think either revert the light values,  make the light cannon a medium one or increase the missile cap on light fighters and turn them into anti ship missiles, dealing purely medium class damage (make sure to QOL this so munitions doesn't have to kill their VLS to supply fighter pilots and to give manual lock because auto lock DOES NOT work and we'd like to actually be able to single out ships) Sorry for the rant but this needs to be addressed as I feel like fighter pilots are never being addressed.

Demoxide:
> In the current trend of munitions changes I think fighters are often forgotten. They might need rebalanced with the new roadmap, or just dropped from the roster if they no longer fit anywhere

They really should be revamped, dropping them from the roster ruins an unique experience that no other server really has, which is being able to make a big difference in a round by helping your ship while dodging and other stuff, you get the idea.

Karmic-Skink:
Fighters are outside of the scope of this PR.

Demoxide:
> Fighters are outside of the scope of this PR.

Sure, they're not the target of this PR, but they're influenced by this PR heavily, so you should have them in consideration and im still waiting for a proper response other than a pratical blank response.

Neglecting to adjust affected areas that this PR reaches into is just bad practice. Changing ship combat as a whole by altering armor values and adding in a whole new armor flag means every single overmap entity is affected, which you did address and balance, except fighters, where you simply skimmed over without actually thinking how this update would affect fighters as a whole.

Listen, we already have it rough as is, non functional missile autolocks, no countermeasures, pratical permadeath unless a competent medic saves you (R.I.P escape pods) and most enemy ships want you dead as soon as they see you. We don't need piss poor damage and being target practice for enemy ships, its simply NOT fun and makes us fighter pilots feel discriminated/neglected from the rest of munitions and overall game.

Unless of course, you address this in a future (but VERY soon) PR.

covertcorvid:
> Unless of course, you address this in a future (but VERY soon) PR.
There will be another PR addressing fighters.

Demoxide:
> There will be another PR addressing fighters.
>
Thanks for the info!

Bokkiewokkie (dismissed):

covertcorvid (approved):

PR #1524, opened 2021-07-07
Makes superstructure crit explosion intensity vary depending on damage dealt to the ship.

Bokkiewokkie:
## About The Pull Request

This PR changes the amount of damage dealt by SScrit explosions, in the current revision they do less damage in total, but will be more spread out and the intensity will vary depending on how much damage is done to the ship at once. (Total explosion strentgh here is still subject to change based on information we gather from the testmerge.)
Explosions generated by jumping the ship while in SScrit are also changed, the longer the ship has been critically damaged for, the more damage you will take when jumping. This was done to encourage retreating out of a system early to avoid being swamped by a powerful enemy fleet if your ship isn't equipped to handle it.

## Why It's Good For The Game

One pdc bullet causing a size 3,4,4 explosion is a bit intense and often times I see ships that can recover fairly easily get shredded because of this. I feel like this will make superstructure crit feel more fair and give people a better chance to recover from it.

## Changelog
:cl:
balance: Changes how superstructure crit explosions work.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/ftl.dm
+++ b/nsv13/code/modules/overmap/ftl.dm
@@ -198,7 +198,7 @@
 					var/name = pick(GLOB.teleportlocs)
 					var/area/target = GLOB.teleportlocs[name]
 					var/turf/T = pick(get_area_turfs(target))
-					new /obj/effect/temp_visual/explosion_telegraph(T)
+					new /obj/effect/temp_visual/explosion_telegraph(T, damage_amount = ((world.time - structure_crit_init)*10/3))
 		SSstar_system.ships[src]["target_system"] = target_system
 		SSstar_system.ships[src]["from_time"] = world.time
 		SSstar_system.ships[src]["current_system"] = null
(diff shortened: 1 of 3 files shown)


nsv13/code/modules/overmap/ftl.dm
```diff
@@ -198,7 +198,7 @@
 					var/name = pick(GLOB.teleportlocs)
 					var/area/target = GLOB.teleportlocs[name]
 					var/turf/T = pick(get_area_turfs(target))
-					new /obj/effect/temp_visual/explosion_telegraph(T)
+					new /obj/effect/temp_visual/explosion_telegraph(T, damage_amount = ((world.time - structure_crit_init)*10/3))
```
DeltaFire15, line 201: You may want to revise the scaling here. Right now it reaches maximum intensity after 6 seconds.

Bokkiewokkie: Yup because I was calculating with minutes in my head, very cool 👍

covertcorvid (approved):

PR #1525, opened 2021-07-08
Some fun comsig stuff

DeltaFire15:
## About The Pull Request
Adds a new comsig for coders / admins, plus some new *fun* to encounter infrequently.
Edit: Also fixes a small oversight in the defend goal that never actually mattered before since they were always next to a supply ship and fixes a small oversight in Carriers.

## Why It's Good For The Game
More *fun* to deal with is nice, plus this comsig should be useful elsewhere too.

## Changelog
:cl:
add: A new syndicate ship design, to be encountered sometimes.
add: Interdiction component.
fix: The Defend goal should work a tiny bit better.
fix: Carriers should no longer sometimes have their brain give out if damaged.
/:cl:


Karmic-Skink:
Heck yes we want interdictors.

DeltaFire15:
OnZImpact is apparently failing, I didn't touch anything the like so ignore CI fail.

nsv13/code/modules/overmap/types/syndicate.dm
```diff
@@ -375,6 +375,45 @@
...
+	obj_integrity = 2250	//Pretty thick hull due to it being a priority target
+	max_integrity = 2250
+	integrity_failure = 2250
+	icon = 'nsv13/icons/overmap/syndicate/syn_light_cruiser.dmi'
+	icon_state = "advanced_cruiser"
+	damage_states = FALSE	//Maybe later
+	sprite_size = 96
+	bound_height = 128
+	bound_width = 128
+	armor = list("overmap_light" = 90, "overmap_heavy" = 25)
```
Bokkiewokkie, line 390: You're now going to want to add overmap_medium to this aswell, and I think 2250 integrity is way too much, maybe make it 1500-1750 max?

code/__DEFINES/components.dm
```diff
@@ -404,3 +404,8 @@
 
 //Heretics stuff
 #define COMSIG_HUMAN_VOID_MASK_ACT "void_mask_act"
+
+//NSV13 comsigs begin
+#define COMSIG_GLOB_CHECK_INTERDICT "check_interdict"							//from interdiction component
+#define BEING_INTERDICTED (1<<0)											//returned on successful interdict
```
Bokkiewokkie, line 410: Why not just make this 1?

Bokkiewokkie (changes requested):

code/__DEFINES/components.dm
DeltaFire15: Future-proofing to allow easyish adding of different interdict behavior via additional bitfields which would not be possible via only a normal number returned.

nsv13/code/modules/overmap/types/syndicate.dm
DeltaFire15: Aight.

code/__DEFINES/components.dm
Karmic-Skink: Agree with the use of bitflags here.

PR #1537, opened 2021-07-13
Plutonium Sludge Placement is more reasonable

### LINKED ISSUE #1526: Radiation sludge can spawn under windows
## Description:
When a random sludge event, a reactor blowout, or digging sludge out of a reactor pit spawns plutonium sludge, it's possible for some of it to spawn under windows or other tile blocking structures, causing it to be unreachable unless someone removes the thing on top of it.

## Round ID:

Most rounds, but 2726 had it also.

## Reproduction:

1. Sludge spawns under windows
2. Wait for it to sludge
3. Radiation poisoning

TheOneAndMightyRed:
## About The Pull Request

Plutonium Sludge spawns in less unreasonable positions.
No longer spawns inside dense and anchored locations (ex. windows, machines, closed airlocks) while allowing canisters and cell charger turfs to still be affected for example.
Can no longer stack with other sludge.
Can no longer spawn on ladders and stairs.

## Why It's Good For The Game

Fixes #1526 and did a bit of extra checks.

## Changelog
:cl:
balance: Plutonium sludge spawns in fewer unreasonable locations.
balance: Plutonium sludge no longer stacks.
/:cl:


nsv13/code/modules/power/stormdrive.dm
```diff
@@ -1520,19 +1527,50 @@ Control Rods
...
+	place_sludge(get_turf(src), TRUE)
 	for(var/turf/open/floor in orange(range, get_turf(src)))
-		if(prob(35)) //Scatter the sludge, don't smear it everywhere
-			new /obj/effect/decal/nuclear_waste (floor)
+		place_sludge(floor, FALSE)
+
 	qdel(src)
 
+/// Tries to place plutonium sludge on 'floor'. Returns TRUE if the turf has been successfully processed, FALSE otherwise.
+/obj/effect/landmark/nuclear_waste_spawner/proc/place_sludge(turf/open/floor, epicenter = FALSE)
```
Karmic-Skink, line 1550: Why do you have an extra proc here for inserting just into fire() anyway?

nsv13/code/modules/power/stormdrive.dm
```diff
@@ -88,6 +88,13 @@ Control Rods
 
 #define MAX_CONTROL_RODS 5
 
+//Plutonium sludge
+
+#define PLUTONIUM_SLUDGE_RANGE 5
```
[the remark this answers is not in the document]
TheOneAndMightyRed: Habit from working on another codebase since the range on the effect spawner were 'magic random numbers' and should be #define'd. Could remove the #define if demanded.

nsv13/code/modules/power/stormdrive.dm
```diff
@@ -88,6 +88,13 @@ Control Rods
...
+#define PLUTONIUM_SLUDGE_RANGE_STRONG 10
```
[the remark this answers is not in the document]
TheOneAndMightyRed: Woops, yeah forgot about that.

nsv13/code/modules/power/stormdrive.dm
TheOneAndMightyRed: I thought the code would be cleaner if the sludge placement code was separated to another proc instead of the main fire() proc. Also the 'orange' returns all turfs except the center so i elected to call my new proc twice, one for the center dense sludge and one normal sludge for every turf except the center.

TheOneAndMightyRed:
Moved the plutonium sludge and it's spawners to a new file since review.

nsv13/code/game/objects/effects/custom_landmarks.dm
```diff
@@ -0,0 +1,66 @@
...
+#define PLUTONIUM_SLUDGE_RANGE_WEAK 3
+
+#define PLUTONIUM_SLUDGE_CHANCE 35
+
+
+/obj/effect/landmark/nuclear_waste_spawner //Clean way of spawning nuclear gunk after a reactor core meltdown.
+	name = "nuclear waste spawner"
+	var/range = PLUTONIUM_SLUDGE_RANGE //tile radius to spawn goop
+	var/center_sludge = TRUE // Whether or not the center turf should spawn sludge or not.
+	var/list/avoid_objs = list( // List of objs that the waste does not spawn on
```
TheFakeElon, line 14:
```suggestion
	var/static/list/avoid_objs = list( // List of objs that the waste does not spawn on
```

nsv13/code/game/objects/effects/custom_landmarks.dm
```diff
@@ -0,0 +1,66 @@
...
+	if(!prob(PLUTONIUM_SLUDGE_CHANCE)) //Scatter the sludge, don't smear it everywhere
+		return TRUE
+
+
+	for(var/obj/O in floor)
+		if(O.density && O.anchored) // Dense and anchored objects block sludge (ex. machines, windows) while others (ex. canisters, cell chargers) do not.
+			return TRUE
+		for(var/blacklisted in avoid_objs)
+			if(ispath(O.type, blacklisted))
+				return TRUE
```
TheFakeElon, line 51: It'd be much cleaner and more performant to use a typecache instead

Karmic-Skink:
Yeah looking much better.

nsv13/code/game/objects/effects/custom_landmarks.dm
TheOneAndMightyRed:
```suggestion
	var/blacklisted_cache = typecacheof(avoid_objs)
	for(var/obj/O in floor)
		if(O.density && O.anchored) // Dense and anchored objects block sludge (ex. machines, windows) while others (ex. canisters, cell chargers) do not.
			return TRUE
		if(is_type_in_typecache(O, blacklisted_cache))
			return TRUE
```

I never did something with typecaches so i'm asking if this is what you had in mind? avoid_obj is static now by the way.
I must have done something wrong because i don't feel like this solution would be more performant than the way i did it tho.

TheFakeElon: `munitions_trolley.dm` contains a good example of typecaches and how to check them

TheOneAndMightyRed: Aaah, perfect, thanks.

nsv13/code/game/objects/effects/custom_landmarks.dm
```diff
@@ -0,0 +1,64 @@
...
+		for(var/obj/effect/decal/nuclear_waste/waste in floor) //Replace nuclear waste with the stronger version
+			qdel(waste)
+		new /obj/effect/decal/nuclear_waste/epicenter (floor)
+		return TRUE
+
+	if(!prob(PLUTONIUM_SLUDGE_CHANCE)) //Scatter the sludge, don't smear it everywhere
+		return TRUE
+
+	for(var/obj/O in floor)
+		if(O.density && O.anchored) // Dense and anchored objects block sludge (ex. machines, windows) while others (ex. canisters, cell chargers) do not.
```
Bokkiewokkie, line 46: I think it would be better to just include grilles, girders, windows, the gateway and gravity generator in the avoid_objs list, because right now this stops sludge from spawning on way too many things, including but not limited to:
1. Directional windows
2. Doors & closed directional firedoors
3. Any machine or computer
4. Tables
5. Wrenched down crates and (closed) lockers
With this behaviour I would not be surprised if someone gets the brilliant idea to cover the entire ship in tables to avoid sludge spawning at all.

Bokkiewokkie (approved):

Karmic-Skink (approved):

PR #1539, opened 2021-07-16
Boarding Cleanup

covertcorvid:
## About The Pull Request
Cleans up some boarding stuff that we're not using right now.
Bonus Squad HUD fix

## Changelog
:cl:
tweak: Gave all pilots both types of pilot access
tweak: Moved squad HUD to right side to uncover job and unlinked it from the sechud
del: Disabled the salvage console
fix: Fixed job selection screen having an awkward 4th column
/:cl:


Vasily2013:
Doesn't actually change any .dmms, so mapping is not applicable.

Karmic-Skink (approved):

Karmic-Skink:
We'll do one more check and then merge it.

PR #1545, opened 2021-07-23
Stops Github Actions from closing stale issues after 180 days

Bokkiewokkie:
## About The Pull Request

Does what it says

## Why It's Good For The <s>Game</s> Codebase

Old issues being closed is more annoying then it is helpful, seeing as we <s>are drowning in debt</s> have a somewhat slower development cycle, and looking at open issues every so often isn't that hard of a job for a maintainer to do.


Karmic-Skink (approved):
-1

PR #1576, opened 2021-08-11
Eclipse layout changes

TheOneAndMightyRed:
## About The Pull Request

Eclipse has become quite stale in the last couple weeks, so i thought doing some changes to the layout will give it some spice.
[R&D](https://i.imgur.com/YX2Bdqx.png) and [Xenobiology](https://i.imgur.com/11wjICQ.png) swapped places.
[Cargo](https://i.imgur.com/UYMj8Yv.png) relocated to where the lower dorms used to be, complete with much missing stamps and the cargo budget. 1 Torp launcher for mission crates included as well.
Mining is now only on the lower external airlock and there is only one ORM now. [Upper airlock](https://i.imgur.com/ETiosZg.png) now consists of regular EVA suits.
Old Cargo space converted into a sort of [construction area](https://i.imgur.com/k3izy3u.png) people can use for building a tesla, singulo, hangar, etc.
Mini atmos got a nitrogen gas miner to complement the oxygen gas miner so eclipse doesn't run out of air after 1 or 2 breaches.
Also fixed the XO landmark spawn icon and name.

## Why It's Good For The Game

Variety good.
Stale bad.

## Changelog
:cl:
add: Added eclipse nitrogen gas miner
del: Removed eclipse lower dormatories
balance: Changed Eclipse layout. (R&D and Xeno swapped, cargo relocated to lower dorms, old cargo converted to const. area)
/:cl:

### DIFF
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -130,8 +130,8 @@ INITIALIZE_IMMEDIATE(/obj/effect/landmark)
 	icon_state = "Chief Engineer"
 
 /obj/effect/landmark/start/head_of_personnel
-	name = "Head of Personnel"
-	icon_state = "Head of Personnel"
+	name = "Executive Officer"
+	icon_state = "Executive Officer"
 
 /obj/effect/landmark/start/librarian
 	name = "Curator"
--- a/code/game/objects/effects/landmarks.dm
+++ b/code/game/objects/effects/landmarks.dm
@@ -130,8 +130,8 @@ INITIALIZE_IMMEDIATE(/obj/effect/landmark)
 	icon_state = "Chief Engineer"
 
 /obj/effect/landmark/start/head_of_personnel
-	name = "Executive Officer"
-	icon_state = "Executive Officer"
+	name = "Executive Officer" //NSV13 Edit
+	icon_state = "Executive Officer" //NSV13 Edit
 
 /obj/effect/landmark/start/librarian
 	name = "Curator"
--- a/_maps/eclipse.json
+++ b/_maps/eclipse.json
@@ -6,7 +6,7 @@
 	"traits": [{"Up": 1, "Linkage": "Self"}, {"Down": -1, "Linkage": "Self"}],
     "space_ruin_levels": -1,
     "shuttles": {
-        "cargo": "cargo_hammerhead",
+        "cargo": "cargo_atlas",
         "ferry": "ferry_fancy",
         "emergency": "emergency_donut"},
     "mine_disable": 1,
--- a/_maps/map_files/Eclipse/Eclipse1.dmm
+++ b/_maps/map_files/Eclipse/Eclipse1.dmm
@@ -30,20 +30,19 @@
 	},
 /area/hallway/primary/fore)
 "ah" = (
-/obj/machinery/atmospherics/pipe/simple/supply/hidden/layer1{
-	dir = 4
+/obj/structure/window/reinforced/spawner,
+/obj/effect/turf_decal/box/white,
+/obj/machinery/door/window/westright{
+	req_one_access_txt = "12"
 	},
-/obj/machinery/atmospherics/pipe/simple/scrubbers/hidden/layer3{
-	dir = 4
-	},
-/obj/structure/cable{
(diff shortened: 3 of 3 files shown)


code/game/objects/effects/landmarks.dm
```diff
@@ -130,8 +130,8 @@ INITIALIZE_IMMEDIATE(/obj/effect/landmark)
 	icon_state = "Chief Engineer"
 
 /obj/effect/landmark/start/head_of_personnel
-	name = "Head of Personnel"
-	icon_state = "Head of Personnel"
+	name = "Executive Officer"
```
Karmic-Skink, line 133: As this is a core file, it needs to contain an //NSV13 tag

Karmic-Skink (changes requested):

Ansome101:
Original map maker and maintainer of the Eclipse here,

although retired, I heard somebody was touching my map! I actually made the "Eclipse MK2" update for this map, which covers many of the issues you had but does not resort to giving the players freebies like a gas miner, but unfortunately did not find the time to PR it. I've handed over the files to an individual I trust can maintain the Eclipse while still maintaining its vision and purpose.

I absolutely condemn the potential change you've proposed as you've clearly taken no consideration for the inherent purpose, style, or "lore" of the map in any regard. The Eclipse was the solution to a sharp playerbase decline now known as "deadpop" and focuses on cutting down on content so the players may play a simplified, but playable version of NSV13. As such, many departments you'd see on a regular ship are either non-present or drastically compressed. I see no reason that a FULL T-comms setup is needed given that the majority of jobs which would use this system are defunct on this map. Continuing on this point, the layout, like many other things on the Eclipse, is intentional. In order to ensure that all parts of the ship are explored and players cannot group up in the upper half of the ship, scarcely going to the back, I purposefully distributed the key departments to several corners of the ship. Combined with the "RP zones" system in place in which there are several chokepoints in which players are likely to encounter one another, it ensures that even the scientist junkie doing only research will inevitably encounter someone else on the vessel on their way to get food or do mining. Your proposed changes will, given that Xenobio is an auxiliary department should population permit it, disrupt this flow and leave players congregating exclusively on the right side of the map.

It is also readily apparent to me that you don't understand what the Eclipse's intention was given the aesthetic of it. This is a civilian-grade cargo and mining vessel that has been lightly refurbished with better armaments to better serve in the ongoing war. While realistically this may entail a more thorough refurbishment of the interior, I opted instead to maintain its aesthetics and somewhat grimy and "lived-in" appearance as a means of differentiating the map stylistically from standard NT vessels.. Despite this, you've pointlessly decided that half of the EVA suits are now standard ones instead of the more practical mining suits they would be using on their day-to-day job and replaced the simplistic one-channel Tcomms for no reason I can discern other than NT mapping standards.

 I politely ask that you, or whomever it may concern in the future, look more closely at the Eclipse and consider whether the design choice made was intentional or not and what its inherent purpose might be. Any map may grow "stale" in nature, but much like when you grow tired of your current living arrangement, you move to a new place or rearrange the furniture and wallpaper. You do not break down several walls in your home, throw out half of your clothes and place uncharacteristic clothing in your drawers, and install a full military-grade radio tower in what used to be your private hobby room for the one radio channel you actually listen to. 
 
 I will have my chosen map maintainer PR the official, updated version of the Eclipse which addresses many of your concerns without compromising any of the aforementioned details in the coming days. From there, I would be more content in having you touch the map given you now take a closer look at why the Eclipse is the way it is. I may be retired and moving on from NSV13, but I do ask that you only make intelligent changes to the work I've done out of courtesy.
 
 Yours truly,
 -Ansome
 
 P.S. - You can go to the atmos room and lower the pressure output to avoid leaking too much air during a breach, though I've heard that not getting hit is also a good way of not losing air.

code/game/objects/effects/landmarks.dm
Vasily2013: Pretty sure that this is not an actual issue that can be noticed in game. It's probable that all our maps do this.

Karmic-Skink: No, that is a rebasing issue.

athena413:
Screenshot_20210811-162120_Discord

PR #1583, opened 2021-08-14
MAPPING FREEZE UNTIL COVERT'S PDC CHANGES ARE IMPLEMENTED.

Vasily2013:
This is an empty PR to signify a mapping freeze pending changes to the PDC systems.


Vasily2013:
Unless there is an emergency fix that is immediately required this will be upheld. If there is an emergency fix to a map those changes should be sent to @covertcorvid with a .dmm file of the affected area to prevent merge conflicts.

covertcorvid:
Also I won't be touching the Whisp, so Aerwhisp can't use this to procrastinate laser AMS :P

PR #1618, opened 2021-09-07
"Fixes" some homing stuff, plus more nukings

DeltaFire15:
## About The Pull Request
This *might* fix some weird behavior for homing munitions (mainly player torps). The main thing it does is it ignores if you click a tile so if you do it just flies straightforward instead of circling that tile. If you click a ship (or thing), it'll still use prior behavior
Also changes the code for mac / nac homing slightly to be safer, and fixes the thing that made artillery sometimes become ultrapiercing even if it wasn't AP.

Also makes torp ships able to throw torp at *certain* player ships because some people were sad nuke ships weren't using nukes at them.

Should be testmerged first. This *might* cause unforseen consequences with how conveluded weapon code is.

## Why It's Good For The Game
Fix man good, *if it works*

## Changelog
:cl:
fix: Torpedos and other homing things no longer home in on empty tiles.
fix: NAC rounds can no longer go into pseudo-godmode
balance: AI ships armed with torpedos now target smaller ships (bigger than Fighters)
/:cl:

### DIFF
--- a/code/modules/projectiles/projectile.dm
+++ b/code/modules/projectiles/projectile.dm
@@ -553,7 +553,7 @@
 	Range()
 
 /obj/item/projectile/proc/process_homing() //Nsv13 - Enhanced the performance of this entire proc.
-	if(!homing_target || QDELETED(homing_target)) //NSV13 - Changed proc to be less performance intensive
+	if(QDELETED(homing_target)) //NSV13 - Changed proc to be less performance intensive
 		return FALSE
 	var/targetAngle = Get_Angle(src, homing_target)
 	var/angle = closer_angle_difference(Angle, targetAngle)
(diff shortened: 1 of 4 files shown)


code/modules/projectiles/projectile.dm
```diff
@@ -553,7 +553,7 @@
 	Range()
 
 /obj/item/projectile/proc/process_homing() //Nsv13 - Enhanced the performance of this entire proc.
-	if(!homing_target) //NSV13 - Changed proc to be less performance intensive
+	if(!homing_target || QDELETED(homing_target)) //NSV13 - Changed proc to be less performance intensive
```
TheFakeElon, line 556:
```suggestion
	if(QDELETED(homing_target)) //NSV13 - Changed proc to be less performance intensive
```

covertcorvid:
Fix man good

Bokkiewokkie (approved):

PR #1638, opened 2021-09-14
regex improvement from last PR

TheFakeElon:
## About The Pull Request

old regex works fine but was slow and messy


code/controllers/subsystem/fail2topic.dm
```diff
@@ -88,10 +88,10 @@ SUBSYSTEM_DEF(fail2topic)
 /datum/controller/subsystem/fail2topic/proc/BanFromFirewall(ip)
 	if (!enabled)
 		return
-	var/static/regex/R = regex(@"(\.0\.)|(\.0$)|(\l+|/|;|&|\||-|%)") // Anything that interacts with a shell should be parsed. Prevents direct call input tampering
-	if(length(ip) > 15 || length(findtext(ip, R)))
+	var/static/regex/R = regex(@"^\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}$") // Anything that interacts with a shell should be parsed. Prevents direct call input tampering
+	ip = findtext(ip, R, "")
```
covertcorvid, line 92: looking at the ref, isn't the third argument of findtext supposed to be a starting index? Also I think 0.0.0.0 and 127.0.0.1 are valid matches for this expression

TheFakeElon: 0.0.0.0 won't block all IPs in netsh without a subnet mask, 0.0.0.0/0 will though and regex would catch that

image

TheFakeElon: default CIDR is /32

covertcorvid (approved):

PR #1641, opened 2021-09-16
Shuttlecraft Chair

BetaCentari:
Adds a more space efficient chair for shuttles and ships. Can be fabricated with 10 sheets of iron and installed onto a wall. The seat folds up and down down to save space. 

Adds more variability to the seating options players and mappers have available to them.

## Changelog
:cl:
add: Added a wall-mounted, folddown chair for shuttlecraft and emergency strapping.
imageadd: Added sprites for said chair.
/:cl:


covertcorvid:
Is there any way to get it back off the wall?

BetaCentari:
> Is there any way to get it back off the wall?

You can wrench it off back into the iron sheets but otherwise, I don't think you can take it off the wall no.

Vasily2013:
A couple things 
- [ ] You can close the chair with a person buckled in it.
- [ ] Change the pixel shift to about 10, it'll look better and resemble that a person is actually in the tile (I haven't tested it but I assume person in chair is still dense)

covertcorvid:
Probably good to go once Vasiliy's comments are addressed

BetaCentari:
> A couple things
> 
> * [ ]  You can close the chair with a person buckled in it.
> * [ ]  Change the pixel shift to about 10, it'll look better and resemble that a person is actually in the tile (I haven't tested it but I assume person in chair is still dense)

Addressed both changes, pixels changed and can no longer close the chair when a mob is bucked into it.

Ikalpo:
Two things from me: change the `buildstackamount` of the chair and the materials in the wallframe to match the 10 iron cost.
And should it even cost 10 iron anyway? I think 5 would be fine.

BetaCentari:
> Two things from me: change the `buildstackamount` of the chair and the materials in the wallframe to match the 10 iron cost.
> And should it even cost 10 iron anyway? I think 5 would be fine.

Suggested changes made

covertcorvid (approved):

nsv13/code/game/objects/structures/shuttlecraft_chair.dm
```diff
@@ -0,0 +1,94 @@
...
+	return ..()
+
+/obj/structure/chair/shuttlecraft_chair/post_buckle_mob(mob/living/M)
+	. = ..()
+	update_armrest()
+
+/obj/structure/chair/shuttlecraft_chair/proc/update_armrest()
+	if(has_buckled_mobs())
+		add_overlay(armrest)
+		AltClick(null)
```
covertcorvid, line 51: this doesn't need to be here, it'll always return on lines 67 and 68 because null isn't a mob/living. 
just adding the has_buckled_mobs() to /obj/structure/chair/shuttlecraft_chair/AltClick(mob/living/user) was enough.

Ikalpo:
> > Two things from me: change the `buildstackamount` of the chair and the materials in the wallframe to match the 10 iron cost.
> > And should it even cost 10 iron anyway? I think 5 would be fine.
> 
> Suggested changes made

still needs a
`	buildstackamount = 5`
otherwise you'll only get one sheet of iron when deconstructing it

covertcorvid (changes requested):
Remove AltClick(null) from update_armrest() and set buildstackamount = 5

BetaCentari:
> Remove AltClick(null) from update_armrest() and set buildstackamount = 5

Done and done.

covertcorvid (approved):

Karmic-Skink (approved):

PR #1646, opened 2021-09-23
Adds temperature propagation

covertcorvid:
## About The Pull Request
So it turns out if you have two turfs next to each other with the same number of moles on them at vastly different temperatures and pressures, those aren't active and do nothing.
Pressure will be a lot harder to fix, but this at least makes temperature work.
See https://github.com/covertcorvid/extools/pull/1/files

## Why It's Good For The Game
The air functions slightly more like air

## Changelog
:cl:
fix: Made turf thermodynamics work better
/:cl:


covertcorvid:
Other fun monstermos facts:
A turf with 104 moles of oxygen will not mix with a turf with 22 moles of oxygen and 82 moles of nitrogen

TheFakeElon:
processing per gas type would probably range anywhere from 2-5x more expensive than what we have now, you could probably pseudo code it with more math estimates but the cost probably isn't worth it. It's dumb but it makes sense.

Speaking of which, does this effect performance much?

Bokkiewokkie:
Going to put this in the testmerge to see how it affects performance.

PR #1660, opened 2021-09-27
Changes light fighter gameplay slightly.

RemervKeberb:
## About The Pull Request
Reduces light cannon ammo capacity. Increases ROF slightly and increases missile damage. 

## Why It's Good For The Game

Diversity is the spice of life. The heavy and light fighter play more or less the same right now. With minor tactical differences. This PR would shake up light fighter gameplay to more short and snappy conflict with a larger emphasis on missiles and RTBing.
This does increase missile damage overall but as is. Missiles are never used offensively by the player ship and the player ship has more than enough defenses against enemy missiles. Even if they get through the damage isn't too rough.

## Changelog
:cl:
tweak: Light fighter cannon holds only 150 rounds.
balance: Light fighter cannon has a higher ROF
balance: Missiles do more damage
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/weapons/projectiles_fx.dm
+++ b/nsv13/code/modules/overmap/weapons/projectiles_fx.dm
@@ -168,7 +168,7 @@ Misc projectile types, effects, think of this as the special FX file.
 	name = "triton cruise missile"
 	icon_state = "conventional_missile"
 	speed = 1
-	damage = 175
+	damage = 250
 	valid_angle = 120
 	homing_turn_speed = 25
 	range = 250
@@ -239,7 +239,7 @@ Misc projectile types, effects, think of this as the special FX file.
 	tracer_type = /obj/effect/projectile/tracer/disabler
 	muzzle_type = /obj/effect/projectile/muzzle/disabler
 	impact_type = /obj/effect/projectile/impact/disabler
-	
+
 /obj/item/projectile/beam/laser/point_defense
 	name = "laser pointer"
 	damage = 30
--- a/nsv13/code/modules/munitions/ammunition/cannon_ammo.dm
+++ b/nsv13/code/modules/munitions/ammunition/cannon_ammo.dm
@@ -5,8 +5,8 @@
 	icon_state = "light"
 	ammo_type = /obj/item/ammo_casing/light_cannon
 	caliber = "mm20"
-	max_ammo = 500
-	
+	max_ammo = 150
+
 /obj/item/ammo_box/magazine/nsv/heavy_cannon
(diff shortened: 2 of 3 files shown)


nsv13/code/modules/overmap/weapons/projectiles_fx.dm
```diff
@@ -168,7 +168,7 @@ Misc projectile types, effects, think of this as the special FX file.
 	name = "triton cruise missile"
 	icon_state = "conventional_missile"
 	speed = 1
-	damage = 175
+	damage = 250
```
Bokkiewokkie, line 171: This makes them deal an equal amount of damage as regular torpedoes, a bit high in my opinion. Maybe consider lowering the damage of PDC rounds or increasing the missile's health slightly instead of doing this?

Bokkiewokkie (changes requested):

nsv13/code/modules/overmap/weapons/projectiles_fx.dm
RemervKeberb: Problem with raising the health is that they'll be even more annoying for the main ship to shoot down. Pretty much requiring the missiles to be going perfectly straight and for PDC's to already be firing in it's direction before they enter visual range.
A speed increase would be the same issue.
Now that you mention it a damage increase isn't the ideal solution either.
More missiles could be a good balance? Having 6 missiles would deal less than the proposed change but still close enough. Only problem I can see there however is getting the missiles in the first place from muni. Since the whole idea of this would be circling around, getting ideal launch parameters and then RTBing once you're out. However for RTBing to be at all useful you need a steady supply of missiles which isn't very easy to acquire. Would have to see how the players adapt, if they don't the change wouldn't be very feasible.

RemervKeberb:
Also if the higher ROF lags the server the ammo count and ROF could be lowered with the damage raised.

nsv13/code/modules/overmap/weapons/projectiles_fx.dm
RemervKeberb: If you like the idea of raising the default missile count I'd also have to change the upgraded launcher's count as it'd be a downgrade at that point.

RemervKeberb:
Tested it out and 6 missiles just doesn't feel as satisfying for whatever reason. I also simulated time dilation by just spawning in a metric fuckton of fleets and then in my own corner against a single ship I realized that the biggest flaw in this is lag. Getting ideal launches when the server is chugging is near impossible. Which sadly is just a constant that we'll never be able to get rid off.
All in all, too much effort.

PR #1672, opened 2021-10-01
Disruption torpedoes and some fixes

DeltaFire15:
## About The Pull Request
The main thing of this PR is that it adds a new torpedo type, disruption torpedoes, which if they hit an AI ship or Fighter cause major disruption in its systems. If they hit a ship with an internal z level, they'll cause an EMP pulse instead, plus a minor explosion. Their damage is lower than normal torpedoes, but their Utility can make them worthwhile nonetheless.
(Torp type AI only for now, though I guess it'd be nice to have a torpedo during combat that focusses more on Utility.. Whatever maintainers want it to be I guess)

Additionally, this amps up the Aspala class stealth Destroyer to be a more useful ship than before. With its stealth, it makes for a good chassis for a harasser / skirmisher, though its weapons did not reflect this. This gives it a special type of light autocannon (it counts as light weapon instead of heavy since thats a Fighter weapon on a big ship). And, the more scary part, it now has a burst disruption torpedo launcher, which fires a quick burst of three disruption torpedoes with a bigger-than-usual reload, making it a major threat if it manages to sneak up on a target.

Additionally, this PR fixes a pretty major bug with the starsystem proc for moving ships around between starsystems (move_existing_object()), which previously had some issues regarding adding things to system contents (which lead to issues with resupply and simillar)

AND, this aims at fixing friendly projectiles hitting eachother. Again. Yeah this is a very hard to track down bug. I might've gotten it with one of the changes I did there though, since they stopped destroying eachother during testing when they did before. I have no idea which of the changes I did to the projectiles-hitting-eachother logic fixed that, or if it got all situations. Only time will tell.

## Why It's Good For The Game
The Aspala as of now is far less of a threat than would be appropriate for its size and role (it's basically just a big Fighter threat wise). This makes it a.. uniqueish.. threat to contend with, due to its main weapon, whilst not too damaging, wreaking havoc on electronics. (And Karmic might enjoy throwing these disruption torpedoes on some pirate ships during the rework, I figure)

The fixes, well, obviously, fix man good.

## Changelog
:cl:
add: added disruption torpedoes (AI only for now)
balance: The syndicate has supplied its Stealth Destroyer class with a new disruptive weapon system. Be mindful of it hiding in the dark during your journeys.
fix: fixed Projectiles hitting eachother when friendly (Again. Maybe.)
fix: ships no longer are in no system's system contents after jumping.
/:cl:


nsv13/code/modules/overmap/weapons/projectiles_fx.dm
```diff
@@ -205,15 +211,41 @@ Misc projectile types, effects, think of this as the special FX file.
...
+		return FALSE
 	return BULLET_ACT_HIT
 
+/obj/item/projectile/guided_munition/proc/spec_overmap_hit(obj/structure/overmap/target)
+	return
+
+/obj/item/projectile/guided_munition/torpedo/disruptor/spec_overmap_hit(obj/structure/overmap/target)
+	if(QDELETED(target))
+		return
+	if(target.occupying_levels.len)
```
TheFakeElon, line 224:
```suggestion
	if(length(target.occupying_levels))
```

nsv13/code/modules/overmap/weapons/projectiles_fx.dm
```diff
@@ -205,15 +211,41 @@ Misc projectile types, effects, think of this as the special FX file.
...
+	else
+		return FALSE
 	return BULLET_ACT_HIT
 
+/obj/item/projectile/guided_munition/proc/spec_overmap_hit(obj/structure/overmap/target)
+	return
+
+/obj/item/projectile/guided_munition/torpedo/disruptor/spec_overmap_hit(obj/structure/overmap/target)
+	if(QDELETED(target))
+		return
```
TheFakeElon, line 223: probably won't need this, I don't think spec_overmap_hit can pass null datums

nsv13/code/modules/overmap/weapons/weapons.dm
```diff
@@ -72,20 +72,26 @@
...
-		fire_projectile(torpedo_type, target, homing = TRUE, speed=3, lateral = TRUE, ai_aim = ai_aim)
-		var/obj/structure/overmap/OM = target
-		if(istype(OM, /obj/structure/overmap) && OM.dradis)
-			OM.dradis?.relay_sound('nsv13/sound/effects/fighters/launchwarning.ogg')
-		var/datum/ship_weapon/SW = weapon_types[FIRE_MODE_TORPEDO]
-		relay_to_nearby(pick(SW.overmap_firing_sounds))
+		var/launches = min(torpedoes, burst)
+		fire_single_torpedo(target, ai_aim)
+		for(var/launched = 1; launched < launches; launched++)
+			addtimer(CALLBACK(src, .proc/fire_single_torpedo, target, ai_aim), (0.25 SECONDS * launched))
```
TheFakeElon, line 82:
```suggestion
			addtimer(CALLBACK(src, .proc/fire_single_torpedo, target, ai_aim), (2.5 * launched))
```

TheFakeElon: IMO this is timer spam. All torpedo firing should be handled in one proc or at least in a way that doesn't use timers. 

Making and deleting timers are pretty expensive, ideally they should only be used for long term delays (i.e 5+ seconds). Using multiple for the same purpose is generally a waste too

nsv13/code/modules/overmap/weapons/weapons.dm
```diff
@@ -72,20 +72,26 @@
...
+		torpedoes -= launches
 		return TRUE
 
-/obj/structure/overmap/proc/fire_missile(atom/target, ai_aim = FALSE)
+/obj/structure/overmap/proc/fire_single_torpedo(atom/target, ai_aim = FALSE)
+	fire_projectile(torpedo_type, target, homing = TRUE, speed=3, lateral = TRUE, ai_aim = ai_aim)
+	var/obj/structure/overmap/OM = target
+	if(istype(OM, /obj/structure/overmap) && OM.dradis)
```
TheFakeElon, line 89:
```suggestion
	if(isovermap(OM) && OM.dradis)
```

nsv13/code/modules/overmap/weapons/projectiles_fx.dm
DeltaFire15: Is a length() call that handles multiple variations of types really more efficient than a single variable access?

nsv13/code/modules/overmap/weapons/projectiles_fx.dm
DeltaFire15: *should* probably not happen yeah. If it does I guess I get to scream at whoever put a sleep somewhere wacky.

nsv13/code/modules/overmap/weapons/weapons.dm
DeltaFire15: I've been told before to favour timers over sleeps or spawns, though I get where you are coming from overheadwise.

nsv13/code/modules/overmap/weapons/projectiles_fx.dm
TheFakeElon: yes

nsv13/code/modules/overmap/weapons/weapons.dm
TheFakeElon: Relatively speaking, they are pretty expensive; they're the main reason why runescape chat has such a performance drag. The more timers that are used the slower everything else that uses them becomes because byond single thread moment

DeltaFire15:
ok, stuff should be resolved

Bokkiewokkie (approved):
As Kmc mentioned some things we could refactor later, but for now it's good as long as it just works.

PR #1681, opened 2021-10-06
Armada sieging now freezes Sol system combat

DeltaFire15:
## About The Pull Request
Title.
While it was interesting for players to arrive to both fleets already banged up, it's kinda anticlimatic for them to drop everything and rush there just for one nuke ship to remain or something like that.
Additionally it'll prevent all jank related to strength of any related fleets finishing combat pre siege timer end.
Yes, this means players *need* to be there now and can't just sometimes be lucky and have the EDF win it.

## Why It's Good For The Game
Should be way better and cause less jank than the current implementation.

## Changelog
:cl:
balance: The armada sieging Sol now freezes simulated combat in the system.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/fleet_combat/solar_siege.dm
+++ b/nsv13/code/modules/overmap/fleet_combat/solar_siege.dm
@@ -5,10 +5,6 @@ Note: This expects the EDF to be in Sol, and not somewhere in Brazil due to admi
 */
 
 /datum/star_system/sol/handle_combat()
-	. = ..()
-	if(. == COMBAT_SKIPPED)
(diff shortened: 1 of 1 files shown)


athena413:
if you *really* wanted to you could have it so that when the mainship jumps to sol it'll deal a bit of damage to each ship in the system but this is good regardless

fix man good

Kmc2000:
> I'll see about adding a button for that sometime though shouldn't be too difficult for anyone to do. Out of scope for this specific PR though.

Feel free to consult with me on that, happy to help on the UI side.

PR #1683, opened 2021-10-07
Front line reporting from White Rapids

covertcorvid:
## About The Pull Request
Sends out more information about the ongoing war, but in mostly smaller text.
Modified fleet names to fit the announcement better.

## Why It's Good For The Game
Gives a better view of what's going on in the war and adds some flavor

## Changelog
:cl:
tweak: Smaller notifications about where AI fleets are moving around
add: Normal-sized notifications when AI fleets enter combat with each other
tweak: specified proper/improper for the fleets and tweaked names where needed
/:cl:


TheFakeElon:
it's all so tiresome
image

Apparently the DM compiler's default function in string datum references is to check for a defined variable called `name` on it completely regardless of type 😭. I wonder if defining `var/name` on the top level datum will break anything 👀 

So I was wrong! There's nothing wrong with the macros at all

PR #1684, opened 2021-10-07
Fixes Dradis ghost radar bug

### LINKED ISSUE #1677: Ghosts can pulse the radar at a dradis console
## Reproduction:
1) Be ghost
2) Find DRADIS computer at the bridge
3) Start spamming radar pulse while crew is repairing cuz I ded restart
4)???
5) Profit

utilbird:
fixes #1677

## About The Pull Request

Fixes two oversights with Dradis consoles. One in ui_data where `send_radar_pulse()` was unconditionally called if it was in active radar mode. And the other in ui_interact which used an invalid mob argument. Both of these allowed any mob (including observers) to spam call radar regardless of MIN_RADAR_DELAY.

## Changelog
:cl:
fix: Fixes Dradis ghost interaction
fix: Radar can no longer be spammed faster than the radar delay
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/radar.dm
+++ b/nsv13/code/modules/overmap/radar.dm
@@ -90,12 +90,12 @@ Called by add_sensor_profile_penalty if remove_in is used.
 
 /obj/machinery/computer/ship/dradis/proc/send_radar_pulse()
 	var/obj/structure/overmap/OM = get_overmap()
-	if(!can_radar_pulse())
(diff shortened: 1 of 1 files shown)


covertcorvid:
So in theory this is test merged right now, and the watchdog actually has rebooted since I did that so I think it's really on there, and I can definitely ping DRADIS as a non-admin observer still provided I am close enough to the console.

covertcorvid:
this works now 👍

PR #1690, opened 2021-10-12
Adds "Dirty" Shells, makes MAC / NAC rounds once again able to recreate certain Expanse moments

DeltaFire15:
## About The Pull Request
First the normal NAC part.
NAC / MAC projectiles (For the AI type, the gun is still named ../mac, projectile is named Artillery Round, so who knows these days) now get relayed onto hit ship z-levels with a weaker but penetrating projectile. People who've been here a while might remember this.. Slightly more work for engineering and Medbay to do save for fixing nuked sections or nothing.
Values obviously can be adjusted if needs be, like if they keep onehitting weapons again like during old times. This relayed type does alot less damage than the actual overmap shell, don't worry.

Now. Dirty Shells.
These are a.. fun.. NAC / MAC AI projectile type which are designed with a ship crew's suffering in mind.
In exchange, they deal less normal damage so you have to worry less about getting hit by them... or should you?
They use a deep penetration fuze, so don't expect them to just harmlessly detonate on your ship's outside.
For now, Pirate gunships get these, until Karmic's redesign of pirates comes along. Should make them a tiiny bit more of a challenger to deal with (and they didn't have anything to use their shots_left aka heavy ammo on anyways)
Bunch of defines, values and other stuff to adjust in there, for if anything seems too strong / weak / common / whateverelse
Projectile itself could use a different sprite, I'll see about it sometime.

Also radiation sludge now ignores explosions, except devastation level ones, since the previous behavior was just inherited from the base ../effect type

## Why It's Good For The Game
MAC / NAC projectiles once again being able to pen through a ship seems like a interesting dynamic to bring back to have more overmap - shipz interactions.
Dirty shells should be another 'fun' addition to deal with in the same area, with the fact Pirates get them making them a *tiny* bit less of a not-issue

## Changelog
:cl:
add: NAC / MAC shells can once again penetrate through ship z-levels.
add: Adds Dirty shells, a devious ammunition used by certain factions.
balance: Pirate Gunships are now slightly less of a weakling.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/weapons/projectiles_fx.dm
+++ b/nsv13/code/modules/overmap/weapons/projectiles_fx.dm
@@ -119,7 +119,7 @@ Misc projectile types, effects, think of this as the special FX file.
 /obj/item/projectile/bullet/dirty_shell_stage_two/proc/release_payload(turf/detonation_turf)
 	var/list/inrange_turfs = RANGE_TURFS(DIRTY_SHELL_SLUDGE_RANGE, detonation_turf) - detonation_turf
 	new /obj/effect/decal/nuclear_waste/epicenter(detonation_turf)
-	for(var/turf/T in inrange_turfs)
+	for(var/turf/T as() in inrange_turfs)
 		if(isgroundlessturf(T))	//Should those kinds of turfs be able to get waste from this? Hmm, I dunno.
 			continue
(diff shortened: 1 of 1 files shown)


Karmic-Skink:
I haven't read the code yet, but this is definitely a step in the right direction. We are looking translating more damage from OM back to ship Z, so this fits the ticket.

Bonus warcrimes.

PR #1696, opened 2021-10-16
Makes shake_with_inertia() NOT kill performance

DeltaFire15:
## About The Pull Request
Shake with inertia, for every call, went through EVERY SINGLE thing in the machinery list. For every mob with a ckey in the worst case aka ship hits. That list has over 17000 things in it, roundstart!
This PR gives inertia dampeners their own list that can be gone through for it, and should NOT kill the server. Also has a check_tick and doesn't force procs to wait, as neither should be neccessary.
Edit: Jumps now also use that list, since those had the same issue (though obviously less pronounced since you don't get 10 jumps in a second usually)

## Why It's Good For The Game

fix man good.

## Changelog
:cl:
code: shake_with_inertia and FTL shake no longer kill performance.
/:cl:

### DIFF
--- a/nsv13/code/game/machinery/inertial_dampener.dm
+++ b/nsv13/code/game/machinery/inertial_dampener.dm
@@ -3,6 +3,8 @@
 #define EfficiencyToRange(part_efficiency) (2.5 * (part_efficiency))
 #define RangeToEfficiency(range) ((range) / 2.5)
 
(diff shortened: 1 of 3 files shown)


TheFakeElon:
> went through EVERY SINGLE thing in the machinery list

No it didn't. Typed list iterations use a tree search functionality, thankfully.

covertcorvid (approved):

PR #1711, opened 2021-10-25
Significant physics subsystem optimizations

TheFakeElon:
## About The Pull Request

Refactors a lot of collision handling and idle checks to be more performant.

I haven't done any profiling but this _should_ provide a noticeable performance increase during heavier overmap fights or any other situation with a lot of instanced colliders

## Why It's Good For The Game

The server should not turn into a turn based strategy once a few boxes start touching

## Changelog
:cl:
refactor: refactored the physics subsystem for performance
/:cl:


nsv13/code/modules/collision/shape.dm
```diff
@@ -66,15 +66,15 @@ Method to set our points to a new list of points
...
-	if(!src.base_points.len || src.base_points.len != points.len)
-		src.rel_points.Cut()
-		src.normals.Cut()
-		for (var/i in 1 to points.len)
-			src.rel_points.Add(new /datum/vector2d(0,0))
-			src.normals.Add(new /datum/vector2d(0,0))
-
-	src.base_points = points
-	src._recalc()
+	if(!length(base_points) || length(base_points.len) != length(points))
```
TheFakeElon, line 69:
```suggestion
	if(!length(base_points) || length(base_points) != length(points))
```

covertcorvid:
It looks okay but removes a lot of safety. I don't trust us that much, so we'll see how the runtimes look.

TheFakeElon:
It's still pretty safe, it's just not excessive anymore. You have to do something very obviously bad to break it

Bokkiewokkie:
Finally testmerged!

covertcorvid (approved):

PR #1719, opened 2021-10-31
Adds dynamic starmap loading (resolve #1716)

### LINKED ISSUE #1716: Hotswappable Starmap via Json
-SSstarsystem currently uses statically defined system types for its declaration.
-The system should be refactored to allow dynamic loading from disk, which will allow configurable maps & hot-loading with a view of (redacted).

Effort estimation: Low
ETA: 2d

Kmc2000:
## About The Pull Request

The starmap is no longer defined by static datums, and can instead be hotswapped out via JSON configs.
Maps can now define their own starmap type, this may come in handy for lowpop ships, if we ever want them to have a smaller / faster paced system map.

Mappers:
You can now declare a "starmap_path" key in your maps, IE:
starmap_path = "comedy.json"
To change the starmap that's loaded at roundstart.

You can also save the current game world (excluding random systems). For now, there is no interface for this until we figure out how (See: if) we want to go about persistence stuff.

Resolve: #1716 

## Why It's Good For The Game

Allows for dynamic starmap loading & saving as required.
More of a framework for future use. More to come! 

## Changelog
:cl:
add: The starmap can now be loaded and saved dynamically per round. 
/:cl:


nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,130 @@ Returns a faction datum by its name (case insensitive!)
...
+	try
+	{
+		_systems = json_decode(rustg_file_read(file(_source_path)))
+	}
+
+	catch(var/exception/ex){
+		//Fallback: Load the hardcoded systems and report an error.
+		instantiate_systems_backup()
+		CRASH("FATAL: Unable to load starmap from: [_source_path]. (Defaulting...): [ex]")
+	}
```
TheFakeElon, line 112: kmc no 😭😭😭

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,130 @@ Returns a faction datum by its name (case insensitive!)
...
+				fleet_type = LAZYACCESS(sys_info, "fleet_type") || null,
+				parallax_property = LAZYACCESS(sys_info, "parallax_property") || null,
+				visitable = LAZYACCESS(sys_info, "visitable") || TRUE,
+				is_hypergate = LAZYACCESS(sys_info,"is_hypergate") || FALSE,
+				preset_trader = (LAZYACCESS(sys_info,"preset_trader")) ? text2path(sys_info["preset_trader"]) : null,
+				system_traits = LAZYACCESS(sys_info,"system_traits") || list(),
+				system_type = (LAZYACCESS(sys_info,"system_type") && sys_info["system_type"] != "null" && sys_info["system_type"] != null) ? json_decode(sys_info["system_type"]) : list(),
+				audio_cues = (LAZYACCESS(sys_info,"audio_cues") && sys_info["audio_cues"] != "null" && sys_info["audio_cues"] != null) ? json_decode(sys_info["audio_cues"]) : list(),
+				wormhole_connections = (LAZYACCESS(sys_info,"wormhole_connections") && sys_info["wormhole_connections"] != "null" && sys_info["wormhole_connections"] != null) ? json_decode(sys_info["wormhole_connections"]) : list(),
+				startup_proc = LAZYACCESS(sys_info,"startup_proc") || null
```
TheFakeElon, line 141: you only need lazy access if you're using a lazy list (a list where you're unsure if it's initialized or not), checking it's existence and length every access just adds unneeded overhead

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -297,16 +411,50 @@ Returns a faction datum by its name (case insensitive!)
...
+		if("STARTUP_PROC_TYPE_BRASIL")
+			addtimer(CALLBACK(src, .proc/generate_badlands), 10 SECONDS)
+			return
+	message_admins("WARNING: Invalid startup_proc declared for [name]! Review your defines (~L438, starsystem.dm), please.")
+	return 1
+
+/datum/star_system/New(name, desc, threat_level, alignment, hidden, system_type, system_traits, is_capital, adjacency_list, wormhole_connections, fleet_type, x, y, parallax_property, visitable, sector, is_hypergate, preset_trader, audio_cues, startup_proc)
 	. = ..()
+	//Load props first.
+	if(name != null)
```
TheFakeElon, line 433: I don't know if it's faster but it's a byond proc so I'd assume so 🤔 
```suggestion
	if(!isnull(name))
```

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -297,16 +411,50 @@ Returns a faction datum by its name (case insensitive!)
...
 	var/dx = other.x - x
 	var/dy = other.y - y
 	return sqrt((dx * dx) + (dy * dy))
 
-/datum/star_system/New()
+/datum/star_system/proc/parse_startup_proc()
+	switch(startup_proc)
+		if("STARTUP_PROC_TYPE_BRASIL")
+			addtimer(CALLBACK(src, .proc/generate_badlands), 10 SECONDS)
+			return
```
TheFakeElon, line 426: switch statement with one condition smh

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,130 @@ Returns a faction datum by its name (case insensitive!)
...
+
+/datum/controller/subsystem/star_system/proc/save(_destination_path = "config/starmap/starmap_default.json")
+	message_admins("Saving current starsystem layout...")
+	var/json_file = null
+	//Read the file in...
+	try
+		json_file = file(_destination_path)
+	catch(var/exception/ex)
+		message_admins("WARNING: Unable to open [_destination_path]: [ex]")
+		return 1
```
TheFakeElon, line 176:
```suggestion
	json_file = file(_destination_path)
	if(!json_file)
		message_admins("WARNING: Unable to open [_destination_path]: [ex]")
		return 1
```

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,130 @@ Returns a faction datum by its name (case insensitive!)
...
+	var/list/file_data = list()
+	for(var/datum/star_system/S in systems)
+		if(S == null || istype(S, /datum/star_system/random))
```
TheFakeElon, line 179: Do systems remove themselves from the list when they die? not sure if they can be null
```suggestion
		if(isnull(S) || istype(S, /datum/star_system/random))
```

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,130 @@ Returns a faction datum by its name (case insensitive!)
...
+			"y" = S.y,
+			"parallax_property"=S.parallax_property,
+			"visitable"=S.visitable,
+			"sector"=S.sector,
+			"is_hypergate"=S.is_hypergate,
+			"preset_trader"=S.preset_trader,
+			"audio_cues" = json_encode(S.audio_cues),
+			"startup_proc" = S.startup_proc
+		)
+		file_data[++file_data.len] = entry
```
TheFakeElon, line 206:
```suggestion
		file_data += entry
```

TheFakeElon (commented):
try and catch isn't needed in most of the cases they've been used, you can accomplish the same thing by just checking the resulting variable (or the proc itself) and crashing if null :)

also CURLY BRACKETS BAD

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,130 @@ Returns a faction datum by its name (case insensitive!)
...
+		return
+	saving = TRUE
+	save()
+	saving = FALSE
+/**
+<summary>Save the current starmap layout to a json file. Used for persistence.</summary>
+<param></param>
+*/
+
+/datum/controller/subsystem/star_system/proc/save(_destination_path = "config/starmap/starmap_default.json")
```
TheFakeElon, line 168: this is very dangerous, the file path, type or proc caller aren't checked at all. And throw fdel into the mix... 💀

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: lazyaccess => l.hasKey(x) ? l[x] : null.
If the starmap designer omits keys, it's valid behaviour for these.

(Was my thinking)

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: I'm not sure what the byond jank is, but it's just checking if the pointer is 0 or not, I don't think we can really get much faster than that.

Language specifics may change this though

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: For future additions, but yeah defo a bit garbage

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: I didn't really want execution to halt here, allowing it to have a fallback to generate the static starmap.

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: Not sure either, but I don't trust my old code for a minute (never again)

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: Yeah god that's horrible you're right 😂

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: Weirdly, I found that didn't work.

I picked this trick up off monster, we have a load of it on our tgui backends. Is this incorrect?

nsv13/code/controllers/subsystem/starsystem.dm
TheFakeElon: it'd be a lot cleaner if you put the fallback generation in a separate proc

nsv13/code/controllers/subsystem/starsystem.dm
TheFakeElon: Nope, turns out you're totally correct!
http://www.byond.com/forum/post/1955485

Now that I recall, I tried to change one of the fighter(?) tgui data procs and broke it by doing this lmao

nsv13/code/controllers/subsystem/starsystem.dm
TheFakeElon: Given that variables can store any type of data, I would _assume_ an evaluation operator checking for null would be slower than a proc just made to check for null pointers. However, I don't know a lot about byond's compiler optimizations so it could just scrap it's data type checking for null checks for all we know.

Either way isnull() looks nicer but this is so pedantic that it doesn't matter :)

Bokkiewokkie: Can this entire check not just be if(name) though? 🤔

QAI-07:
Okay so apparently asteroids just don't spawn anymore. No clue what's causing it but it'd save us all a headache if we didn't have to manually spawn them in every time they're needed.

config/starmap/README.md
```diff
@@ -0,0 +1,76 @@
...
+- Escape all special chars (IE: "") with \
+
+## Format:
+
+The starmap json is simply a json encoded list of lists, where each list represents a system.
+Each individual list element is directly serialized into a starsystem object.
+
+An example system would be:
+
+```{"name":"Sol","desc":null,"threat_level":0,"alignment":"nanotrasen","hidden":0,"system_type":"{\"tag\":\"planet_earth\",\"label\":\"Planetary system\"}","system_traits":5,"is_capital":1,"adjacency_list":"[\"Alpha Centauri\",\"Outpost 45\",\"Ross 154\"]","wormhole_connections":"[]","fleet_type":"/datum/fleet/nanotrasen/earth","x":70,"y":50,"parallax_property":null,"visitable":0,"sector":1,"is_hypergate":0,"preset_trader":null,"audio_cues":"null"}```
```
4310V343k, line 32: wrong markdown
these ``` should be on different lines

nsv13/code/controllers/subsystem/starsystem.dm
Kmc2000: Yeah true say, just habits from work

nsv13/code/controllers/subsystem/starsystem.dm
TheFakeElon: They're not going to runtime if your key search doesn't find anything/returns null. You'd use `LAZYACCESS` if you weren't sure if `sys_info` exists or not, it doesn't change any behavior relating to the key

TheFakeElon: This is true for all lazy defines (that I know of), their main function is to dynamically initialize and check the existences of lists to save memory.

nsv13/code/controllers/subsystem/starsystem.dm
TheFakeElon: It looks like they do thankfully 😎

covertcorvid (changes requested):

covertcorvid:
An admin tool for changing the starmap file would be nice, but not a requirement at this point

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,149 @@ Returns a faction datum by its name (case insensitive!)
...
+		instantiate_systems_backup()
+		CRASH("FATAL: Unable to load starmap from: [_source_path]. (Defaulting...): [ex]")
+	}
+
+	for(var/i = 1; i <= _systems.len; i++)
+		//Try instancing this system from JSON, jump out if anything goes wrong.
+		var/list/sys_info = _systems[i]
+		try{
+			//Fields that must be present are accessed unsafely. The rest is fine if null.
+			var/datum/star_system/next = new /datum/star_system(
```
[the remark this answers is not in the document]
TheFakeElon: If a malicious person has access to the config you have much bigger problems to worry about lol

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,149 @@ Returns a faction datum by its name (case insensitive!)
...
+	try
+	{
+		_systems = json_decode(rustg_file_read(file(_source_path)))
+	}
+
+	catch(var/exception/ex){
+		//Fallback: Load the hardcoded systems and report an error.
+		instantiate_systems_backup()
+		CRASH("FATAL: Unable to load starmap from: [_source_path]. (Defaulting...): [ex]")
+	}
```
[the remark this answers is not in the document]
TheFakeElon: `CRASH()` doesn't actually crash the game, It does what you said. Saves a runtime log and exits the proc

Bokkiewokkie (dismissed):

nsv13/code/controllers/subsystem/starsystem.dm
Bokkiewokkie: ~~fixed~~ made even worse in the latest commit :)

nsv13/code/controllers/subsystem/starsystem.dm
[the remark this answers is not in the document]
Bokkiewokkie: Actually doesn't work if you just check whether the file exists, the try catch is still in but with game_log instead of CRASH aswell.

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -639,6 +802,16 @@ Returns a faction datum by its name (case insensitive!)
 /datum/star_system/proc/lerp_y(datum/star_system/other, t)
 	return y + (t * (other.y - y))
 
+//End the round upon entering O45.
+/datum/star_system/after_enter(obj/structure/overmap/OM)
+	if(CHECK_BITFIELD(system_traits, STARSYSTEM_END_ON_ENTER))
+		to_chat(world, "yeah [src.name]: [system_traits]")
```
DeltaFire15, line 808: Lets not leave in debugcode in here.

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -29,6 +34,8 @@ SUBSYSTEM_DEF(star_system)
 
 	var/obj/structure/overmap/main_overmap = null //The main overmap
 	var/obj/structure/overmap/mining_ship = null //The mining ship
+	var/saving = FALSE
+	var/foo = 0
```
covertcorvid, line 38:
```suggestion
```
Doesn't seem to be in use

nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -86,17 +91,145 @@ Returns a faction datum by its name (case insensitive!)
...
+	saving = FALSE
+/**
+<summary>Save the current starmap layout to a json file. Used for persistence.</summary>
+<param></param>
+*/
+
+/datum/controller/subsystem/star_system/proc/save(_destination_path = "config/starmap/starmap_default.json")
+	// No :)
+	var/list/directory = splittext(_destination_path, "/")
+	if(directory[1] != "config")
```
covertcorvid, line 167:
```suggestion
	if((directory[1] != "config") || (directory[2] != "starmap))
```

Please don't let them delete or overwrite *any* json in the config directory

covertcorvid (changes requested):
Please also remove the debug logging and combine the after_enter procs

Kmc2000:
> An admin tool for changing the starmap file would be nice, but not a requirement at this point

Will want to be its own issue, can backlog it.

Bokkiewokkie:
Apparently loading the starmap causes some issues with fleets moving: when a fleet leaves a system it will set its alignment to be unaligned instead of the initial. Probably want to make it so alignment is determined by beacons à la Stellaris outposts which the ship and some fleets could deploy.

DeltaFire15:
> Apparently loading the starmap causes some issues with fleets moving: when a fleet leaves a system it will set its alignment to be unaligned instead of the initial. Probably want to make it so alignment is determined by beacons à la Stellaris outposts which the ship and some fleets could deploy.

Starsystem alignments check against the initial faction of the system, except since it's loaded dynamically initial() probably returns unaligned since that uses the ''base'' var before any modifications.

Bokkiewokkie:
> > Apparently loading the starmap causes some issues with fleets moving: when a fleet leaves a system it will set its alignment to be unaligned instead of the initial. Probably want to make it so alignment is determined by beacons à la Stellaris outposts which the ship and some fleets could deploy.
> 
> Starsystem alignments check against the initial faction of the system, except since it's loaded dynamically initial() probably returns unaligned since that uses the ''base'' var before any modifications.

Yeah this seems to be it

Bokkiewokkie:
Merge conflicts my beloathed

PR #1732, opened 2021-11-19
CIC Reworks part 2: Tycoon

Absolutely-Steph:
## About The Pull Request
Reworks the tycoon CIC much like i did the atlas one, more consoles, bit different layout, yageddit.
Here is a picture of V1, for more recent ones see the mapdiffbot:

Screenshot 2021-11-19 100842

## Why It's Good For The Game
Adding more consoles means more information. Information is critical for good decisions. CIC is supposed to be where the captain makes decisions, so we need consoles.
Also adding the drawer means less paper clutter when CC inevitably sends 30 messages again.

## Changelog
tweak: tweaked the Tycoon CIC
/:cl:

### DIFF
--- a/_maps/map_files/Tycoon/Tycoon2.dmm
+++ b/_maps/map_files/Tycoon/Tycoon2.dmm
@@ -575,6 +575,12 @@
 	dir = 4
 	},
 /obj/machinery/computer/ship/tactical/console,
+/obj/machinery/button/door{
+	id = "gaussgang";
+	name = "public gauss access";
+	pixel_y = 26;
+	req_one_access_txt = "69"
+	},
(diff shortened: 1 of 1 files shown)


Karmic-Skink:
Here are a few points:
- The interior wall type mismatches with the rest of the ship 
- You have a wall mounted view screen sitting on a table in the center
- There are two inaccessible desks on the starboard side of the room

Here are a few questions:
- Why have the emergency supply closets been replaced with radiation closets?
- Why was the inducer removed?
- Why does the Bridge/CIC need all of these consoles?

Absolutely-Steph:
This was fun to make, it seems this was its only merit. It also looks better imo.

PR #1750, opened 2021-12-13
Adds a stationary AI flag & goal

DeltaFire15:
## About The Pull Request
Title. This AI flag should only be added to things one intends to only sit in one place and occasionally shoot things, as it'll override all other possible goals due to its own goal's high priority.
An overmap object with this flag will attempt to remain in place, maintain its original orientation (e.g. if it turned by getting rammed it'll try to turn back if the object can turn) and acquire nearby targets to shoot.
Nothing has this flag as of now, as e.g. traders do not have their AI enabled. Add this to stations that shouldn't move but should do stuff like use gun.

Also does some very minor improvements that aren't noteworthy which I noticed when looking at the files.

## Why It's Good For The Game
Requested by Karmic.

## Changelog
:cl:
add: Stationary AI flag & goal, intended for station use.
/:cl:


Bokkiewokkie (approved):

nsv13/code/modules/overmap/ai-skynet.dm
```diff
@@ -1348,15 +1348,23 @@ Seek a ship thich we'll station ourselves around
...
+		OM.seek_new_target()
+	OM.brakes = TRUE
+	
+
 /obj/structure/overmap/proc/choose_goal()
 	//Populate the list of valid goals, if we don't already have them
 	if(!GLOB.ai_goals.len)
-		for(var/x in subtypesof(/datum/ai_goal))
-			//I'll fix this jank later
+		for(var/x in (subtypesof(/datum/ai_goal) - typesof(/datum/ai_goal/human)))
```
covertcorvid, line 1366: It's probably more efficient to separately define this list than recreate it every time we update the goal

DeltaFire15: Yes thats is exactly how thst works. It checks if the list has not been setup and sets it up if thats the case.

covertcorvid (approved):

Issue #1734, opened 2021-11-20
AI can view but not manipulate any modular computers 


IndusRobot:
## Testmerges:

#1703 
#1692 
#1726 

## Reproduction:

Join as AI, then open up a modular computer, then try to access one of its programs

IndusRobot:
I might take a look at this later

PR #1752, opened 2021-12-17
Fixes AI interaction with modular computers

Bokkiewokkie:
## About The Pull Request

Makes get_overmap() check for the z level of the location of an atom instead of just the z level of the atom, because inside containers your z level is 0.

## Why It's Good For The Game

Fixes: #1734

## Changelog
:cl:
fix: Fixed modular computer interaction for AI units.
/:cl:


covertcorvid:
if loc is ever null this'll runtime

nsv13/code/__HELPERS/overmap.dm
```diff
@@ -1,12 +1,12 @@
 /atom/proc/get_overmap() //Helper proc to get the overmap ship representing a given area.
 	RETURN_TYPE(/obj/structure/overmap)
-	if(!z)
+	if(loc && istype(loc, /obj/structure/overmap))
+		return loc
+	if(!loc?.z)
```
TheFakeElon, line 5: if loc is null you'll probably want it to runtime
```suggestion
	if(isovermap(loc))
		return loc
	if(!loc.z)
```

TheFakeElon:
This'll work for the majority of cases however, it can still runtime if it's located in containers recursively (i.e modular computer inside a bag inside a closet). You could use get_turf() to get around this

nsv13/code/__HELPERS/overmap.dm
Bokkiewokkie: > if loc is null you'll probably _want_ it to runtime, as something else has gone catastrophically wrong

The problem is there's plenty of things in nullspace calling get_overmap() for some reason and I'm gonna have to fix all of them then...

Bokkiewokkie:
> This'll work for the majority of cases however, it can still runtime if it's located in containers recursively (i.e modular computer inside a bag inside a closet). You could use get_turf() to get around this

Tbh should just refactor most of this proc to not be several layers of lasange and spaghetti stacked on top of one another

nsv13/code/__HELPERS/overmap.dm
TheFakeElon: Better to have the runtime log than not though right? Null loc is (almost) always unexpected behavior

TheFakeElon:
> Tbh should just refactor most of this proc to not be several layers of lasange and spaghetti stacked on top of one another

#1689 :sob:

PR #1759, opened 2021-12-21
Attempts to resolve the frankly ungodly amounts of lag currently caused by stuff

DeltaFire15:
## About The Pull Request
Testmerge this before merging.

This PR attempts to cut down the enormous amount of lag, though some sacrifices had to be made.
Firstly, things no longer get sucked out through the ceiling if it is breached (air still leaves). While this is cool, this would lead to a thrown up -> openspace -> fall down -> repeat **infinite** loop, which is especially bad with auxmos which can handle more at once.
Secondly, this massively cuts down auxmos's equalize_hard_turf_limit from 2000 to 400. This will slow down atmos events, but at least not fully grind the server to a halt hopefully (from localtesting, a single breach was previously able to cause >120TD)
Thirdly, this applies a fix used by Putnam elsewhere in run_delay_heuristics, where a incorrect (1-x) instead of x caused auxmos to pause when not needed.

Also applies a few other fixes, mostly regarding multiz explosions, zfalling & turbolift runtimes.

## Why It's Good For The Game
image
image
image
Lag man BAD.

## Changelog
:cl:
code: Attempts to combat atmos / explosion / multiz lag.
fix: Turbolift wires no longer runtime.
fix: zfalls no longer fail to select a highest object.
fix: multiz explosions should now work less-bad.
/:cl:


covertcorvid:
bless

Bokkiewokkie (dismissed):

code/game/atoms_movable.dm
```diff
@@ -105,13 +105,17 @@
 	return !(movement_type & FLYING) && has_gravity(src) && !throwing
 
 /atom/movable/proc/onZImpact(turf/T, levels)
+	if(QDELETED(src))	//NSV13change - (if unlikely) safety
+		return	//I hate this
 	var/atom/highest = null
 	for(var/i in T.contents)
 		var/atom/A = i
+		if(QDELETED(A)) //NSV13change - (if unlikely) safety
+			continue
```
TheFakeElon, line 114: Needless overhead

code/game/atoms_movable.dm
```diff
@@ -105,13 +105,17 @@
 	return !(movement_type & FLYING) && has_gravity(src) && !throwing
 
 /atom/movable/proc/onZImpact(turf/T, levels)
+	if(QDELETED(src))	//NSV13change - (if unlikely) safety
+		return	//I hate this
```
TheFakeElon, line 109: can't happen, you'll get an error trying to call the proc in the first place

code/modules/atmospherics/environmental/LINDA_turf_tile.dm
```diff
@@ -131,8 +131,11 @@
 
 /////////////////////////GAS OVERLAYS//////////////////////////////
 
+//NSV13change - moves this proc up to /turf due to auxmos sometimes calling it on closed turfs and runtiming.
+/turf/proc/update_visuals()
+	return	//Avoid issues from closed turfs getting turf visuals updated (?? why is it even doing that)
```
TheFakeElon, line 136: auxmos shouldn't be updating visuals on closed turfs. If it is somehow, this is just fixing the symptoms not the root problem (and it's associated performance burden). Bee has had it on open only forever and it works fine

code/modules/turbolift/turbolift_machines.dm
```diff
@@ -83,14 +83,20 @@ GLOBAL_LIST_EMPTY(turbolifts)
 	if(locked)
 		return
 	locked = TRUE
-	wires.ui_update()
+	//NSV13change - prevents a rare runtime pre-wire-init.
+	if(wires)
+		wires.ui_update()
+	//NSV13change end.
```
TheFakeElon, line 89: you can bring it down to a one line diff if you use `.?`

TheFakeElon (commented):
All non-modular comments so far have been `//NSV` or `//NSV13`, it's easier to perform quick lookups if you keep in line with that instead of `//NSV13change`

code/modules/atmospherics/environmental/LINDA_turf_tile.dm
TheFakeElon: Would you happen to have the runtime log of update_visuals being called on closed turfs?

DeltaFire15: Auxmos callbacks, only thing I got was
image

DeltaFire15: (aware this only fixes the symptom, I just have absolutely zero idea what specifically causes it to also target /closed turfs)

Bokkiewokkie (approved):

DeltaFire15:
Please do not merge thi yet (though it doesn't appear to be *causing* any issues)
Have to investigate some more stuff that'd probably belong in this aswell when I have the time.

Bokkiewokkie (approved):

PR #1762, opened 2021-12-24
Munitions Machinery Maintenance Mechanics: Ammo Stackers/Sorters

Absolutely-Steph:
## About The Pull Request
Ammo Stackers/Sorters now require regular maintenance, if such is not provided, they may jam up and fail to load/unload their cargo.
All information needed is conveyed to the player by examining the machine.
Unclear jams using a crowbar.
Perform maintenance by opening it up with a screwdriver and applying oil.

Features discussed that are missing in the initial implementation:
- [x] durability loss when loading
- [x] randomized jamming based on durability
(if i add them later I'll tick them here.)

## Why It's Good For The Game
After talking to munitions mains about what they think of the current gameplay, they said they'd like maintenance back.
I personally believe this to be merely busywork, but i can see how this gives them something to do.

## Changelog
:cl:
add: Added maintenance need to ammostackers
balance: balanced ammostackers maintenance decay to require maintenance roughly every 50-100 uses.
soundadd: added jam/unjam sounds to ammostackers
spellcheck: fixed a typo in the ammostackers examine text
/:cl:


nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/revision2/automation.dm
```diff
@@ -356,25 +412,46 @@
...
+
+/obj/machinery/ammo_sorter/proc/weardown()
+	if(durability > 0) //don't go under 0, that's bad, this whole if statement handles specifically the wear&tear and it's jamming at 0% durability.
+		durability -= rand(0,3) //pulling stuff out wears it down, now randomly. Feel free to tell me how this is done correctly.
+	else 
+		jammed = TRUE // if it's at 0, just kinda jam it.
+		durability = 0 // in case an admin plays with this and doesn't know how to use it, we reset it here for good measure.
+	jamchance = ((durability - 100) * (-1)) / 4 //inverting the 100-0 durability into a 0-100 probability to jam, then halving it
+	if(prob(jamchance))
+		jammed = TRUE
```
Karmic-Skink, line 457: Line ending.

Absolutely-Steph: i'm sorry, please elaborate. do you want me to add a new line at the end?

Absolutely-Steph: consulting with other sources has confirmed my suspicions about this cryptic review. I have added a new line.

TheFakeElon:
>  durability loss with time passed

This sounds really annoying

Absolutely-Steph:
> > durability loss with time passed
> 
> This sounds really annoying, significantly moreso than the rest at least. It doesn't make much sense either

oh yeah it was just something they brought up when we were first throwing ideas at the wall, frankly i don't think it's a good idea, but i can see how that would require them to maintain it outside combat too.
nothing in that list was really "planned" to be added, i just did the others because they seemed to be good enough.

From my point of view, this PR is "done", barring a QA pass from the maintainers ofc.

covertcorvid:
We're going to need some changes to Shrike if we do this to ammo sorters, since the ammo sorter room is MAA access
@Karmic-Skink

Bokkiewokkie:
I don't think this is a good idea. Adding maintenance to what is essentially a protective bomb storage locker (which likely will only become a real problem when it is needed most, and sounds like it would be severely detrimental to munitions gameplay fun levels) seems unnecessary. The spelling mistake fix is good though.

Absolutely-Steph:
> I don't think this is a good idea. Adding maintenance to what is essentially a protective bomb storage locker (which likely will only become a real problem when it is needed most, and sounds like it would be severely detrimental to munitions gameplay fun levels) seems unnecessary. The spelling mistake fix is good though.

This entire PR only happened because some muni mains told me it would be fun, i'd say we at least give it a try. If this turns out to not be good, i still plan to add maint to NACs itself at a later date.

Absolutely-Steph:
> Sorters would be useful for letting cargo dump lots of crates down a chute to have munitions figure out, if set up properly. It was an idea I toyed with long ago but didn't really finish the thought process for

sorters aren't covered by this pr(as in disposal sorters, if you meant those), this pr only covers the "boxes" that you can insert ammo and powder and such into, i just used stackers/sorters/whatever because i have no idea what their actual name is, in-code they're sorters, spy calls them stackers, no clue

Cyaneer:
Alright, now that it was CLANGED on an actual round, here is the main funny:

- For proccing jams, it counts EACH AND EVERY MOVEMENT INTO THE SORTER - INCLUDING MOBS VOLUNTARILY RUNNING INTO IT - and each single item being moved into it, not even checking if the item is an actual munition.
- Extra points that if an item doesn't get into the sorter, it continues proccing further movements, so any item on a conveyor not going into the box is a guaranteed jam.

This mostly means that the sorters will be jammed most of the time, period.

Cyaneer:
If it wasn't obvious, I'm proposing a check that only tries to load an item into the rack if it's a munition, which removes CLUNK and makes jams do the intended function.

covertcorvid (approved):

PR #1777, opened 2022-01-06
"Fixes" a runtime involving putting a fox in a box

IndusRobot:
## About The Pull Request

Fixes a runtime in `GetSafeLoc` causing CI to fail when specimen objectives are rolled. The issue stems from specimen cargo objectives where the following steps are performed in code
- creating a large wooden cargo crate (`/proc/deliver_package` in `freight_type.dm`), 
- then adding the specimen to the contents of the box (`/proc/deliver_package`), 
- then creating a supply pod to send to the shipping lane (`/proc/send_supplypod` in `traders.dm`)
- then using `forceMove` to move the large wooden cargo crate into the supplypod (`/proc/send_supplypod`),
- then checking its zlevel to make sure it's not on the overmap, as a result of calling `forceMove` (`/proc/GetSafeLoc` in `overmap.dm`)

Why `GetSafeLoc` is being called with no atom to check, I don't know and honestly don't really care. The fox does not care, it shows up in cargo as it should, with and without the runtime 

## Why It's Good For The Game

fox cargo 

## Changelog
:cl:
fix: Fixes a runtime in GetSafeLoc when passing no atom 
/:cl:


Karmic-Skink (changes requested):
This needs a changelog.

Bokkiewokkie (approved):

PR #1794, opened 2022-01-23
Tweaks overmap interior explosions on destruction

DeltaFire15:
## About The Pull Request
This changes the explosion that happen when a ship with an interior is destroyed to be 7-0-0 (deva-heavy-light), instead of 30-30-30. Now, this 30-30-30 was actually 4-8-16 because it didn't have ignorecap, so the real issue was the 1024 tiles of light explosion per explosion in every area in one operation.

Now it instead has 196 tile pure-deva explosions, which should still do more damage than the 4 radius deva ones before, with a 3 second wait before it triggers another (could probably juse use a check_tick instead, either should be fine but slow explosions is kinda neat.

## Why It's Good For The Game
Currently, the server just freezes when the main ship dies due to it trying to explode 100000 tiles in one tick. This makes it more managable of a load.

## Changelog
:cl:
tweak: Overmap interior explosions when they are destroyed should now be less laggy (primary for ones with more areas)
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/ai_interiors.dm
+++ b/nsv13/code/modules/overmap/ai_interiors.dm
@@ -32,11 +32,13 @@
 	special()
 
 /proc/overmap_explode(list/areas)
+	set waitfor = FALSE	//Lets not freeze everything
 	if(!areas)
 		return
 	for(var/area/AR in areas)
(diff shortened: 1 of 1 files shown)


Bokkiewokkie (approved):
ship_go_byebye

covertcorvid (approved):

PR #1820, opened 2022-02-12
(VERY HYPOTHETICAL AND SUBJECT TO BALANCE) Stamina System

Absolutely-Steph:
## About The Pull Request
**DO NOT MERGE THIS WITHOUT ADJUSTING THE RUN DELAY TO ~0.5, AND THE WALK DELAY TO ~2.
I'M LOOKING AT YOU BOKKIE, YOU BETTER ADJUST THEM.**
This adds a sort of stamina system using bee's native stamina damage stat, here's what I want out of it:
- Generally everyone should walk by default.
- When needing to go somewhere quickly, running is still your best option.
- Running 24/7 will tire you out until you collapse.

Here's how it works in detail:
Increased run speed first and foremost, it's gotta be worth the trouble.
When running, take slight stamina damage with every step.
When over 50 stamina damage, begin taking slight oxy damage and gasping.

Additional things(maybe, probably?) to come later could be;
- Hardsuits affecting it.
- A quirk you can take to have more stamina, but it makes you have to eat more
- stamina bonus for certain jobs?(cargotechs, paramedics)
- adding disgust gain when fat?
- fat characters have less stamina?

### IT'S ONLY 0.5 LESS SPEED THAN USUAL, YOU'RE NOT THAT MUCH SLOWER
- Adds depth to Chases, adding the option to tire out your pursuer
- Being in walk intent most of the time results in more blocking, did you know that wrenches can block stuff? no? Oh because NO ONE EVER WALKS!?
- Adds depth to chemical stimulant choices, you now have to decide whether you like meth's speedboost or ephedrines stamina enhanincing more, both leading to very different tactical options.
- Deciding whether to go someplace quickly or not enables you some beneficial stuffs, like having to consider if you want to conserve stamina when running to apprehend a criminal.
- Fight dynamics will likely change to be much more interesting.

## Changelog
:cl:
add: Stamina System
/:cl:

### DIFF
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -136,7 +136,14 @@
 		move_delay = old_move_delay
 	else
 		move_delay = world.time
-
+	//NSV13 START STAMINA SYSTEM
+	if(L.m_intent == MOVE_INTENT_RUN)
+		if(L.getStaminaLoss() > 50)
+			if(prob(25) && iscarbon(mob))	
+				L.emote("gasp")
+			L.adjustOxyLoss(1)
+		L.adjustStaminaLoss(1)
+	//NSV13 END
 	if(L.confused && L.m_intent == MOVE_INTENT_RUN && !HAS_TRAIT(L, TRAIT_CONFUSEIMMUNE))
 		var/newdir = 0
 		if(L.confused > 40)
--- a/code/modules/mob/mob_movement.dm
+++ b/code/modules/mob/mob_movement.dm
@@ -139,7 +139,7 @@
 	//NSV13 START STAMINA SYSTEM
 	if(L.m_intent == MOVE_INTENT_RUN)
 		if(L.getStaminaLoss() > 50)
-			if(prob(25) && iscarbon(mob))	
+			if(prob(10) && iscarbon(mob))	
 				L.emote("gasp")
 			L.adjustOxyLoss(1)
 		L.adjustStaminaLoss(1)


IndusRobot (commented):
Has this been tested with items that boost speed? Drugs like ephedrine and meth, or items like bluespace tiles and wings of flight (we rarely ever see these but it's still attainable through xeno)? Would these cause the person to collapse faster for procing on each Move?

Absolutely-Steph:
I tested it with a few different chems, here are my results:
Meth(30u):

- run stupidly fast, for quite a long time(like usual), but after a while you start collapsing. This isn't that much of an issue since you regen stamina damage more quickly too, but the oxy damage starts catching up with you after collapsing a few times, leading to you shambling along after running around the tycoon's bridge for approx 3 laps.
- this can be avoided by walking, but while walking on meth is still faster than without meth, running without meth is still faster, allowing for people to catch up to you as long as you stick to walking.
- I can imagine that skillfully mixing walking and running sprints will lead to quite the dastardly hard criminal to catch, though it still takes more skill than default meths constant high speed.

Stimulant pill(ephedrine, coffee, antihol):

- Might be the ephedrine doing the most work, it allows you to run until you quite literally collapse choking for a bit.
- Not much more speed, though the large increase in stamina and seemingly crackheaded ability to keep on running while choking makes for a fun mix to witness.

Combat Stims:
- a middle of the road option it seems, this doesn't increase your speed that much, but it negates a lot of the slowdown of bad stamina, allowing you to run at faster-than-average for a good long while.

Overall it seems this has brought chems into two categories, those that increase stamina and those that increase speed, I very much like this finding.

Absolutely-Steph:
For anyone that didn't yet realize this, despite me very clearly saying so multiple times:
This doesn't actually slow people down that much, it mostly makes walk intent the default(now nearly as fast as previous running)
and makes running a very fast, but limited way to get somewhere.
Ideally i'd have this slow people down substantially, but people are already whining with only a 0.5 slowdown, so i won't.

TheFakeElon:
just do what goon did and have an sprint button if you really want stamina movement. This is terrible in all aspects of the word

Absolutely-Steph:
> Do what goon did and have a sprint button if you really want stamina movement. I can't see a single upside to slapping a stamina penalty on the movement mode 99% of people use 99% of the time other than maybe appeasing a hyper realist HRP player or two. Walking from point A to point B doesn't add anything in itself.
> 
> Additionally, stamina management is almost never enjoyable, especially for something as passive as running. The added so-called "depth" is holding shift until stamina bar runs out, move slowly while it recharges, begin holding shift again.
> 
> I've never seen a game that locks running behind a stamina wall save for one or two heavy simulation games, sprinting is sometimes based on stamina but only when running is sufficiently fast enough to quickly get around.
> 
> > IT'S ONLY 0.5 LESS SPEED THAN USUAL, YOU'RE NOT THAT MUCH SLOWER
> 
> This is very misleading, a lie even. "only 0.5" is 33% slower. For comparison walking is usually 50% slower than running 👀

Didn't feel that slow to me in tests but okay, this meme was long over and never had any prospects to be merged anyways. Even if i think that this, or something like it, would be quite interesting to have in game.
Still might wanna think about slowing run speed down a tad, it's a bit high right now i feel.

PR #1471, opened 2021-06-21
Adds jetpacks back to the mining vendor

bluey89962:
## About The Pull Request
Decided to split my last PR in two, this one is only regarding the jetpack upgrade.

## Why It's Good For The Game
quoting from the original PR:

> Bee mining is very different from NSV mining, the rebase was bound to have some changes that are unecessary in here. Jetpacks, while useless in Bee's Lavaland, are incredibly useful to fly around asteroids.

in other words jetpack make miner go :)

## Changelog
:cl:
add: Jetpack upgrade is back on the mining vendor. Rejoice!
/:cl:

PR #1833, opened 2022-02-18
Brings back the mining jetpack upgrade

someone543:
## About The Pull Request
Adds the jetpack upgrade to the mining equipment vendor

## Why It's Good For The Game
Jetpacks were originally removed by a beebase for having little use for mining in lavaland and was considered to be a niche item. This is not a problem for NSV mining as having the ability to fly around asteroids which typically have no gravity would be very useful. 

## Changelog
:cl:
add: Brings the jetpack upgrade back to mining vendors
/:cl:


Karmic-Skink:
Adding something like this while we are in the middle of a mining rework is rather foolhardy.

Bokkiewokkie:
This has also already been rejected in several past PR's and I was doing other stuff so I wanted to be clear this wouldn't make it in

Pontenerd:
> This has also already been rejected in several past PR's and I was doing other stuff so I wanted to be clear this wouldn't make it in

Which PR's?

Bokkiewokkie:
So yes this change has already been denied a few times in these past PR's:
#1471 and #1341
As it stands we're not looking to deviate more from bee in this, though adding jetpacks back to mining is something people are clearly interested in, and as such will be talked about in relation to the mining rework we're currently working on.
Basically, you'll be without a jetpack (though you could always ask for a proper one from EVA storage or build your own.) for now, until we've implemented a better way of rewarding miners for their hard work.

PR #1840, opened 2022-02-21
Race Specific Emotes

BetaCentari:
## About The Pull Request
Adds several new emotes. Some of which are specific for individual races. 
Tweaks a define to be like the rest of them. 

Adds more variability to the races of NSV, allowing for more expression and accentuation of roleplay. 

## Changelog
:cl:
add: Emotes for several races
add: New wake-up sound effect for IPCs
tweak: Tweaked the felinid iscatperson define to work like the rest of them. 
soundadd: Dialup sound effect.
soundadd: The purr of a headcoder's cat.
/:cl:


Absolutely-Steph:
I can't help but think that these sounds don't really fit in with the other emotes, I'd personally like it more if they were like 5 seconds long at most.
A singular purr instead of drawn out purring, maybe only the most recognizable section from the dialup.
Have you also thought about changing the *scream sound for other races instead of just moths/lizards? I can imagine cat screams would be quite funny, maybe IPC screams could be the dialup instead?

BetaCentari:
> I can't help but think that these sounds don't really fit in with the other emotes, I'd personally like it more if they were like 5 seconds long at most. A singular purr instead of drawn out purring, maybe only the most recognizable section from the dialup(though it might fit better like this, since it's not really an emote). Have you also thought about changing the *scream sound for other races instead of just moths/lizards? I can imagine cat screams would be quite funny, maybe IPC screams could be the dialup instead?

That's kind of the point. Is that these emotes stand out from the rest as race specific and more accurate to the species the player is. Lizards should be able to hiss. Ethereals don't have a tongue but have a specific electrical organ in their mouth that vibrates electrical frequencies to make sound and language like voltaic which is just a bunch of zappy sounds. None of the emotes are longer than 5sec, I'm also limited by the sounds that come stock with NSV or what I can acquire without breaking copyright. That purr is from a headcoders cat which I thought would be neat but was difficult to clean up and make it sound nice. And there's no way I'm letting IPCs make dialup sounds on command. I've already gotten threats that if they could they'd spam it which is why it plays when they come back to life after being repaired. And I have certainly considered doing more with this PR. If anything this is just the beginning and it opens up the possibilities and suggestions for more stuff like this. I can shorten the dialup sound effect if needed but I thought it'd be better as the whole thing.

Bokkiewokkie:
Why is this not a PR on beestation instead of NSV?

BetaCentari:
The simplest answer is that I don't know anyone there or procedure for
upstream code modification or PRs and I'm scared to do so. Wouldn't it take
a while for upstream code to make it us anyway?

covertcorvid:
I can't imagine getting all the different race maintainer permissions for this on Bee. maybe if people like it they'll port it from us lol

Crossedfall:
> I can't imagine getting all the different race maintainer permissions for this on Bee. maybe if people like it they'll port it from us lol

What? You just need to ask a single maintainer if it's OK. You don't need several individuals signing off on it, that'd be silly.

I'll give my sign off. This is something that would ideally go upstream initially.

Bokkiewokkie (changes requested):
Everything that *isn't* related to catpeople should probably be ported upstream as well.

Bokkiewokkie (approved):

PR #1850, opened 2022-03-07
Adds drone dispensers to some ships

Pontenerd:
## About The Pull Request

Adds drone dispensers to the atlas, galactica and tycoon

## Why It's Good For The Game

Some players expressed an intrest in seeing drones added again, and I think NSV is mature enough to be trusted with drones (and i trust our admin team to remove those who are not.)

## Changelog
:cl:
add: Adds drone dispensers to the robotics department on some vessels

/:cl:


Karmic-Skink:
We intentionally do not have drone dispensers in our maps.

Pontenerd:
> We intentionally do not have drone dispensers in our maps.

Screenshot_20220307-224323

I checked with bokkie beforehand, he said I could go for it

Karmic-Skink:
> > We intentionally do not have drone dispensers in our maps.
> 
> Screenshot_20220307-224323
> 
> I checked with bokkie beforehand, he said I could go for it

Cool, don't care.

Pontenerd:
> We intentionally do not have drone dispensers in our maps.

Can you elaborate on why this is?

Bokkiewokkie:
> > We intentionally do not have drone dispensers in our maps.
> 
> Can you elaborate on why this is?

Well I definitely can, and I also would like to ask if you can refrain from using my discord messages without their context to justify things. I never stated outright this change would be approved in any way.
Anyways, the main reason we don't have drone dispensers anymore is the amount of people that straight up abused them or otherwise misbehaved as a drone. This made us decide the best thing would be to focus on other ways of getting people back in the game rather than dealing with the administrative nightmare drones are.
Referring to my previous statements about the drone dispensers: I said it *could* work *if* people actually did what they were supposed to as a drone, which is something we'll either have to find out after some testing or after making changes, which we would do only if we decide we do want to keep them as an active part of the game anyways.

Pontenerd:
> and I also would like to ask if you can refrain from using my discord messages without their context to justify things. I never stated outright this change would be approved in any way besides that one (arguably vague) passing remark in a conversation about drones.

My appologies, that was a misinterpretation on my behalf. 

>  I said it could work if people actually did what they were supposed to as a drone, which is something we'll either have to find out after some testing or after making changes
I agree, I was hoping to see them in use first to help determine what modifications were needed, I have a few ideas which i'll go through now.

idea 1: make drones completely unable to interact with other players
While this would solve the issue of drones metagrudging people, it would make them isolated, kinda ruining the point of playing on a server with other people

Idea 2: Make drones able to speak and understand binary, and have them be similar to borgs
I think this idea is more ideal solution as it would bring them more in line with borgs with a positronic brain, however this may end up detracting from the use of positronic brains.

If anyone has any ideas please write a comment and suggest it

Bokkiewokkie:
> I agree, I was hoping to see them in use first to help determine what modifications were needed, I have a few ideas which i'll go through now.
> 
> idea 1: make drones completely unable to interact with other players While this would solve the issue of drones metagrudging people, it would make them isolated, kinda ruining the point of playing on a server with other people
> 
> Idea 2: Make drones able to speak and understand binary, and have them be similar to borgs I think this idea is more ideal solution as it would bring them more in line with borgs with a positronic brain, however this may end up detracting from the use of positronic brains.
> 
> If anyone has any ideas please write a comment and suggest it

So one of the main issues was the fact that drones actually have hands, meaning they can stow half of the ship into their backpack like the true greytider bots they are at heart. The best solution then seems to practically make them engieborgs... which we already have... or making it so they can only pick up a limited selection of items, which is kind of dumb IMO.
One idea is adding drones as a cheaper variant of a borg, still made in robotics, using the same laws and having the same relation to the AI as borgs.

Pontenerd:
> So one of the main issues was the fact that drones actually have hands, meaning they can stow half of the ship into their backpack like the true greytider bots they are at heart

I feel like this could be solved by removing their massive duffles (iirc it's larger then even a BoH?) and replacing it with a normal backpack (or equivalent) 

> One idea is adding drones as a cheaper variant of a borg, still made in robotics, using the same laws and having the same relation to the AI as borgs. 

seems like an intresting idea, might be worth further consideration

Vasily2013:
Pending changes to drones as a balance and administrative issue I don't think this is going anywhere. Please correct me if I'm wrong, @Bokkiewokkie. 

From a mapping standpoint there likely isn't any issue, but, there are obvious balance and administrative concerns that need to be addressed before implementation of this.

Bokkiewokkie:
Yeah I'm gonna say no to this for now, if we want to do this we'll revisit it later.

PR #1853, opened 2022-03-13
AI ships no longer try to predict shots when using hitscan weapons

DeltaFire15:
## About The Pull Request
Title. Effectively, AI ships with hitscan weaponry currently have stormtrooper aim, as they'll try predicting the ship's movement, and subsequently almost always miss because well, the projectile is instant and the ship isn't.
This should make them as scary are they should be.

## Why It's Good For The Game
Hitscan weapons don't need any trajectory prediction as they're hitscan.

## Changelog
:cl:
tweak: AI hitscan weapons aim dead center instead of trying to predict trajectory (and missing)
/:cl:


Pontenerd:
Does this mean AI using hitscan weapons will hit 100% of the time?

DeltaFire15:
> Does this mean AI using hitscan weapons will hit 100% of the time?

Yes. It's very difficult to miss with an oversized laserpointer and on par with their intended dangerousness (for the very few ships which possess such kinds of weapons).

covertcorvid (approved):

Karmic-Skink:
Basically this should only be affecting SolGov ships at this stage.

PR #1854, opened 2022-03-14
Kills the asteroid chasm (maybe this time for real)

DeltaFire15:
## About The Pull Request
The rivers on that one asteroid ruin have chasms as basetype, which means if a rock spawns ontop of them and then is destroyed, there'll now be a chasm. Needless to say, chasms shouldn't be existing on asteroids.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Water turfs no longer turn into chasms if a rock spawns ontop them.
/:cl:

### DIFF
--- a/code/game/turfs/open/water.dm
+++ b/code/game/turfs/open/water.dm
@@ -4,7 +4,7 @@
 	desc = "Shallow water."
 	icon = 'icons/turf/floors.dmi'
 	icon_state = "riverwater_motion"
-	baseturfs = /turf/open/chasm/lavaland
+	baseturfs = /turf/open/floor/plating/asteroid/airless //NSV 13 - rocks don't have chasms and no cases currently exist where they'd make sense.
 	initial_gas_mix = OPENTURF_LOW_PRESSURE
 	planetary_atmos = TRUE
 	slowdown = 1


Pontenerd:
how many times has this been fixed now? 3?

covertcorvid:
I attempted to edit this variable on the map itself and it doesn't seem to have helped but we shall see

Bokkiewokkie:
I already know for a fact this will have no effect because it is not the cause of the issue we currently still face

DeltaFire15:
> I already know for a fact this will have no effect because it is not the cause of the issue we currently still face

Location of chasms during prior round was the same as that of previous water turfs where rock turfs were placed onto by the ruin generation. Additionally, rock turfs placed where water would've been had chasms in the base_turfs list. Please specify what makes you so sure this will not assist the situation.

DeltaFire15:
Localtesting is yielding expected results.

covertcorvid:
https://github.com/BeeStation/NSV13/blob/054da621241a7f9ce18e7563cd37ee14c62f8a72/_maps/map_files/Mining/nsv13/ruins/mining10.dmm#L28

if your change works and mine doesn't then I understand nothing

DeltaFire15:
> https://github.com/BeeStation/NSV13/blob/054da621241a7f9ce18e7563cd37ee14c62f8a72/_maps/map_files/Mining/nsv13/ruins/mining10.dmm#L28
> 
> if your change works and mine doesn't then I understand nothing

Turf code has a bunch of initial()s used which may have been overruling the change done as a map-varedit.

Bokkiewokkie (approved):
I guess this did actually work, I owe you an apology for doubting it.

PR #1874, opened 2022-03-28
random optimizations and code cleaning

TheFakeElon:
## About The Pull Request

Cleaned up some stuff while I was looking for an asteroid bug (which I didn't end up finding)

## Changelog
:cl:
code: Misc optmizations
/:cl:

### DIFF
--- a/nsv13/code/controllers/subsystem/starsystem.dm
+++ b/nsv13/code/controllers/subsystem/starsystem.dm
@@ -87,7 +87,7 @@ Returns a faction datum by its name (case insensitive!)
 	RETURN_TYPE(/datum/faction)
 	if(!name)
 		return //Stop wasting my time.
-	for(var/datum/faction/F in factions)
+	for(var/datum/faction/F as() in factions)
 		if(lowertext(F.name) == lowertext(name))
 			return F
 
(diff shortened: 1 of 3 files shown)


nsv13/code/controllers/subsystem/starsystem.dm
```diff
@@ -87,15 +87,15 @@ Returns a faction datum by its name (case insensitive!)
 	RETURN_TYPE(/datum/faction)
 	if(!name)
 		return //Stop wasting my time.
-	for(var/datum/faction/F in factions)
+	for(var/datum/faction/F as() in factions)
```
IndusRobot, line 90: What does `as()` do that's more efficient than `in`? I commonly use for in statements in my code and would like to know more

TheFakeElon: By default, `as` is set to the type of the variable in the for loop. So in this case (pre-optimization), the loop would implicitly be
```for(var/datum/faction/F as(datum/faction) in factions)```

the `as` argument is used to enumerate over only the target type/subtypes via a tree lookup(?) 
`as(anything)` and `as()`, overrides the implicit lookup type and allows the loop to enumerate over it's contents without any sort of type checks, significantly increasing it's speed. Though this obviously only works so long as your target list doesn't contain any types adjacent or above your variable type

IndusRobot: Good to know for iterating over lists with contents of a strict type. Thanks!

Bokkiewokkie (approved):

PR #1878, opened 2022-04-01
WIP - Wire Guided Torpedoes

Karmic-Skink:
## About The Pull Request
This PR introduces 'Wire-Guided' Torpedoes for use by big ships presently, and small ships later (medium_fighter).

Currently this feature is only available on the ATLAS, and may be rolled out to future ships in time.

Todo:
- [ ] Care about NSV13
- [ ] Finalize Silo Sprites
- [ ] Implement AI using these torps
- [x] Update AMS to target these torps
- [ ] Fun TM

KNOWN BUGS:
- Sometimes you don't get returned to the Tactical console after a torpedo explodes and just locks out the TAC
For your convivence, a screwdriver has been included next to the TAC console to assist resolving any lockout issues.

## Why It's Good For The Game
Looking at improving torpedoes to be an alternative to NAC as a main weapon.
If not, maybe just MEDIUM_FIGHTER.

## Changelog
:cl:
refactor: Completely rebuilt torp system for testing
refactor: Reworked Atlas munitions bay
/:cl:


red031000:
since apparently nobody has reported here, apparently firing the torps forward oh atlas/testlas causes them to hit the ship itself

Karmic-Skink:
> since apparently nobody has reported here, apparently firing the torps forward oh atlas/testlas causes them to hit the ship itself

As per progress and the above push, I have changed the arming time to 3 seconds.
If this continues I will increase it further.

red031000:
ok, now that I have some exprience I feel I can share feedback

overally it's an... interesting system, it does change it up quite a bit

- at the moment it feels a bit... jank tbh, feels almost like this should be separate from the tac through another console with another BO? I can see a lot of people not using it at all because it takes the focus away from the ship, and they'd prioritise NACs
- when zoomed out, firing a torpedo zooms back in, same by refocussing on the ship, it probably shouldn't do that
- torpedos just... disappear after a while, regardless of whether it's been hit, whether you've hit the target or not, or anything
- some targets are extremely difficult to hit, interceptor's hitboxes were always a bit fucked, but this is more obvious with this, and smaller ships are also difficult to hit
- the above point is made worse by the no inertial assistence, but that may just be a skill issue

hope this is helpful, will post more as I have more to say

Karmic-Skink:
> ok, now that I have some exprience I feel I can share feedback
> 
> overally it's an... interesting system, it does change it up quite a bit
> 
> * at the moment it feels a bit... jank tbh, feels almost like this should be separate from the tac through another console with another BO? I can see a lot of people not using it at all because it takes the focus away from the ship, and they'd prioritise NACs
> * when zoomed out, firing a torpedo zooms back in, same by refocussing on the ship, it probably shouldn't do that
> * ~torpedos just... disappear after a while, regardless of whether it's been hit, whether you've hit the target or not, or anything~ EDIT: I apparently can't read and this is due to the 30 second fuse
> * some targets are extremely difficult to hit, interceptor's hitboxes were always a bit fucked, but this is more obvious with this, and smaller ships are also difficult to hit
> * the above point is made worse by the no inertial assistence, but that may just be a skill issue
> 
> hope this is helpful, will post more as I have more to say

Honestly I'm very wary of giving the players the ability to have an addition main weapon being used through an extra console by a player outside of the TAC position.
Which is why I kinda figure this might just be better on MEDIUM_FIGHTER as it doesn't really have a huge DPS potential atm, but quite a lot of range.

I'll look into the zooming issue, I was aware of it, but base functionality took priority.

I'm honestly not too concerned about not easily being able to hit smaller hit boxes, ie hitting a fighter with an anti-capital torpedo?

INS or No INS. That remains to be worked out.

red031000:
I should clarify, my above point about hard to hit targets is not just fighters, it's also smaller ships like the marine/boarder ships, those are also fairly difficult to hit

red031000:
the normal torp tubes seem to be broken nowl, this is a problem for firing stuff like science probes

Karmic-Skink:
> the normal torp tubes seem to be broken nowl, this is a problem for firing stuff like science probes

No big loss there. I didn't actually touch torp code, only the weapons assigned to Testlas.

red031000:
yeah since the normal torp tubes are not assigned as weapons, you can't fire them, it does present an issue, as science probes also do not work with the wire torpedo tubes, they don't hit the wormhole at all, they just go straight through them

Karmic-Skink:
> yeah since the normal torp tubes are not assigned as weapons, you can't fire them, it does present an issue, as science probes also do not work with the wire torpedo tubes, they don't hit the wormhole at all, they just go straight through them

Just ignore it for the moment. I disabled it on testlas to avoid confusion of weapon selection.
Probes even still being attached to that is dumb and will either be fixed or removed. Like all of science should be.

ta3370:
> yeah since the normal torp tubes are not assigned as weapons, you can't fire them, it does present an issue, as science probes also do not work with the wire torpedo tubes, they don't hit the wormhole at all, they just go straight through them

In meantime, I've found Heavy Fighters can be rather effective, or actually more effective, for wormhole research.

Crus0:
The wire guided VHS tubes do not behave life standard ones as far as collision goes. Normal vhs tubes can be walked over without trouble, but the wire guided ones block movement. Is this intentionall?

Karmic-Skink:
> The wire guided VHS tubes do not behave life standard ones as far as collision goes. Normal vhs tubes can be walked over without trouble, but the wire guided ones block movement. Is this intentionall?

Yes. It is a giant hole for putting a torp in. Why would you be able to stand on it?

Crus0:
> > The wire guided VHS tubes do not behave life standard ones as far as collision goes. Normal vhs tubes can be walked over without trouble, but the wire guided ones block movement. Is this intentionall?
> 
> Yes. It is a giant hole for putting a torp in. Why would you be able to stand on The earlier vhs tubes allowed it?

The previous tubes allow it. I suppose one just stands on the edges? The wire guided VHS tubes also prevents torps from passing over or falling into them, which definitely doesn't make sense.

Karmic-Skink:
> > > The wire guided VHS tubes do not behave life standard ones as far as collision goes. Normal vhs tubes can be walked over without trouble, but the wire guided ones block movement. Is this intentionall?
> > 
> > 
> > Yes. It is a giant hole for putting a torp in. Why would you be able to stand on The earlier vhs tubes allowed it?
> 
> The previous tubes allow it. I suppose one just stands on the edges? The wire guided VHS tubes also prevents torps from passing over or falling into them, which definitely doesn't make sense.

And? You'll soon learn the difference between KMC weapons and Karmic weapons.

I don't pander to bump loading lazy bums.

nsv13/code/modules/overmap/camera.dm
```diff
@@ -46,7 +46,8 @@
 	user.click_intercept = src
 	if(position != "observer")
 		user.add_verb(overmap_verbs) //Add the ship panel verbs
-	if(mass < MASS_MEDIUM)
+	
+	if(mass < MASS_MEDIUM || !istype(src, /obj/structure/overmap/torpedo))
```
DeltaFire15, line 50: should probably be like this if you want torpedoes to not return after that check. Otherwise, the check is basically always true.
```suggestion
	if(mass < MASS_MEDIUM && !istype(src, /obj/structure/overmap/torpedo))
```

Vasily2013:
Alright, just so it's documented. What is required to make this function on a given map?

Karmic-Skink:
> Alright, just so it's documented. What is required to make this function on a given map?

Pretty much moving forward, you'll likely see that every one of our player ships is going to have a unique apply_weapons() entry, as the auto generated one doesn't cover everything and results in a long list of empty entries to cycle through if we add everything.

Specific block for missle_cruiser attached:
image

Karmic-Skink:
> > Alright, just so it's documented. What is required to make this function on a given map?
> 
> Pretty much moving forward, you'll likely see that every one of our player ships is going to have a unique apply_weapons() entry, as the auto generated one doesn't cover everything and results in a long list of empty entries to cycle through if we add everything.
> 
> Specific block for missle_cruiser attached: image
> 
> And in this specific case, its line 180.

This will probably be moved to the base unit once I start testing AI using these tbh.

Vasily2013 (changes requested):
Fix the pipes that were moved, please.

image

Karmic-Skink:
> Fix the pipes that were moved, please.
> 
> image

Updated.

Vasily2013 (approved):
map changes ok.

red031000:
disposals seem to be broken on the atlas, bottom deck where escape leads into the chapel, after turning on the conveyor belt, all the stuff just gets jettisonned there

covertcorvid:
People are still having a really hard time getting torpedoes past the enemy PDC/flak defenses. We might consider increasing the AI delay for defensive weapons.
They're also still sometimes hitting the player ship.

red031000:
image
here's a view of the disposals issue in t-ray mode

red031000:
just checked, this is apparently in the normal atlas as well, apologies

covertcorvid:
I have received additional player feedback!
- Players would like IAS on
- Torpedoes take your full attention for too long to be good as a main weapon. They might be better if you could disconnect early or the torpedo was controlled by a different console. People actually are using PDC defensively and can't do so while torpedoing
- They don't feel like they are very effective on target
  - They are still shot down too easily. Enemy defenses are 360 and thus still hard to evade.
  - Torpedoes cost more per damage than NAC rounds
- People feel that two-seater fighters would be rad
- At least two people miss having to hand-build torpedoes
- Delta suggested a "laser-guided" system where torpedoes homed in on free-moving camera/dot that you could switch away from

Karmic-Skink:
End of support for this test.

May return later in the manner I want, and not per player polls.

PR #1886, opened 2022-04-05
Implements spatial partitioning for physics collision detection

TheFakeElon:
## About The Pull Request

Implements a dynamic quadtree data structure for efficient segmentation of physics objects, allowing us to ONLY enumerate over physics objects that are near us, greatly cutting down CPU usage when there are a large amount of physics objects over a wide area

## Why It's Good For The Game

runtime performance good

## Changelog
:cl:
code: makes collision detection significantly faster during heavy load
refactor: refactors physics collision detection to use spatial partitioning
/:cl:


TheFakeElon:
There's still plenty of room for further optimization but that can be worked on in a separate PR otherwise this'll never be finished

TheFakeElon:
With just a little tweaking, this could be slapped onto a lot of other spatial systems too, such as sounds and large range() enumerations

covertcorvid (approved):
Seems based, is not currently breaking anything

PR #1902, opened 2022-05-02
Fixes dradis overmap linkage

### LINKED ISSUE #1881: Consoles default to last dradis type console uses
Issue reported from  Round ID: 4067 (NSV 13 | Official | MRP | 98% Free Lag | Spaceship Combat Server)

Reporting client version: 514.1575

how to test: open the cargo freight delivery console, close the cargo console, go to the helm, open the helm, see the dradis is now using the cargo delivery consoles dradis rather then the normal dradis, cry when you cant ping

TheFakeElon:
fixes #1881

## About The Pull Request

Fixes the cargo freight delivery console from linking with the ship as the primary Dradis console

## Changelog
:cl:
fix: fixed freight consoles overriding primary dradis consoles as the ship's primary type
/:cl:


TheFakeElon:
I don't think this error is me, it looks like we're trying to use defines inside the map files for some reason?
image

Bokkiewokkie (approved):
46965721274_0ab81faf1b_o

PR #1905, opened 2022-05-04
Updating department_radio_keys localization

Archemagus:
## About The Pull Request

Replacing cp1251 radio key encoding with utf-8

## Why It's Good For The Game

You know after 513 and TGstation's localization update all cp1251 encoding can (and i guess should) be replaced with UTF-8
https://github.com/tgstation/tgstation/blob/master/code/modules/mob/living/living_say.dm#L32

Or let me guess... Upstream Issue?

## Changelog
:cl:
fix: Cyrillic in radio prefixes
/:cl:


Bokkiewokkie (approved):

covertcorvid (changes requested):
Please add NSV13 comment tags if you're not planning on doing this upstream

Archemagus:
A-a-and done. Thanks for waiting
https://github.com/BeeStation/BeeStation-Hornet/pull/6865/files

Archemagus:
Meanwhile still can't see reason why Exploration channel disabled.
Comment says "we need Q" but cyrillic key stands 'Ю' for ">"
May be put that channel back on ">" radio key
or
move ATC channel on other key? Like "A" or "R" or "F"?
What are you thinking about it?

Bokkiewokkie:
> Meanwhile still can't see reason why Exploration channel disabled. Comment says "we need Q" but cyrillic key stands 'Ю' for ">" May be put that channel back on ">" radio key or move ATC channel on other key? Like "A" or "R" or "F"? What are you thinking about it?

NSV does not use the exploration channel, so there's no point in moving our ATC channel from the Q key, which people have gotten used to using for it by now.

Bokkiewokkie (approved):

PR #1913, opened 2022-05-13
Renaming station engineers to engineers

Evie7056:
## About The Pull Request

This PR renames several (alomost all of there except where it doesn't have sense) instances of Engineer's being called "Station Engineer". No job path changed. No icon state changed. 

Technically this should not affect anything game-wise since everything that uses a job name as a pointer is renamed too. The most biggest concerns with such thing are (of course) upstream computability and ban preservation, since I doubt that every engineer ban would be saved, but this is not something that I can fix or reproduce to be 100% sure about. Keep this in mind. Don't recommend until decision about this is made.   

## Why It's Good For The Game

Can you see a station somewhere around here? Yeah, I too.
I guess you can call it "Engineer gaming"

## Changelog
:cl:
add: Station Engineer is killed
del: Long live the Engineer!
/:cl:

### DIFF
--- a/SQL/ban_conversion_2018-10-28.py
+++ b/SQL/ban_conversion_2018-10-28.py
@@ -62,7 +62,7 @@ def parse_role(bantype, job):
         role = "Server"
     else:
         #TG: Some legacy jobbans are missing the last character from their job string.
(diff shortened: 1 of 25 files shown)


Evie7056:
About ban concerns: probably the most easiest way to not break it is to run a simple quarry that replaces every Station Engineer ban with Engineer if bd is using name as a key like the dm.

Evie7056:
image

> 
> Need to update the rank config if we do this as well
I updated rank config files as well, even unused ones, I think I don't understand what you mean...

And somehow it decided to fail with runtimes, huh which is weird.

Karmic-Skink:
Save this for a mass change of jobs if we choose to do so.

PR #1858, opened 2022-03-17
moths can open their wings

red031000:
## About The Pull Request

Moths can now open their wings
partial port from yogstation13/Yogstation#11040

in addition, adjust the timings for flap and aflap to be half the time, as they did not seem fast enough

## Why It's Good For The Game

looks beautiful, adds flavour to moths

## Changelog
:cl:
add: Add the ability for moths to open their wings
tweak: Change time for flap and aflap to half
imageadd: Add moth wings open images from yogstation
/:cl:

PR #1922, opened 2022-05-20
Moth wings can now open (#6495) [Port from beestation]

red031000:
* moth wings can now open

* wing emote can only be used when flight level is flying or above

* change to istype rather than getting the organ again

* missed one

## About The Pull Request

Ports BeeStation/BeeStation-Hornet#6495
Which in turn ports #1858 
Which in turn ports yogstation13/Yogstation#11040

Basically a more robust and balanced version of #1858 which upstream has approved and merged

## Why It's Good For The Game

Adds flavour to moths, while also looking nice

## Changelog
:cl:
add: Add the ability for moths to open their wings when they are able to fly, port from upstream
tweak: Change time for flap and aflap to half
imageadd: Add moth wings open images from yogstation
/:cl:


Bokkiewokkie (approved):
Just an upstream port so good to merge

PR #1931, opened 2022-06-03
Refactors overmap keybindings

covertcorvid:
## About The Pull Request
Moves key handling for most things into keybinding datums. Adds keybindings for the first 4 selectable weapons.
It's a secret tool that will help us later.

Before merging:
- [x] AI no longer flies in straight lines all the time
- [ ] Eclipse and Atlas don't fly backwards when brakes are applied

## Why It's Good For The Game
Doing it this way helps us not have to worry about key conflicts as much, and the weapon thing is probably useful even though it may wind up different per ship

## Changelog
:cl:
add: Top row numbers 1-4 can be used to select weapons while gunning
refactor: Moved overmap key handling into the keybind system
/:cl:

### DIFF
--- a/code/datums/keybinding/_defines.dm
+++ b/code/datums/keybinding/_defines.dm
@@ -4,6 +4,7 @@
 #define CATEGORY_HUMAN "HUMAN"
 #define CATEGORY_MISC "MISC"
 #define CATEGORY_ROBOT "ROBOT"
+#define CATEGORY_OVERMAP "OVERMAP" //NSV13 - overmap controls
 
 #define WEIGHT_HIGHEST 0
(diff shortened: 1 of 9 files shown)


covertcorvid:
Re: #1950 Oddly enough the AI were still able to turn in my local testing, but looking through the changes I think I see what could've happened and I've added a possible fix

Bokkiewokkie:
> Re: #1950 Oddly enough the AI were still able to turn in my local testing, but looking through the changes I think I see what could've happened and I've added a possible fix

Can hopefully TM and merge if everything works fine then

Bokkiewokkie (approved):

PR #1933, opened 2022-06-03
Ghost ship logging and minor fix

covertcorvid:
## About The Pull Request
This makes game.log tell us when ghost ships happen, as well as a couple possible reasons they might not. Also added an early return if the RNG doesn't give us a ghost ship since that was sometimes causing a runtime.

## Why It's Good For The Game
We can tell if this is actually ever happening

## Changelog
:cl:
admin: Added ghost ship spawning info to game.log
fix: No longer tries to pick from an empty list when RNG decides not to spawn a ghost ship
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/ai-skynet.dm
+++ b/nsv13/code/modules/overmap/ai-skynet.dm
@@ -700,8 +700,13 @@ Adding tasks is easy! Just define a datum for it.
 					message_admins("Failed to spawn ghost ship due to insufficent players.")
 					return
 
+			if(!length(ship_list))
+				message_admins("Failed to spawn ghost ship due to lack of valid ship types")
+				return
 			var/target_location = locate(rand(round(world.maxx/2) + 10, world.maxx - 39), rand(40, world.maxy - 39), OM.z)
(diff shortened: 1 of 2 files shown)


nsv13/code/modules/overmap/ai-skynet.dm
```diff
@@ -685,24 +685,30 @@ Adding tasks is easy! Just define a datum for it.
 			if(SSovermap_mode.override_ghost_ships)
 				message_admins("Failed to spawn ghost ship due to admin override.")
 				return
+			if(!prob(10))
+				return
+
```
Bokkiewokkie, line 690: image

Bokkiewokkie (dismissed):

Karmic-Skink:
This probably needs a check fix for if its chosen to spawn something like a fighter and the fleet doesn't have any fighters.

1 default ship per faction?

covertcorvid:
> This probably needs a check fix for if its chosen to spawn something like a fighter and the fleet doesn't have any fighters.
> 
> 1 default ship per faction?

yeah fair, either a default ship type or just if(!length(fighter_types)) ship_list += battleship_types

PR #1951, opened 2022-06-14
make gauss guns require no power

### LINKED ISSUE #1855: You cannot repair Gauss Turrets and use them afterward
## Reproduction:
Destroy gauss by any means necessary. Disassembling is also fine, I guess.
Rebuild the gauss turret on the upper deck.
Rebuild the gauss chair wherever you want on the lower deck. Preferably underneath the turret, but you'll see why it doesn't matter in a bit.
Buckle yourself to the chair. If you're lucky, you'll go up into the turret.
Once in the turret, click frantically at the control panel and realize you can't do anything.
Descend from the gauss turret. 
Repeat at necessary to further cement that you cannot rebuild the gauss turret and be able to use it afterward.

I've done some testing on this with someone else on the test server. We've tried wacking them with multitools, we've tried moving the construction of the gauss turret and chair, we've tried building the chair first THEN the turret, all to no avail. 

Also, the sprite for the gauss turret doesn't line up with anything once rebuilt. Below is a picture of a rebuild turret. The red is where the machine frame was assembled, the yellow is where the hitbox actually is, and you can see the sprite for yourself 

image

Kenionatus:
## About The Pull Request

Gauss guns currently require power in the turret's room to operate. Prebuilt turrets seem to be powered by default but rebuilt turrets are usually unpowered because they're in an area without APC. Making them require no power is an easier fix than making new powered areas in space for each gauss battery.

## Why It's Good For The Game

bugfix without much side effects (Fixes #1855)

## Changelog
:cl:
fix: rebuilt gauss guns should now always be usable
/:cl:

### DIFF
--- a/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/gauss_gun.dm
+++ b/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/gauss_gun.dm
@@ -31,6 +31,8 @@
 	var/BeingLoaded //Used for gunner load
 	var/list/gauss_verbs = list(.verb/show_computer, .verb/show_view, .verb/swap_firemode)
 	circuit = /obj/item/circuitboard/machine/gauss_turret
+	//INTERACT_MACHINE_OFFLINE added because gauss guns are usually unpowered because they're usually in near space
+	interaction_flags_machine = INTERACT_MACHINE_OFFLINE | INTERACT_MACHINE_WIRES_IF_OPEN | INTERACT_MACHINE_ALLOW_SILICON | INTERACT_MACHINE_OPEN_SILICON | INTERACT_MACHINE_SET_MACHINE
 
 /obj/machinery/ship_weapon/gauss_gun/MouseDrop_T(obj/structure/A, mob/user)
 	. = ..()


Karmic-Skink:
Fix it properly.

Kenionatus:
@Karmic-Skink 
> Fix it properly. -1

I'm not much of a mapper. If you go through all of the maps and add APC powered areas, I'll see if I can change mapper placed Gauss Guns to no have power by default so unpowered ones get caught early.

Karmic-Skink:
> @Karmic-Skink
> 
> > Fix it properly. -1
> 
> I'm not much of a mapper. If you go through all of the maps and add APC powered areas, I'll see if I can change mapper placed Gauss Guns to no have power by default so unpowered ones get caught early.
> 
> Addendum: NAC turrets were previously changed to not require power, so this PR would bring them in line with each other.

NACs are chemically propelled.

Draw power from the chair location.

Bokkiewokkie (changes requested):
Changing the maps to have the gauss bay areas cover the turrets on the other deck would be a better solution.

Karmic-Skink:
> Changing the maps to have the gauss bay areas cover the turrets on the other deck would be a better solution.

That sounds like a dodgy work around.

Kenionatus:
I think the best solution would be to wire a cable to every gauss gun and make them draw a significant (but not too large not not majorly mess up energy balancing) amount from it.

Bokkiewokkie:
> I think the best solution would be to wire a cable to every gauss gun and make them draw a significant (but not too large not not majorly mess up energy balancing) amount from it. I just don't really feel like mapping.
> 
> Taking power from below would be quite the overhaul, since the "actual" gauss gun is the turret, which is classified as a machine. Loading rack is an optional part automatically created below if possible and gunner chair needs to be constructed manually and is also optional.

Can you please (at least just by coding, the mapping can be done by someone else if you really don't want to) add that instead then? We're not gonna merge this PR in its current state

Kenionatus:
Ok. Put this on the top of my mental NSV todo list.

PR #1959, opened 2022-06-19
[READY] Atlas Maintenance 

Vasily2013:
## About The Pull Request

It is again time for me to emerge from hibernation to help update some maps. Today we tackle the atlas, starting with matching the turfs on the upper and lower z levels so you can't go up/down into the ship through what should be solid hull.

## Why It's Good For The Game

Glitches are bitches and we should get rid of them. We can do this by mapping!

## Changelog
:cl:
tweak: Matches Atlas upper and lower deck turfs to prevent going up/down through what should be hull plating.
add: Added gas pump to Atlas reactor feed line.
fix: Replaced protolathe in Atlas sec with techfab.
add: Holopad in Atlas brig control.
tweak: Re-jigger Atlas waste cooler to be more clearly piped.
add: Borg chargers in Atlas maint south and north of the PKA room.
tweak: Atlas: Moves one (1) hull repair juice tank to the hull armor on the bow.
tweak: Moved one crate in Atlas sci maint to block an abandoned door. (This will be the only OSHA compliance fix of this PR)
/:cl:


Koekiegast:
I am once again asking for a gas pump on the fuel line of the reactor in the reactor room.

someone543:
Brig has a protolathe instead of a techfab.

SerynEngi:
Request you fix the following cursed piping setups:
1. The cooler inside atmospherics sitting on top of one of the pipes its connected to, There is the space to reorient that, and the layer adapter is hidden underneath the disposals bin outside of mapping view, as is the connection point for the cooler itself.

2. The whole area around the Z-Level adapters for atmospherics technically violates mapping standards, given that it passes through walls, and it does not need so many layer adapters in the area. Z-Level adapters are capable of being attached to different layers, even if the sprite would then need to be pixel adjusted.

3. Airlocks only need one layer adapter to function if you pipe to them properly. Request you fix this for external airlocks, as it has forced me into bizarre alternate setups to emergency vent rooms that have been hit with hellfire torps on the upper deck.

In addition:
1. The whole point of having a pump next to the stormdrive input is so that you can adjust the stormdrive input quickly if running an experimental mix, so I agree with koekiegast here.

2. I doubt you're going to change it, but once again, it would be nice if this wasn't the only map (other than the Snake, which has its own cursed atmospherics setup, and the Shrike, which has it arranged the way it does for an actual reason) that doesn't have the atmos tank and distribution was arranged in the standard 02/N2/CO2/N2O/Plasma layout that every other ship that utilizes a stormdrive favors.


Minor requests:
1. Request a borg charger somewhere in engineering or the aft end of the ship so I don't have to run halfway across the ship if I want to charge at a decent rate. It is very odd to me that a section of ship likely to have borgs, IPCs or (admittedly rare) ethereals doesn't have one.

2. Request that the hull-foam tanks be relocated somewhere other than engineering next to the reactor. Maybe in the rather large amounts of available space in munitions or in the area that the reactor mix tank is? We don't use them as engineers very often (if at all) and they mostly just get in the way.

red031000:
maintenance crates in sci maint block access, while it's not a *massive* issue, it's still annoying for radstorms and access etc

Vasily2013:
- [ ] Added gas pump to reactor feed line.
- [ ] Replaced protolathe in sec with techfab.
- [ ] Holopad in brig control.
- [ ] Re-jigger waste cooler to be more clearly piped.
- [ ] Z level wallpipes are not a concern as they are there to eliminate a debug check from triggering.
- [ ] Upper deck airlocks already only use one layer manifold. 
- [ ] Borg chargers added in maint south and north of the PKA room.
- [ ] Moves one (1) hull repair juice tank to the hull armor on the bow.
- [ ] Moved one crate in sci maint to block an abandoned door. (This will be the only OSHA compliance fix of this PR)

Bokkiewokkie (approved):
image

Vasily2013:
Moff

sunofang:
I just remembered there is intellicard or secret documents cabinet, having the secret documents in the safe is stupid.

PR #1966, opened 2022-06-24
Beg mappers to talk to maintainers

Kenionatus:
## About The Pull Request

There was a bunch of talk about "mapping standards" in the Discords but I wasn't able to find them yet. Turns out those are WIP documents circulated in private channels but are already being applied to new maps. This is understandable, imo, but I think potential mappers should be warned of it and directed to talk to maintainers to make sure their ideas are considered mergable. Already having a section titled "Mapping Standards" makes it seem like what is listed there is everything a map needs to conform to.
My way of formulating it probably isn't the best, but I'm strongly of the opinion *something* should be in CONTRIBUTING.md.

## Why It's Good For The Game

Helps to avoid a lot of wasted time and drama.

## Changelog

No in game changelog needed imo.


Pontenerd:
If this happens, it would probably also be a good idea to pin it on the discord somewhere

covertcorvid (approved):

PR #1975, opened 2022-06-30
Tycoon Medbay redo

Absolutely-Steph:
## About The Pull Request
old medbay was fine, I brought this concept up to the map maintainers and they said they're not against the idea, so i'm trying it, if only to learn how the heck disposals work.
medbay redone to feature more avenues for RP while still being effective.

## Why It's Good For The Game
RP good, this medbay allows for more RP without slowing things down too much, also muh AESTHETIC.

## Changelog
:cl:
tweak: redid the tycoon medbay
/:cl:


bluey89962:
some pictures would be handy for this PR I think

Absolutely-Steph:
> some pictures would be handy for this PR I think

You can see the full changes with the mapdiffbot

Though I have no real hope for this PR since bokkie doesn't like the layout.

Karmic-Skink:
> > some pictures would be handy for this PR I think
> 
> You can see the full changes with the mapdiffbot
> 
> Though I have no real hope for this PR since bokkie doesn't like the layout.

Not everyone has access to see the mapdiffbot.

It sometimes doesn't do a full render, and it definitely frequently doesn't result in a true to game capture of the game space.
The image request is reasonable.

Absolutely-Steph:
> > > some pictures would be handy for this PR I think
> > 
> > 
> > You can see the full changes with the mapdiffbot
> > Though I have no real hope for this PR since bokkie doesn't like the layout.
> 
> Not everyone has access to see the mapdiffbot.
> 
> It sometimes doesn't do a full render, and it definitely frequently doesn't result in a true to game capture of the game space. The image request is reasonable.

Aight never heard of that but it sounds reasonable, im gonna drop some in game pics:
As an aside, I'm growing fond of turning the main hallway into a 3 wide one, i think that'd look pretty dope.
General treatment ward including patient rooms, aux surgery, sleeper room/lobby/entrance area:
image
Emergency Ward including ER 1 & 2/Cryocloning/surgery/Emergency medicine cabinet:
image
Staff area featuring the lounge, chemistry, genetics, secpoint, cmo officer, and storage:
image
Quarantine ward featuring virology lab, monkey pen, and the morgue:
image

Absolutely-Steph:
> Aesthetic mismatch. Tycoon doesn't use ShrikeTM floor lines.

This is meant to be an NT aftermarket refurbishment, which is why there are no sg banners anywhere. The disconnect in tone is intentional. If you want them gone I can do that tho.

Karmic-Skink:
> > Aesthetic mismatch. Tycoon doesn't use ShrikeTM floor lines.
> 
> This is meant to be an NT aftermarket refurbishment, which is why there are no sg banners anywhere. The disconnect in tone is intentional. If you want them gone I can do that tho.

Yes, that was indeed a joke.
However, please keep the medbay atheistic consistent with the rest of the map.

ta3370:
CMO Office doesn't overlook anything, it may need moved to the room on the right so it's more central and a CMO could actually see and interact from things while in their office.
There's too many supplies all over the place. Your "emergency" supply closet is stocked like it's a primary darned storage instead of "oh shit I need to grab whatevers availible" or a highly limited supply (talking like individual items outside of medkits, so like some loose trauma kits and such, without the vendor).
Entry Surgery needs removed, otherwise it will be the only one used.
Too many sleepers, if used they should be that one specialized room.
Too many stasis beds, many should be replaced with normal beds. Entry Surgery could instead become a triage room, perhaps being the place for stasis beds (which I generally dislike in general, especially their "haha invincible surgery on perfect table isn't surgery table", but I understand their use besides wishing their use for surgery was removed entirely or greatly hampered, but that's for other code work).
Virology quarantine needs more windows. Any time in which that gets actually used, you have players stuck there for a while who don't need to be treated worse than brig prisoners in isolation unable to talk to eachother between cells. Preferrably viro workstations would have the virologist sat in range to interact with them aswell. I'm surprised that wasn't considered in a rework labeled as "improving RP opportunities".
Medbay needs either a Bathroom with Shower, or a dedicated Decon room. Either or, for treating Engineers.

This one is just a side statement for a potential sidegrade separate from the critiques: The room to the right of the "emergency" supply closet could be rejigged into something like a Protective ward/Custody ward for "patients under threat/prisoners undergoing treatment", and/or as a dialysis/IV/Rad-treatment room. Could even be a blood-drive room with empty bags, IV's, and chairs.
Could just as much move Genetics there in a compact-office so they're across from Cloning. Personally I'm partial to it being either the Custody/Protective Ward with it getting the door named, or the Genetics Office.

Absolutely-Steph:
God so many points, i'll address them all in sequence.
**First some points to put us on the same page.**
From your surgery room comment I believe you think that the General treatment area will be the one primarily used, I'm not convinced of that at all, worst case our doctors will never set even a single foot in that entire wing, especially if the alternative floorplan is implemented instead. The "ideal" strat with this medbay is probably to move all supplies into the ERs and bolt open the doors so you can use it like any ordinary medbay.

If we went by the mapping standards, assuming the tycoon is planned to handle ~20 players, we would only have a single stasis bed/sleeper combo, by the way. Most maps are in violation of said standards.

**CMO office:**
I'm not sure the office necessarily has to overlook anything, also because this medbay was designed in part to have hidden nooks and crannies, i'd argue the CMO would be in that office only when sleeping/chilling/changing access, nothing of which needs a central 360° view. I'd rather keep it like this, but If you really want this though I'd have to go through with the alternative floorplan detailed at the bottom of this or with the plan offered in the TLDR/Recap.

**Supplies/Sleepers/Stasis**
Yeah the supplies situation is a bit iffy I feel, **I'd be well open to turning the emergency supplies to a small cab that maybe has some useful stuff to quickly stabilise maybe 5-10 people.**

As to the sleepers, the original tycoon medbay has 2 sleepers next to one another in the main treatment room, this one has 2 spread over the emergency rooms(that would take the place of the original ones) plus an additional one in the general treatment ward. I'd be open to removing the sleeper from GT but I fear it would make people use GT less, this is also my concern with your proposed removal of the GT surgery area.

Stasis beds I got no clue how to fix(if it even needs fixing), we got 3 stasis beds in the original medbay basically right next to each other(well no, 2 of them are cramped in some weird side corridor), whereas my medbay has **8!!** spread over a wide area, which might just balance it out maybe? Frankly I feel that stasis beds are required for the GT rooms to keep their usefulness beyond just RP, but they are also very needed in the emergency rooms due to life or death scenarios... 

**Quarantine ward:**
My idea here was to have the virologist administer medicines to the patients in their rooms, I was also extremely starved for space here, as you can probably tell. A seperate desk on the inside of the quarantine ward sounds very interesting and will definitely happen. The doors in here do not have access restrictions, so the patients can freely walk around the lobby area or retire to their rooms. I even left some fancy paper/ink pen in there in case they need it to write any last wills. More windows are very much possible and were actually planned at first, but I forgot to bring them over when I moved the virology wing back down from the upper deck.

*Genetics/Custody/Protective Ward/Bathrooms*
First off, fuck yeah bathrooms, totally forgot about them, will add.
Genetics should probably stay in the staff area, I placed chem and gene in there specifically to curbstomb them handing out stuff to tiders.
The protective ward idea is genius and I will definitely add that somehow or somewhere, there will probably be space once I adjust all the stuff mentioned here.

### To recap/TLDR
**Changes that will happen:**
-Quarantine windows and interior virology desk
-less supplies and smaller cabbie
-main hallway made 3 tiles wide by shrinking cryo and ER2 by one tile horizontally
-ER2 will be remade into a Protective Ward
-bathrooms will be added somewhere

**Changes that I am unsure about, see the points I raised above:**
-Messing with GT surgery
-Messing with the CMO office(thinking about it, I could swap it with ER2 and move the protective ward into the space left.)
-Reducing the amount of stasis beds(beyond removing 2 cuz of ER2 of course), that'd bring us to 3 more stasis beds than original, and 1 less sleeper than original.


**Frankly I think we should stop worrying so much about balance, as long as it's fun we're doing what we're supposed to do, which is make a fun game. I think we should try this even with it's 6 stasis beds, see what the crew even does with it. Actually, I know what the crew will do with them, as the aetherwhisp has 3 stasis beds/3 sleepers, which is also much more than it's crew would normally need as per mapping standards. And they're doing just fine.**


**Regarding the "alternative floorplan"**
Main idea is to make it more "traditional" by bringing it in line with all other (bad)medbays that just cram all tools into a single room for maximum efficiency.
This would involve extending ER1 all the way down and merging it with the supplies cabbie and sleeper room, turning it into a traditional-style big room with all tools needed.
**This will probably result in this being the only room ever used.**
The new ER will have 1-2 sleeper(s) and 2 stasis beds. The sleeper(s) will be in a seperate section of the room which is the current sleeper room of the GT, seperated by thin glass.
all in all this would be familiar, boring, and defeat the entire point of this remapping, beyond maybe the gained style, though i would rather have this than it not be accepted at all.

Absolutely-Steph:
Changes in the newest version:
-Removed ER2(and all the supplies in it), replaced with protective custody/dialysis room
-Removed Supply cabby, replaced with decon shower booth
-Moved GT sleeper into GT surgery, replaced by restroom
-Added windows and an interior booth to quarantine, replaced room 3 with a restroom
-Widened the central hallway
image
image

Net changes on balance:
3 less stasis/sleeper
one less of each dedicated medkit
more bloodbags/straitjackets

Absolutely-Steph:
Since I finished the atmos and disposals, this is now functional. As such I am moving it out of draft.

Absolutely-Steph:
denied by bokkie

PR #1976, opened 2022-06-30
[WIP] Boarder Nerf

Kenionatus:
## About The Pull Request

Limits NPC boarder fire and attack rate by adding a cooldown between each shot/attack. Number changes the same way as their move speed, changing from 0.4 on low pop to 0.2 on high pop. Number up to discussion. Also refactors a bunch of comments to be doc comments.
Limits boarder firing range to the player's sight range. They can still see the player much further out, but will hold fire until the player can see them.

## Why It's Good For The Game

I (and I believe other people too) consider boarders in their current state to be quite OP. They can see and engage the player from beyond the player's sight range and their damage output is way beyond what a player could achieve with the same gear. I believe they also out dps players with assault rifles.

Nerfing them prevents cases of frustration where players (imo rightfully) feel that they had no fighting chance when they got killed by a machineguning boarder from beyond their vision.

## Changelog
:cl:
balance: Boarders can no longer fire from out of sight (but can still see farther than players).
balance: Boarders have a limited fire rate, based on the number of players.
refactor: I like doc comments
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/knpc.dm
+++ b/nsv13/code/modules/overmap/knpc.dm
@@ -458,6 +458,12 @@ This is to account for sec Ju-Jitsuing boarding commandos.
 	var/dist = get_dist(H, target)
 	//We're holding a gun. See if we can shoot it....
 	HA.pathfind_to(target) //Walk up close and YEET SLAM
+	var/default_view = getviewsize(CONFIG_GET(string/default_view))
+	var/viewdist = list(round(default_view[1]/2), round(default_view[2]/2))
+	var/is_in_x_view = ISINRANGE(target.x, H.x - viewdist[1], H.x + viewdist[1])
+	var/is_in_y_view = ISINRANGE(target.y, H.y - viewdist[2], H.y + viewdist[2])
+	if (!(is_in_x_view && is_in_y_view))
+		return
 	var/obj/item/gun/G = null
 	if(istype(A, /obj/item/gun))
 		G = A
--- a/nsv13/code/modules/overmap/knpc.dm
+++ b/nsv13/code/modules/overmap/knpc.dm
@@ -34,7 +34,7 @@ GLOBAL_LIST_EMPTY(knpcs)
 	///How long we delay between actions
 	var/action_delay = 6
 	///Delay between attacks
-	var/attack_delay = 4
+	var/attack_delay = 2
 	///world time when delay for attacking will be over
 	var/next_attack = 0
 	var/knpc_traits = KNPC_IS_DODGER | KNPC_IS_MERCIFUL | KNPC_IS_AREA_SPECIFIC
@@ -58,9 +58,11 @@ GLOBAL_LIST_EMPTY(knpcs)
 			if(1) //Lowest pop, so slow them down
 				move_delay += 2
 				action_delay += 2
+				attack_delay += 2
 			if(2 to 3)
 				move_delay += 1
 				action_delay += 1
+				attack_delay += 1
 
 /datum/component/knpc/Initialize()
 	if(!iscarbon(parent))
@@ -400,9 +402,6 @@ This is to account for sec Ju-Jitsuing boarding commandos.
 
 /datum/ai_goal/human/proc/reload(datum/component/knpc/HA, obj/item/gun)
 	var/mob/living/carbon/human/ai_boarder/H = HA.parent
-	if(world.time < H.next_attack)
-		return FALSE
-	H.next_attack = world.time + H.attack_delay
 	if(istype(gun, /obj/item/gun/energy))
 		var/obj/item/gun/energy/E = gun
 		if(E.selfcharge) //Okay good, it self charges we can just wait.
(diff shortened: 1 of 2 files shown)


Karmic-Skink:
Working as intended?

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -1,30 +1,44 @@
-#define AI_TRAIT_BRAWLER 1 //The "marines". These guys will attempt to stand and fight
-#define AI_TRAIT_SUPPORT 2 //The "medics", civvies, etc. These lads will call for backup
+///The "marines". These guys will attempt to stand and fight
+#define AI_TRAIT_BRAWLER 1
```
covertcorvid, line 2: the vscode extensions don't link defines to anything, so no need to worry about fixing their comments

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -575,10 +595,14 @@ This is to account for sec Ju-Jitsuing boarding commandos.
...
 	var/id = null
-	var/next_id = null //id of the node that this one goes to. Alternatively, a list of ids which will all be possible next destinations.
-	var/previous_id = null //id of the node that precedes this one
-	var/obj/effect/landmark/patrol_node/previous //-- This isn't actually used anywhere.. - Delta
-	var/list/next_nodes	= list() //List of possible followup nodes set by next_id. If multiple entities exist in the list, one will be chosen at random on every occasion.
+	///id of the node that this one goes to. Alternatively, a list of ids which will all be possible next destinations.
+	var/next_id = null
+	///id of the node that precedes this one
+	var/previous_id = null
+	///-- This isn't actually used anywhere.. - Delta
```
covertcorvid, line 602: Can we... get rid of it then?

covertcorvid (commented):
I thought doc comments were supposed to work in-line but oh well.
Fix your changelog though.

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -383,6 +400,9 @@ This is to account for sec Ju-Jitsuing boarding commandos.
 
 /datum/ai_goal/human/proc/reload(datum/component/knpc/HA, obj/item/gun)
 	var/mob/living/carbon/human/ai_boarder/H = HA.parent
+	if(world.time < H.next_attack)
```
Karmic-Skink, line 403: This code is far too static and does not consider the dynamic scaling code of KNPC spawns.

Karmic-Skink (changes requested):
Also as a blanket rule, is far too restrictive particularly on things like funni admeme icebots.

nsv13/code/modules/overmap/knpc.dm
Kenionatus: Thanks for the feedback. Changed it to be handled the same way that movement speed is in the current version.

nsv13/code/modules/overmap/knpc.dm
Kenionatus: Not my code ¯\\_(ツ)_/¯

Kenionatus:
Thinking about it, if the attack range part gets merged we could increase their view range even more, making them seem a bit more clever. Downside is that the cost of looking for enemies in oview would grow in the second power.

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -1,30 +1,44 @@
...
 GLOBAL_LIST_EMPTY(knpcs)
 
 /datum/component/knpc
 	var/ai_trait = AI_AGGRESSIVE
 	var/static/list/ai_goals = null
 	var/datum/ai_goal/human/current_goal = null
-	var/view_range = 12 //How good is this mob's "eyes"?
-	var/next_backup_call = 0 //Delay for calling for backup to avoid spam.
+	///How good is this mob's "eyes"?
+	var/view_range = 12
```
Karmic-Skink, line 13: Tag 1

Karmic-Skink, line 13: Link this to your newly used variables.

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -436,6 +458,12 @@ This is to account for sec Ju-Jitsuing boarding commandos.
 	var/dist = get_dist(H, target)
 	//We're holding a gun. See if we can shoot it....
 	HA.pathfind_to(target) //Walk up close and YEET SLAM
+	var/default_view = getviewsize(CONFIG_GET(string/default_view))
+	var/viewdist = list(round(default_view[1]/2), round(default_view[2]/2))
+	var/is_in_x_view = ISINRANGE(target.x, H.x - viewdist[1], H.x + viewdist[1])
+	var/is_in_y_view = ISINRANGE(target.y, H.y - viewdist[2], H.y + viewdist[2])
+	if (!(is_in_x_view && is_in_y_view))
```
Karmic-Skink, line 465: Shallow binary implementation.

Karmic-Skink (changes requested):
This is a poor solution to KNPCs effectively 'sniping' players off screen, any solution needs to be more dynamic, scaling and modular.

Please consider that players can and will make effective attacks on KNPCs from outside their own view range on the target and can do so quite effectively, and as such, KNPCs should also be able to.

There are many possible avenues to approach on this including, but not limited to; accuracy penalties, target delays, the Bioshock effect.
Again, include the scaling effects of the difficulty master controller.

In addition to all of this, I require you to have a bypassing flag that can be set on the KNPC for things that really should be able to snipe off screen.

nsv13/code/modules/overmap/knpc.dm
TheFakeElon: Shallow binary? What's that

TheFakeElon: Ah, I think Karmic was pointing out that you're using a NAND evaluation, which doesn't look right here.
NAND inverts the output of AND, not the inputs.
image
As you can see, if x and y are in view the condition will be true and the proc will return, which doesn't appear to be the desired behavior

Kenionatus:
Closed in favour of #1999

PR #1980, opened 2022-07-03
Escape pod docking fix

covertcorvid:
## About The Pull Request
A better fix for an issue seen earlier where an escape pod was unable to dock with the main ship

## Why It's Good For The Game
Let them in!

## Changelog
:cl:
fix: Made sure escape pods have a system
fix: Did a proper fallback for ships without a system trying to dock
/:cl:


Bokkiewokkie (dismissed):

Bokkiewokkie:
Syndicate sabre escape pod couldn't be picked up by a NT sabre or dock at the NT main ship, is that intentional?

covertcorvid:
> Syndicate sabre escape pod couldn't be picked up by a NT sabre or dock at the NT main ship yesterday, is that intentional?

Definitely not intentional. There are no obvious faction checks in the docking_act proc. I'm unsure whether any sabre can pick up any escape pod at the moment, but it should've been able to dock with the main ship just fine...

covertcorvid:
I varedited an escape pod's faction to syndicate and was still able to dock, so I think it's probably fine?

PR #1981, opened 2022-07-03
Ammo sorter features

Ikalpo:
## About The Pull Request
* Changes the way you link sorters to a console, now you select the console first and just hit all the sorters you want to connect.
* Shrinks the examine text on full sorters by an order of magnitude, and now it shows the amount it can hold.
  * also missile assemblers
* Changes some icons on the sorter console UI
* Adds a collapsing options button to the sorter console UI
* Adds buttons to the sorter console UI to change the order of them in the list
* The multitool no longer bonks any NSV13 machinery that actually has a multitool_act()
* Sorters and assemblers can be disassembled
* NSV13 machinery that can be unanchored can now also be assembled in an unanchored state
* Sorter consoles UI now updates when sorters die, and when they are loaded
## Why It's Good For The Game
nice to haves are nice to have
also you just should not be able to bash a computer to death (eventually) by using a multitool on it.
## Changelog
:cl:
add: You can change the order of ammo racks in the console
add: The options for renaming/removing ammo racks are hidden by default
tweak: The examine text for ammo racks groups items of the same type together
tweak: Ammo rack console linking is a bit easier
fix: Teaches spacemen that they don't need to hit things so hard with their multitools to make them work
/:cl:

<details>

image
image
image
image

</details>

### DIFF
--- a/nsv13/code/game/machinery/computer/helm.dm
+++ b/nsv13/code/game/machinery/computer/helm.dm
@@ -30,9 +30,10 @@
 		playsound(src, 'nsv13/sound/effects/computer/startup.ogg', 75, 1)
 		to_chat(user, "<span class='warning'>Autopilot [linked.ai_controlled ? "Enabled" : "Disengaged"].</span>")
 	..()
+	return TRUE
 
-// Helm and tactical in one console, useful for debugging 
-// This console should not be made available for the player ship, looking at you mappers 
+// Helm and tactical in one console, useful for debugging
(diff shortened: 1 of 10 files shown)


Bokkiewokkie (dismissed):
🙏 thank you for contributing to the tech debt repayment fund

Bokkiewokkie:
During PVP testing the ammo racks of the shrike (id = null) were linked to the Hammurabi's ammo rack console (id= hammurabi1)

covertcorvid (approved):

PR #1982, opened 2022-07-05
Knpc pathfinding pass (and a bit of normal pathfinding)

DeltaFire15:
## About The Pull Request
At the moment, knpcs have some issues with their pathfinding and some lag generation. Additionally, normal non-knpc mobs are unable to use firelocks, even if powered.
This PR aims to improve both of these somewhat, namely, bots should be able to pass firelocks (provided they are powered), while knpcs can just generally pass any firelock that isn't sealed.
Knpcs additionally should have some minor improvements towards them, namely actually accounting for tables as things they can pass aswell as not resetting their patrol node when picking up weaponry (since, they might drop theirs quite sometimes when patrolling which would cause them to just return to their old node if such a table is close)
Additionally, knpcs now gain an increasing level of "timeout" if they find themselves unable to pathfind at all, which should serve to prevent them severely hampering processing (pathfinding through the full tree ten times a second is uh, not the best processing wise). This might lead to them behaving slightly dumber in some few cases, but should work better in general.
They also shouldn't ask themselves for assistance anymore, which led to a runtime.
Additionally fixes a few random runtimes that showed up while debugging.

Someone may want to look this over or do their own investigating because this is mild suffering:tm:

## Why It's Good For The Game
uhhhhhhhhhh

## Changelog
:cl:
code: Improved some knpc pathfinding stuff.. Probably.
tweak: Things that pathfind can now pass firelocks.
fix: Knpcs no longer call themselves for backup.
fix: Fixed some random runtimes.
/:cl:

### DIFF
--- a/code/__DEFINES/flags.dm
+++ b/code/__DEFINES/flags.dm
@@ -168,6 +168,16 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define KNPC_IS_DOOR_BASHER 1<<4		//Does the KNPC kick the door off its hinges if it doesn't have a valid ID?
 #define KNPC_IS_DOOR_HACKER 1<<5		//Does the KNPC hack the door open if it doesn't have a valid ID?
 
+//Knpc pathfinding return values
+#define KNPC_PATHFIND_SUCCESS 0 //Path successfully generated
+#define KNPC_PATHFIND_SKIP 1 //Regenerating of path isn't needed, incapable of moving, etc
+#define KNPC_PATHFIND_TIMEOUT 2 //Pathfinding is currently in timeout due to having failed previously.
+#define KNPC_PATHFIND_FAIL 3 //No path found
+
+//Knpc pathfinding timeout defines
+#define KNPC_TIMEOUT_BASE (3 SECONDS) //The base timeout applied if an knpc's pathfinding fails.
+#define KNPC_TIMEOUT_STACK_CAP 9 //Every consecutive pathfinding fail adds a stacks; Timeout applied uses them as multiplier up to a cap of this value.
+
 //Starsystem Traits
 #define STARSYSTEM_NO_ANOMALIES 1<<0	//Prevents Anomalies Spawning
 #define STARSYSTEM_NO_ASTEROIDS 1<<1	//Prevents Asteroids Spawning
--- a/code/__DEFINES/flags.dm
+++ b/code/__DEFINES/flags.dm
@@ -168,16 +168,6 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
 #define KNPC_IS_DOOR_BASHER 1<<4		//Does the KNPC kick the door off its hinges if it doesn't have a valid ID?
 #define KNPC_IS_DOOR_HACKER 1<<5		//Does the KNPC hack the door open if it doesn't have a valid ID?
 
-//Knpc pathfinding return values
-#define KNPC_PATHFIND_SUCCESS 0 //Path successfully generated
-#define KNPC_PATHFIND_SKIP 1 //Regenerating of path isn't needed, incapable of moving, etc
-#define KNPC_PATHFIND_TIMEOUT 2 //Pathfinding is currently in timeout due to having failed previously.
-#define KNPC_PATHFIND_FAIL 3 //No path found
-
-//Knpc pathfinding timeout defines
-#define KNPC_TIMEOUT_BASE (3 SECONDS) //The base timeout applied if an knpc's pathfinding fails.
-#define KNPC_TIMEOUT_STACK_CAP 9 //Every consecutive pathfinding fail adds a stacks; Timeout applied uses them as multiplier up to a cap of this value.
-
 //Starsystem Traits
 #define STARSYSTEM_NO_ANOMALIES 1<<0	//Prevents Anomalies Spawning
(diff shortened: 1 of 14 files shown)


code/game/objects/structures/railings.dm
```diff
@@ -127,3 +127,15 @@
 	air_update_turf(1)
 	ini_dir = dir
 	add_fingerprint(user)
+
+//NSV13 - railings don't completely block the turf for A* purposes
+/obj/structure/railing/CanAStarPass(obj/item/card/id/ID, to_dir, atom/movable/caller)
```
Karmic-Skink, line 132: Sorcery.

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -131,13 +139,41 @@ GLOBAL_LIST_EMPTY(knpcs)
...
+		var/turf/next_turf = get_step_towards(H, path[1])
+		var/turf/this_turf = get_turf(H)	
 		//Walk when you see a wet floor
-		if(T.GetComponent(/datum/component/wet_floor))
+		if(next_turf.GetComponent(/datum/component/wet_floor))
 			H.m_intent = MOVE_INTENT_WALK
 		else
 			H.m_intent = MOVE_INTENT_RUN
 
+		for(var/obj/machinery/door/firedoor/fuckingMonsterMos in next_turf)
```
Karmic-Skink, line 150: Yeah, lets not be cussin'

Karmic-Skink (changes requested):
Just a general code standard here, can we please be a little bit more professional? Thanks.

Kmc2000:
:(

code/__DEFINES/flags.dm
```diff
@@ -168,6 +168,16 @@ GLOBAL_LIST_INIT(bitflags, list(1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 204
...
+//Knpc pathfinding return values
+#define KNPC_PATHFIND_SUCCESS 0 //Path successfully generated
+#define KNPC_PATHFIND_SKIP 1 //Regenerating of path isn't needed, incapable of moving, etc
+#define KNPC_PATHFIND_TIMEOUT 2 //Pathfinding is currently in timeout due to having failed previously.
+#define KNPC_PATHFIND_FAIL 3 //No path found
+
+//Knpc pathfinding timeout defines
+#define KNPC_TIMEOUT_BASE (3 SECONDS) //The base timeout applied if an knpc's pathfinding fails.
+#define KNPC_TIMEOUT_STACK_CAP 9 //Every consecutive pathfinding fail adds a stacks; Timeout applied uses them as multiplier up to a cap of this value.
+
```
Bokkiewokkie, line 180: This file is for bitflags put these regular defines somewhere else

code/game/machinery/doors/firedoor.dm
```diff
@@ -508,6 +508,32 @@
...
+		return TRUE
+	if(!density)
+		return TRUE
+	if(!(dir in dir_to_cardinal_dirs(to_dir)))
+		return TRUE
+	if(welded)
+		return FALSE
+	if(!hasPower())
+		return isknpc(caller)
+	return TRUE
```
Bokkiewokkie, line 523: Should probably be ported upstream as well (excluding the knpc proc of course)

nsv13/code/__HELPERS/overmap.dm
```diff
@@ -1,11 +1,15 @@
...
 	if(isovermap(loc))
 		return loc
 	if(!z) // We're in something's contents
 		if(!loc) // Or not...
 			return FALSE
 		return loc.get_overmap() // Begin recursion!
+	if(!SSmapping.z_list) //Being called by pre-mapping init call
+		if(!failsafe) //Only retry once
+			addtimer(CALLBACK(src, .proc/get_overmap, TRUE), 30 SECONDS)
+		return FALSE
```
Bokkiewokkie, line 12: I can only support this one proc getting more absurdly complex purely to handle all the edge cases

Bokkiewokkie (changes requested):

code/__DEFINES/flags.dm
Bokkiewokkie: A seperate file in the nsv folder is also fine

DeltaFire15: .. I thought I had those somewhere reasonable, huh..

DeltaFire15:
> KNPCs are trying to get onto glass tables, even if there is another simpler path, stunning themselves in the process

Hmm, this actually opens up a problem of JPS as it is implemented currently. Issue is, right now all tile steps "cost" the same, which well, opens up issues with things that should be *able* the be pathed through, but which are more detrimental to the pather. Right now, either they see it as a normal tile pathing wise, or they never paththrough it, which, neither of which should be the case for glass tables (or tables in general). I guess for now just not letting them path glass tables would work because I don't feel like reworking JPS in general.

Karmic-Skink:
> > KNPCs are trying to get onto glass tables, even if there is another simpler path, stunning themselves in the process
> 
> Hmm, this actually opens up a problem of JPS as it is implemented currently. Issue is, right now all tile steps "cost" the same, which well, opens up issues with things that should be _able_ the be pathed through, but which are more detrimental to the pather. Right now, either they see it as a normal tile pathing wise, or they never path through it, which, neither of should be the case for glass tables (or tables in general). I guess for now just not letting them path glass tables would work because I don't feel like reworking JPS in general.

Yeah not pathing over glass tables is also an issue. A little disappointed they didn't actually add support for pathing costs, but oh well.

covertcorvid (approved):

Bokkiewokkie (approved):

PR #1999, opened 2022-07-24
Boarder refactor

Kenionatus:
## About The Pull Request
Improves boarding code a bit. The addition of minimum player arguments allows to spawn boarders with a proc call on the ship, even if player numbers are below the normal minimum. Maybe I'll add a better way to call that in the future.
Adds some doc comments to procs. (Putting doc comments in separate lines above vars has been deemed too inconsistent with existing code by maintainers, so that was reverted.)
Makes boarders able to differentiate between guns they can fire and guns they can't when they are looking for one to pick up. Uses the same function as the gun when firing if I'm not mistaken, so should be pretty all encompassing and future proof.

## Why It's Good For The Game
Doc comments make the code more readable with an IDE, as long as they don't make the code itself less readable. Moving code into functions makes code more readable. Boarders picking up guns they can't fire unintentionally weakens them in those situations.

## Changelog
:cl:
add: Boarders now only pick up guns they can actually fire.
refactor: Adds some doc comments to procs in knpc code.
refactor: Improves boarding code a bit.
admin: Adds arguments to spawn_boarders proc of overmap ships. This allows to manually adjust minimum player numbers. Arguments are now "amount, faction_selection="syndicate", min_players = 5, min_players_for_ghosts = 20"
/:cl:


Karmic-Skink (changes requested):
Reformating all of these comments will cause the comments to be inconsistent with the majority of the rest of the NSV13 codebase.

It also makes it harder to review your actual code changes.

Code review pendingTM

Kenionatus:
> will cause the comments to be inconsistent with the majority of the rest of the NSV13 codebase
Does that mean there is an official code standard to never use above line doc comments for variable documentation in favour of end of line comments?

DeltaFire15:
Inconstistancy with current nsv code comments is of little matter when the entire thing is that most nsv things have little autodoc functionality actually used which is very useful when properly implemented due to the huge variety of individuals working on code.
Autodoc comment changes look good imo.

nsv13/code/modules/overmap/boarding/boarding.dm
```diff
@@ -28,57 +66,64 @@
...
-		return TRUE
+/** Attempts to spawn knpc boarders for this ship.
+ * Arguments:
+ * * amount: Number of boarders to spawn. Will be automatically choosen, based on SSovermap mode difficulty if !amount.
+ * * faction_selection: String for the faction the boarders belong to. See [/obj/structure/overmap/proc/spawn_boarders] for valid values.
+ * Returns TRUE is successful, FALSE otherwise.
+ */
+/obj/structure/overmap/proc/spawn_knpcs(amount = 3, faction_selection="syndicate")
+	if(!amount)
+		amount = CEILING(1 + (SSovermap_mode.mode.difficulty / 2), 1)
```
Karmic-Skink, line 80: Why are we defining an amount in the base proc if we then have code to handle if there is no amount defined?

nsv13/code/modules/overmap/boarding/boarding.dm
```diff
@@ -118,21 +163,23 @@
...
 			var/endings = strings(PIRATE_NAMES_FILE, "endings")
-			H.fully_replace_character_name(H.real_name, "[callsign] [pick(beggings)][pick(endings)]")
+			H.fully_replace_character_name(H.real_name, "[callsign] [pick(beginnings)][pick(endings)]")
 			H.mind.add_antag_datum(/datum/antagonist/pirate/boarder)
 			H.mind.assigned_role = "Pirate Boarder"
 			log_game("[key_name(H)] became a space pirate boarder.")
 			message_admins("[ADMIN_LOOKUPFLW(H)] became a space pirate boarder.")
 			operatives += H
-		 //No audio warning?
+		relay('nsv13/sound/effects/ship/boarding_pod.ogg', "<span class='userdanger'>You can hear several tethers attaching to the ship.</span>")
```
Karmic-Skink, line 180: Space Pirate Boarders having no audio warning is intentional.

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -366,11 +380,17 @@ This is to account for sec Ju-Jitsuing boarding commandos.
 			if(H.put_in_hands(target_item))
 				H.visible_message("<span class='warning'>[H] grabs [target_item]!</span>")
 
+///Checks that G exists, has ammo and that H can fire it. Returns G if yes, FALSE otherwise.
+/datum/ai_goal/human/acquire_weapon/proc/gun_suitable(mob/living/carbon/human/H, obj/item/gun/G)
+	if(!G)
+		return FALSE
+	return G.can_shoot() && G.can_trigger_gun(H) ? G : FALSE
```
Karmic-Skink, line 387: Include a check for either the KNPC having ammo for that gun or the gun itself actually having ammo.

covertcorvid:
> > will cause the comments to be inconsistent with the majority of the rest of the NSV13 codebase
> 
> ~Does that mean there is an official code standard to never use above line doc comments for variable documentation in favour of end of line comments?~
> 
> Edit: kinda got offended/scared/upset at the thought of my work not being accepted, even though it is "totally superior to the previous version of the code". I guess Code review pendingTM means I'll hear an official stance later? Also, there is a form of doc comment that should work inline `//!`, but it's currently broken. Not sure if it would be worth using it in the hope it'll get fixed.

autodoc2

/// appears to work inline

nsv13/code/modules/overmap/boarding/boarding.dm
Kenionatus: This provides an interface to pass in a custom amount. This happens if spawn_boarders is called with a custom amount (for instance for adminbus).

nsv13/code/modules/overmap/knpc.dm
Kenionatus: Whether the gun has ammo is checked in G.can_shoot()
Checking whether the knpc has ammo on themselves for the gun is a good idea, I was just too lazy to add that too. (Wasn't checked previously either and I was focussing on the firing pins. One step at a time.)

nsv13/code/modules/overmap/boarding/boarding.dm
Kenionatus: 👍

Kenionatus:
@covertcorvid Did you actually reload the language server? Because it doesn't work on my machine.

nsv13/code/modules/overmap/boarding/boarding.dm
Karmic-Skink: Yes, however I'm questioning the decision to predefine it to be 3 and then also include code for when its not defined.

nsv13/code/modules/overmap/knpc.dm
Karmic-Skink: They do sometimes get disarmed or drop weapons when climbing over things. So picking up gun if they still have mags would be a good idea.

nsv13/code/modules/overmap/boarding/boarding.dm
Karmic-Skink: Just remove amount = 3 and let the catch resolve it.

covertcorvid:
> @covertcorvid Did you actually reload the language server? Because it doesn't work on my machine.

I got duped, it was showing the comment from the line above

nsv13/code/modules/overmap/boarding/boarding.dm
Kenionatus: Oh yeah, that's totally a bug. Thanks for catching it.

Kenionatus:
Interesting, the bot removed the Administration label which Bookie manually added.

nsv13/code/modules/overmap/boarding/boarding.dm
```diff
@@ -1,18 +1,56 @@
...
+		message_admins("Failed to spawn boarders for [name], does it have an interior?")
+		return FALSE //Cut off here to avoid polling people for a spawn that will never work.
+	var/list/candidates = list()
 
-	///Finds a "safe" place to dump a boarding pod, with a bit of distance from the transition edge to avoid visual hiccups.
+	//20 or more players? You're allowed "real" boarders.
+	if(player_check >= min_players_for_ghosts) // Remove the low pop boarder camping
+		candidates = pollCandidatesForMob("Do you want to play as a boarding team member?", ROLE_OPERATIVE, null, ROLE_OPERATIVE, 10 SECONDS, src)
+	//No candidates? Well! Guess you get to deal with some KNPCs :))))))
+	if(!LAZYLEN(candidates))
```
TheFakeElon, line 39:
```suggestion
	if(!length(candidates))
```
lazylist defines should only be used for lazy lists

nsv13/code/modules/overmap/boarding/boarding.dm
```diff
@@ -28,57 +66,64 @@
...
+			knpc_types = list(/mob/living/carbon/human/ai_boarder/assistant)
+		if("nanotrasen")
+			knpc_types = list(/mob/living/carbon/human/ai_boarder/ert, /mob/living/carbon/human/ai_boarder/ert/commander, /mob/living/carbon/human/ai_boarder/ert/medic, /mob/living/carbon/human/ai_boarder/ert/engineer)
+
+	var/list/possible_spawns = list()
+	for(var/obj/effect/landmark/patrol_node/node in GLOB.landmarks_list)
+		if(node.get_overmap() == src)
+			possible_spawns += node
+	//Can we establish a drop zone?
+	var/turf/LZ = (possible_spawns?.len) ? get_turf(pick(possible_spawns)) : null
```
TheFakeElon, line 97:
```suggestion
	var/turf/LZ = (length(possible_spawns)) ? get_turf(pick(possible_spawns)) : null
```

nsv13/code/modules/overmap/boarding/boarding.dm
```diff
@@ -28,57 +66,64 @@
...
+	if(!LZ)
+		message_admins("KNPC boarder spawn aborted. This ship does not support KNPCs (add some patrol nodes!))")
+		return FALSE
+
+	var/obj/structure/closet/supplypod/centcompod/toLaunch = new /obj/structure/closet/supplypod/syndicate_odst
```
TheFakeElon, line 102: Is there a reason you're casting a regular supply pod to a centcompod, or was this an accident?
```suggestion
	var/obj/structure/closet/supplypod/centcompod/toLaunch = new()
```

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -339,12 +352,13 @@ This is to account for sec Ju-Jitsuing boarding commandos.
 		var/dist = get_dist(O, H)
 		if(istype(O, /obj/structure/closet) && dist <= best_distance)
 			var/obj/structure/closet/C = O
-			target_item = locate(/obj/item/gun) in C.contents
-			if(target_item && C.allowed(H))
+			var/obj/item/gun/G = locate(/obj/item/gun) in C.contents
+			if(C.allowed(H) && gun_suitable(H, G))
+				target_item = G
```
TheFakeElon, line 357: It's better to null check here instead of in gun\_suitable(). Callers should handle null args, not procs

nsv13/code/modules/overmap/knpc.dm
TheFakeElon, line 387:
```suggestion
	return G.can_shoot() && G.can_trigger_gun(H)
```

Kenionatus:
Currently working on doing #1951 the way it is requested, so it'll take me a while to get around to handle the feedback.

nsv13/code/modules/overmap/boarding/boarding.dm
Kenionatus: Thanks, didn't know length was more performant.

nsv13/code/modules/overmap/boarding/boarding.dm
Kenionatus: fixed (tho I deflect blame for that to IndusRobot)

nsv13/code/modules/overmap/boarding/boarding.dm
Kenionatus: Not my code, so I'm hesitant to change that.

nsv13/code/modules/overmap/boarding/boarding.dm
```diff
@@ -28,57 +66,64 @@
...
+ * Arguments:
+ * * amount: Number of boarders to spawn. Will be automatically choosen, based on SSovermap mode difficulty if !amount.
+ * * faction_selection: String for the faction the boarders belong to. See [/obj/structure/overmap/proc/spawn_boarders] for valid values.
+ * Returns TRUE is successful, FALSE otherwise.
+ */
+/obj/structure/overmap/proc/spawn_knpcs(amount, faction_selection="syndicate")
+	if(!amount)
+		amount = CEILING(1 + (SSovermap_mode.mode.difficulty / 2), 1)
+	var/list/knpc_types = list()
+	switch(faction_selection)
```
Karmic-Skink, line 82: This is missing a zombie selection.

nsv13/code/modules/overmap/boarding/boarding.dm
TheFakeElon: you should probably change it

Kenionatus: It would make sense to use a syndicate supply pod but I don't know whether that would cause any issues. I'm currently busy with other code and my mental health being a wreck, so changing this is not on the top of my priority list, but I'll try to have a look at it sometime.

Kenionatus:
I've started introducing error handling by returning error values (defined in nsv13\code\__DEFINES\boarding.dm). It starts to feel a bit anachronistic tho (like C code).
Think
```switch(faction_selection)
    if(syndicate)
        do something
    ditto for other factions
    else
        return KNPC_SPAWN_INVALID_FACTION
```
The alternative I see would be to use try/catch, but that would mean even further indentation, begging me to split it up into further functions (which might not be such a bad thing...)

Kenionatus:
Spawning zombies does currently cause a runtime because they don't have a job but the spawn code tries to give them an ID with their job on it.

Bokkiewokkie (approved):

PR #2004, opened 2022-07-24
Fixes sabres being able to bring all overmap objects into a system

covertcorvid:
## About The Pull Request
After some reserved Z fixes, if a sabre tried to enter a system with a reserved z of 0 and nothing occupying that system, all the overmap objects would come out of storage but not be moving.

## Why It's Good For The Game
This bug is bad.

## Changelog
:cl:
fix: Fixed sabre undocking importing all the objects from nullspace
/:cl:


covertcorvid:
Jumped without bringing stuff with me in round 4663

PR #2010, opened 2022-08-06
Stops ships from firing while being deleted

DeltaFire15:
## About The Pull Request
Title.
Some conditions can cause a ship to attempt firing while being deleted or in nullspace. This adds a check to the proc firing a projectile to prevent this.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Ships shouldn't try firing while being deleted.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/physics.dm
+++ b/nsv13/code/modules/overmap/physics.dm
@@ -568,6 +568,8 @@ This proc is to be used when someone gets stuck in an overmap ship, gauss, WHATE
 	return ..()
 
 /obj/structure/overmap/proc/fire_projectile(proj_type, atom/target, homing = FALSE, speed=null, user_override=null, lateral=FALSE, ai_aim = FALSE, miss_chance=5, max_miss_distance=5) //Fire one shot. Used for big, hyper accelerated shots rather than PDCs
+	if(!z || QDELETED(src))
+		return FALSE
 	var/turf/T = get_center()
 	var/obj/item/projectile/proj = new proj_type(T)
 	if(ai_aim && !homing && !proj.hitscan)


nsv13/code/modules/overmap/physics.dm
```diff
@@ -568,6 +568,8 @@ This proc is to be used when someone gets stuck in an overmap ship, gauss, WHATE
 	return ..()
 
 /obj/structure/overmap/proc/fire_projectile(proj_type, atom/target, homing = FALSE, speed=null, user_override=null, lateral=FALSE, ai_aim = FALSE, miss_chance=5, max_miss_distance=5) //Fire one shot. Used for big, hyper accelerated shots rather than PDCs
+	if(!z || QDELETED(src))
```
Bokkiewokkie, line 571:
```suggestion
	if(!loc || QDELETED(src))
```
wouldn't you just want to check for the null loc? z=0 when you're in the contents of an object

Bokkiewokkie (changes requested):

nsv13/code/modules/overmap/physics.dm
DeltaFire15: z is 0 when you aren't on a turf and there is no case where a overmap should be not on a turf and firing, is there?

Bokkiewokkie: > z is 0 when you aren't on a turf and there is no case where a overmap should be not on a turf and firing, is there?

There *shouldn't* but just to be safe y'know.

DeltaFire15: ... this *is* the "to be safe" option. There is no occasion any overmap should be *firing* while not on a turf.

TheFakeElon (commented):
I don't think this is an ideal fix. The issue most likely stems from overmaps being destroyed while peforming asynchronous tasks, in which case the checks should be done in that loop, not the proc(s) it calls

DeltaFire15:
> I don't think this is an ideal fix. The issue most likely stems from overmaps being destroyed while peforming asynchronous tasks, in which case the checks should be done there, not the proc(s) it calls

The proc this check has been added in is the only common link inbetween everything that tries to fire a bullet. Adding it here or in all the places this proc gets called in is pretty much the same processing-wise, but likely more long-term consistant (aka someone just calling fire_projectile somewhere else some other time) and less prone to having overlooked a case (like for example a system being "unloaded" triggering this aswell). I'd rather be done with this bug than reinvestigate it as soon as someone writes a new firing method or flak variation.

TheFakeElon:
The bug isn't game breaking, it should be fixed but it's certainly not a high priority issue as far as I can see. It'd be better to fix it properly than bandaid it.
It's not the future proof solution either, it only stops the issue from manifesting in this one specific proc, any other procs that are/can be called by the nullspaced object can cause these issues too

covertcorvid (approved):

Bokkiewokkie (approved):

PR #2041, opened 2022-08-31
Adds balefire/fusion torps.

sunofang:
## About The Pull Request

Adds torpedos that release fusion and a fleet that carries said torpedos.
God forgive me, For i have sinned.

## Why It's Good For The Game

More hard enemies is always good. And having a fleet for the adminbus to send at bad crews is good too.

## Changelog
:cl:
add: Added balefire fleets. Special fleets that carry balefire torpedos.
add: Added balefire torpedos. Torpedos that release fusion.
imageadd: Added the icons for balefire destroyers and torps.
/:cl:


covertcorvid:
We don't actually want an instant round-ending projectile. That doesn't seem fun.

Karmic-Skink:
This is just excessive as a weapon and at the same time actually does very little other than upset the atmos system.
Nukes/Superweapons will return at some point in time, but in a better fashion than just fire etc.

As it is balefire, let us leave it outside (of the codebase).

-1

sunofang:
> We don't actually want an instant round-ending projectile. That doesn't seem fun.

In my tests it doesnt acutally do much, As it either melts the floor or gets confined by firelocks.
And yes i agree its slightly excessive. Maybe it would be best to keep it as an admin-only weapon?

Bokkiewokkie:
> > We don't actually want an instant round-ending projectile. That doesn't seem fun.
> 
> In my tests it doesnt acutally do much, As it either melts the floor or gets confined by firelocks. And yes i agree its slightly excessive. Maybe it would be best to keep it as an admin-only weapon?

We still have an admin-only nuclear_impact() proc if the ship needs to be toasted, so this isn't needed for that either

PR #2056, opened 2022-09-10
prevent the SD from burning out lights

red031000:
## About The Pull Request

Prevents the Stormdrive from burning out lights, as discussed on discord

## Why It's Good For The Game

The current meta is to have a high powered stormdrive to produce enough power for the entire ship and have spare, some ships like the aetherwhisp requires the stormdrive to be very high powered in order to power everything it needs, however this often leads to lights burning out around the ship, and plunging the whole engineering department (and sometimes the entire ship) into darkness
Replacing these lights leads to more burnouts

## Changelog
:cl:
tweak: stops the stormdrive from burning out lights
/:cl:

### DIFF
--- a/nsv13/code/modules/power/stormdrive.dm
+++ b/nsv13/code/modules/power/stormdrive.dm
@@ -989,10 +989,7 @@ Control Rods
 							L.flicker()
 			if(prob(1))
 				for(var/obj/machinery/light/L in orange(10, src))
-					if(prob(25))
-						L.burn_out()
-					else
-						L.flicker()
+					L.flicker()
 			if(prob(0.01))
 				for(var/obj/machinery/power/grounding_rod/R in orange(5, src))
 					R.take_damage(rand(25, 50))
@@ -1003,9 +1000,6 @@ Control Rods
 				for(var/obj/machinery/light/L in GLOB.machines)
 					if(prob(50) && shares_overmap(src, L))
 						L.flicker()
-			if(prob(5))
-				for(var/obj/machinery/light/L in orange(12, src))
-					L.burn_out() //If there are even any left by this stage
 			if(prob(0.1))
 				for(var/obj/machinery/power/grounding_rod/R in orange(8, src))
 					R.take_damage(rand(25, 75))


Karmic-Skink:
This is the intended state and consequences of using a high powered Stormdrive.

The "meta" state of leaving it cranking at a bazillionty MW is a decision made by the players out of laziness, the consequences are designed as minor inconveniences for using the Stormdrive as a set and forget engine, rather than managing the power output as required.

red031000:
not true, to use the whisp as an example again, it requires 15MW for shields, 3MW for the pumps, and at least 3MW for the guns, by default this can only draw from the SD, this greatly exceeds the "excessive" 8MW output

covertcorvid:
The Whisp is an exception rather than an example. If necessary we could consider scaling the energy requirements of the Whisp's weapons downwards since they are not buildable. However, the original intent as I understand it was that the SD be run at a lower level most of the time, and increased to higher levels to repair armour and recharge weapons.

Karmic-Skink:
> not true, to use the whisp as an example again, it requires 15MW for shields, 3MW for the pumps, and at least 3MW for the guns, by default this can only draw from the SD, this greatly exceeds the "excessive" 8MW output

The Aetherwhisp has three engines. One of them being the Stormdrive, which was never designed to be used to solely power that ship.
Ships using higher power draws have (or should have) an AGCNR for that exact purpose, it generates more power.

As Corvid stated, the Aetherwhisp is the exception.

covertcorvid:
Neither of the statements you removed would knock out lights across the entire ship by the way. This effect extends not more than 12 tiles from the drive.

Karmic-Skink:
> I agree with @Karmic-Skink in the point that a meta of actively managing the SD during high power demand is better than set and forget. However, I don't think breaking or even just flickering lights are a good way to achieve that. It's, as Karmic said, a minor inconvenience, which doesn't stop engies who like to eek out every bit of power from doing so out of combat. Also, in their current state, both engines don't really lend themselves to variable power production. On the side of the RBMK, it's the very slow ramp up and on the side of the SD its that rods decay faster with higher insertion (correct me if I'm wrong and the lower temp completely cancels that out). Changing the mix for either engine is virtually impossible when using the fuel mix tank (or rather insanely slow).
> 
> I think if we really want to encourage active engine management, we need a different way to do so. One quick, off the top of my hat, idea would be to make the control rod efficiency vary over time (up and down). The higher the power output, the stronger and quicker the variation.

What are you guys doing with mix tanks that doesn't allow for rapid changing of mixtures? Pumping in at 4500kpa?

Rod decay rate is correlated with insertion amount, but that isn't the full picture, it relates to reaction rate or should do so.

But the tl;dr of this is that we aren't going to accept a PR that is just removing a feature, come up with a more interesting and better solution than just axing a consequence of exceeding design parameters/falling into the intended trap.

covertcorvid:
Not so much a volume pump as an injector and a closed room, but the difference is mostly irrelevant.

Bokkiewokkie:
>

Bokkiewokkie:
misinput

IndusRobot (commented):
Instead of removing `burn_out`, try replacing this with a new proc call under the stormdrive object that calls `burn_out`. Then update this proc on the child object (aether's snowflake mk5 ion stormdrive) to jus return false or do nothing 

```
/obj/stormdrive/proc/handle_burn_out()
burn_out()

/obj/stormdrive/mk5/handle_burn_out()
//does nothing
```

This way you get the functionality you want on the ship causing the light fixture problem, without affecting other ships

Kenionatus:
It's not difficult to replace the light, but imo that adds to the point: it's a mild annoyance which doesn't really work that well to encourage engineers to actively manage the engine (they often don't have to do the light replacing themselves since it's under the janitor's responsibility).

Karmic-Skink:
> It's not difficult to replace the light, but imo that adds to the point: it's a mild annoyance which doesn't really work that well to encourage engineers to actively manage the engine (they often don't have to do the light replacing themselves since it's under the janitor's responsibility).

The assumption there that this is intentionally placed here to encourage engineers to manage their engines is somewhat false.
The intention is to show instability in the grid due a higher load, if you would like to offer a suggestion for improvement, please do so.

Karmic-Skink:
> Just did.

Removing is not improving, so no, you didn't.

The Aetherwhisp has 3 reactors available which is an _outlier_ on the mapping standards.

General rule of thumb as follows:
- Smaller ships with lower power requirements use the Stormdrive
- Larger ships with higher power requirements use the AGCNR

Kenionatus:
> The "meta" state of leaving it cranking at a bazillionty MW is a decision made by the players out of laziness, the consequences are designed as minor inconveniences for using the Stormdrive as a set and forget engine, rather than managing the power output as required.

> The assumption there that this is intentionally placed here to encourage engineers to manage their engines is somewhat false.

Those two statements seem to contradict each other.

> The intention is to show instability in the grid due a higher load

But why? What is the point of that instability, if not to encourage active SD management?

Karmic-Skink:
> 



> > The "meta" state of leaving it cranking at a bazillionty MW is a decision made by the players out of laziness, the consequences are designed as minor inconveniences for using the Stormdrive as a set and forget engine, rather than managing the power output as required.
> 
> > The assumption there that this is intentionally placed here to encourage engineers to manage their engines is somewhat false.
> 
> Those two statements seem to contradict each other.
> 
> > The intention is to show instability in the grid due a higher load
> 
> But why? What is the point of that instability, if not to encourage active SD management?

The end goal intention of that net instability is real personal danger from being near the device (hence tesla zaps), but there needed to be intermediate steps to show some degree of ramp up in the loss of grid stability.

Karmic-Skink:
> Mapping standards were never explicitly made public, so this does not apply

That actual decision was made by KMC and Myself when the AGCNR was coded and predates official mapping standards.

covertcorvid:
> Mapping standards were never explicitly made public, so this does not apply

This isn't relevant though, the Aetherwhisp with its higher power draw and many engines is an outlier among our ships. That's just a fact.

covertcorvid:
> Just did.
> 
> @Kenionatus Alternatively you could change the max power before lights flicker and burnout on the mk5 child object to something that the aether is expecting for operation. About 25MW should do, as typical operation calls for powering shields (15MW), armor pumps (3MW), lasers (about 8MW) and the rest of the ship (500something kW)

The SD on its own shouldn't be powering all of this though, is the thing. That's why the ship has so many engines.

covertcorvid:
Something fishy is going on then, since the burnout chances for each of these are 0.25% and 5%

red031000:
the main reason is due to the extremely frequent burnouts, if they were *significantly* more rare (i.e. only a few times per shift) imo that would also be acceptable (and would likely give the janitor something to do, which is an added bonus)

covertcorvid:
> the main reason is due to the extremely frequent burnouts, if they were _significantly_ more rare (i.e. only a few times per shift) imo that would also be acceptable (and would likely give the janitor something to do, which is an added bonus)

If you can modify the frequency of the lights breaking or come up with a different warning sign for high power levels to replace it with, that would be better than just removing this altogether.

Bokkiewokkie:
Closing this because an alternative fix was already merged for the thing adressed in this PR. Further discussion should be moved to a thread on the forum, git discussions or discord chat.

PR #2064, opened 2022-09-18
Alicorn 2: The Great Smoothmove boogaloo

QAI-07:
## About The Pull Request

As it turns out, having a ship sprite size that is not divisable by 32 breaks smooth movement for _**SOME**_ godforsaken reason. This fixes it, as well as slaps a few bugs that came up in testing.

## Why It's Good For The Game

The Alicorn no longer breaks features that it has nothing to do with, yay. In layman's terms, we can actually use it now.
Fix man good.

## Changelog
:cl:

tweak: tweaked the taunt the Alicorn says when it enters the system
tweak: tweaked the naming routine of the Alicorn's fighter squadron to be shorter and have random numbers assigned.
tweak: the Alicorn will now use Battleship AI instead of destroyer, hopefully meaning it'll peruse it's target a bit more ruthlessly.
balance: rebalanced the special weapon and round of the Alicorn to hopefully be a bit easier to avoid.
fix: fixed the alicorn destroying smooth movement when it so much as dares to exist.

/:cl:

### DIFF
--- a/nsv13/code/datums/weapon_types.dm
+++ b/nsv13/code/datums/weapon_types.dm
@@ -378,11 +378,10 @@
 	name = "Prototype Bluespace Artillery"
 	default_projectile_type = /obj/item/projectile/bullet/prototype_bsa
 	burst_size = 1
-	fire_delay = 32 SECONDS
+	fire_delay = 22 SECONDS
 	range_modifier = 200
 	overmap_firing_sounds = list('nsv13/sound/weapons/bsa_fire.ogg')
 	overmap_select_sound = 'nsv13/sound/weapons/bsa_select.ogg'
 	ai_fire_delay = 32 SECONDS
-	lateral = TRUE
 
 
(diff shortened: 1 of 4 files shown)


QAI-07:
image

nsv13/code/datums/weapon_types.dm
```diff
@@ -378,11 +378,10 @@
 	name = "Prototype Bluespace Artillery"
 	default_projectile_type = /obj/item/projectile/bullet/prototype_bsa
 	burst_size = 1
-	fire_delay = 32 SECONDS
+	fire_delay = 22 SECONDS
 	range_modifier = 200
 	overmap_firing_sounds = list('nsv13/sound/weapons/bsa_fire.ogg')
 	overmap_select_sound = 'nsv13/sound/weapons/bsa_select.ogg'
 	ai_fire_delay = 32 SECONDS
-	lateral = TRUE
```
covertcorvid, line 386: TRUE was the default value anyways, so this won't change its firing arc at all. Which is fine, I just want to make sure that's what you wanted.

covertcorvid (approved):

Karmic-Skink:
You were trying non 32 divis?

nsv13/code/datums/weapon_types.dm
QAI-07: Ah that's just a leftover from experimentation with ship weapons, not needed.

PR #2069, opened 2022-09-21
Stormdrive overload numbers tweak

Karmic-Skink:
## About The Pull Request
This PR changes some of the overload state numbers to reduce the amount of that lights burn out, but slightly increase the rate tesla related incidents.

## Why It's Good For The Game
Less minor maint, more major maint?

## Changelog
:cl:
balance: Rebalanced stormdrive overload state
/:cl:


covertcorvid (approved):
Limiting the number of lights to attempt to break each pass should help quite a bit.

Bokkiewokkie (approved):

red031000:
even with these changes, it still feels like lights are breaking too often, especially on the whisp

red031000:
or rather, the issue is that many of the lights burn out at the same time (perhaps too many) rather than just a few, or one or two burning out at once but more often, that would be better imo

Bokkiewokkie:
> even with these changes, it still feels like lights are breaking too often, especially on the whisp

The main cause on the whisp is the AGCNR not the SD

red031000:
no? I was under the impression the AGCNR could only cause the burnouts if it's wired directly into the grid. Most of the time it goes straight into the SMES, and power is distributed from there. if I am wrong please feel free to correct me

if you are right then it's likely the numbers need tweaking for the AGCNR as well on the whisp

Karmic-Skink:
> no? I was under the impression the AGCNR could only cause the burnouts if it's wired directly into the grid. Most of the time it goes straight into the SMES, and power is distributed from there. if I am wrong please feel free to correct me
> 
> if you are right then it's likely the numbers need tweaking for the AGCNR as well on the whisp

We aren't going to tailor our engines for the Aetherwhisp.

red031000:
it still is an issue that this occurs on the whisp, and if it's not being adjusted for generally it needs to be adjusted specifically for the whisp

Karmic-Skink:
> it still is an issue that this occurs on the whisp, and if it's not being adjusted for generally it needs to be adjusted specifically for the whisp

Mapping issues are outside the scope of this PR.

red031000:
this isn't a mapping issue as it still occurs on other ships, it's just most obvious on the whisp

covertcorvid:
> this isn't a mapping issue as it still occurs on other ships, it's just most obvious on the whisp

On any map other than the Whisp, if you're running hot enough to constantly break your lights, you're outputting over 2.67x the necessary power needed to run all systems at maximum.

New thread for Whisp/solgov discussion: https://github.com/BeeStation/NSV13/discussions/2080

red031000:
if I may suggest something, rather than breaking all the lights at once, maybe break one or two at the same time, but more often, which would make it easier to manage imo

covertcorvid:
> if I may suggest something, rather than breaking all the lights at once, maybe break one or two at the same time, but more often, which would make it easier to manage imo

This is exactly what this PR does

covertcorvid:
The maximum number of lights that will break is either 3 or 5

red031000:
hm, from what I've noticed it seems like a lot of lights were breaking pretty fast, I may not have noticed it was happening in 3s to 5s then, which probably means that it's happening slightly too often then

covertcorvid:
I hacked a copy of this to run at a set power level, set it to 8 MW and put 10 lights near it, and still had at least half of them 10-15 minutes later. I think this is fine.

PR #2070, opened 2022-09-23
Fixes that shuttle shutter button issue

### LINKED ISSUE #947: KNOWN ISSUE: Cargo Shuttle's Shutters refusing to open.
## Note: This has been reported to the admins over and over again, and usually the shutters just get's replaced with atmos blockers as a temporary patch. This has happened every single round from a certain point.

### LINKED ISSUE #1015: Cargo shuttle shutters break sometimes.
Issue reported from  Round ID: 1221 (NSV 13 | Official | MRP | 98% Free Lag | Fresh Tupelo)

Reporting client version: 513.1532

1. Open the cargo shutters and close them
2. Wait until the button breaks.
3. Either you have an unopenable wall or a depressurized shuttle.

### LINKED ISSUE #2065: Shutter buttons inside cargo shuttles permanently stop working after some trips to centcom
Version: at commit a2397cd18f2e4275a27ba9649f3278035ebed1a5

## Reproduction:
Call cargo shuttle. Press one shutter button twice. Send cargo shuttle back. Call shuttle again. Press the same button twice. Repeat until shutters stop opening and closing. Send cargo shuttle back and call it again. press button twice. Repeat until satisfied.

## Observed Behaviour:
Shutters work two times (two shuttle calls), but permanently stop working afterwards. That process only starts when the shutter button has been used. They continue to be not work for the rest of the round. If you view the button's variables, you will notice that the shutters stop working when the button's `device` variable gets set to `null`.

Addition: Buttons situated on the main ship don't appear to break, which is why it only really impacts gameplay on the Tycoon, which doesn't have any ship side buttons for the shuttle shutters (afaik).

## Expected Behaviour:
Pressing the button should always open the shutter as long as there is power and it wasn't manipulated in other ways.

Bobbanz1:
## About The Pull Request

Fixes the following issues:
Resolves #2065
Resolves #947
Resolves #1015

Addendum: This PR now makes the cargo shuttle not sell the contents of any machines that are anchored

Someone on Bee apparently fucked up when they made the supply shuttle code and the export code also include everything contained _inside_ anchored objects. 

Specifically this PR [#2412](https://github.com/BeeStation/BeeStation-Hornet/pull/2412)

So what's been happening to us for the past two fucking years is that the supply shuttle code is selling everything that's allowed to be sold inside anchored objects, which is probably fine. 

The problem occurs when you also include it deleting anything inside those anchored objects that can't be sold, so basically the following is happening: 
1.  Shuttle gets sent to Centcom
2.  Shuttle exports everything inside of its area, including things INSIDE anchored objects, such as the buttons and their electronics
3. After having gone through the list of sellable objects and, being unable to find a match with the electronic, it proceeds to delete them.

This PR adds a conditional that prevents ANCHORED machines on the supply shuttle, from having their internal items deleted.

## Why It's Good For The Game
Increased Cargo Efficiency by allowing them to actually fucking use the shutters again instead of going through the doors.

## Changelog
:cl:
fix: Fixes so button electronics no longer get sold!
/:cl:


Bokkiewokkie (commented):
Please port this fix to Beestation as well and remove the //nsv13 tags

covertcorvid (approved):

PR #2086, opened 2022-10-04
Lets maps use the mining ship again

covertcorvid:
## About The Pull Request
It re-adds the part of the code where you can have a mining ship as an opt-in

## Why It's Good For The Game
Stops making admins have to fix the Galactica

## Changelog
:cl:
fix: Allowed mining ships to exist again
/:cl:

### DIFF
--- a/code/datums/map_config.dm
+++ b/code/datums/map_config.dm
@@ -149,15 +149,15 @@
 		log_world("map_config space_empty_levels is not a number!")
 		return
 
-	if(!("mine_disable" in json)) //Bypass mineload so we don't load any mining vessels period.
+	if("mine_disable" in json)
+		mine_disable = json["mine_disable"]
+	if(!("mine_disable" in json) || !mine_disable) //Bypass mineload so we don't load any mining vessels period.
 		mine_file = json["mine_file"]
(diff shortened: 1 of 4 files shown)


code/datums/map_config.dm
```diff
@@ -149,15 +149,15 @@
 		log_world("map_config space_empty_levels is not a number!")
 		return
 
-	if(!("mine_disable" in json)) //Bypass mineload so we don't load any mining vessels period.
+	if("mine_disable" in json)
+		mine_disable = json["mine_disable"]
+	if(!("mine_disable" in json) || !mine_disable) //Bypass mineload so we don't load any mining vessels period.
```
Bokkiewokkie, line 154: Wait does if("mine_disable" in json) mean if the variable is present, and not the value of said variable?
Meaning we were never actually checking if it was set to 0 or 1, so by setting the default to 1 all attempts to make it 0 again in the json also failed?

Bokkiewokkie (approved):

code/datums/map_config.dm
covertcorvid: Good point

Bokkiewokkie (approved):

PR #2089, opened 2022-10-05
Fixes shield FX not deleting properly

### LINKED ISSUE #1923: Overmap shield overlay isn't deleted after playing its animation.
## Round ID:

4323

## Testmerges:

not relevant

## Description:

When a ship with a shield is hit, it plays a little animation to indicate the presence of the active shield. But it looks like the overlay isn't actually removed when it's done playing, leading to something like this:
image
(This is after the ship got hit by some pdcs)
These overlays also keep disappearing and reappearing as if the ship kept being hit despite nothing happening. Jumping to another system seemed to finally delete the overlays though.

TheFakeElon:
Fixes #1923

## About The Pull Request

The shield hit effect was not being deleted correctly when applied as an overlay. This bug was likely causing hard deletes too

## Changelog
:cl:
fix: fixed shield hit effect
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/shieldgen.dm
+++ b/nsv13/code/modules/overmap/shieldgen.dm
@@ -356,6 +356,7 @@
 
 /obj/effect/temp_visual/overmap_shield_hit/Destroy()
 	overmap?.vis_contents -= src
+	overmap = null
 	return ..()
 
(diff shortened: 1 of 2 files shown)


Bokkiewokkie (approved):

covertcorvid:
Now it's only like 3 shields but they're still on loop instead of playing once

PR #2091, opened 2022-10-06
MULEbot AI Summoning

### LINKED ISSUE #2027: Mulebot AI Waypoint Summoning
## Round ID:
Issue discovered as a result of adding Mulebots through https://github.com/BeeStation/NSV13/pull/1983

## Testmerges:
https://github.com/BeeStation/NSV13/pull/1983

## Reproduction:
Mulebots are listed in the remote bot control window the AI has access to, while this isn’t an issue per say as they can assist in sending crates to areas. 

The issue is the fact that they have the capability of calling the mulebot remotely to some destination, which is normal for every bot, the problem is that the mulebot is incapable of following such an order, instantly defaulting to finding the path to the delivery beacon currently listed as their last destination. 

An example of this would be a mulebot with the destination set for medical, suddenly being summoned by the AI placing down a waypoint in security, the mulebot will start heading towards the destination someone last selected for it, which is medical in this scenario, after reaching medical it will ping the AI saying that it has reached the destination and is heading back to its home beacon.

Bobbanz1:
## About The Pull Request
Resolves #2027

Makes it so the MULEbot automatically resets the next_destination to null, rather than constantly remembering it from the last time it was changed.

## Why It's Good For The Game
MULEbot AI Summoning is good.

## Changelog
:cl:
fix: MULEbot is now capable of being summoned by the AI repeatedly instead of only once.
/:cl:


Bobbanz1:
**_IT'S NOT A FUCKING SPRITE_**

Bokkiewokkie:
> **_IT'S NOT A FUCKING SPRITE_**

There's clearly a logged change for nuke.dmi... Why did you make IDB ignore it instead of undoing the change?
Also I would recommend starting a new branch off of master to avoid unrelated commits/changes from making it into your PRs

Bokkiewokkie (approved):

PR #2092, opened 2022-10-07
Slightly Better Collisions

Ikalpo:
## About The Pull Request

fixes the algorithm for elastic collisions

## Why It's Good For The Game

https://www.desmos.com/calculator/zivipnw7zu

## Changelog
:cl:
fix: fixed the elastic collision algorithm
/:cl:

## More Info
"debounce" is to stop things from running the collision if they aren't actually colliding, like if the collision happens twice in a row.
fun fact: wikipedia/desmos's arctan function is (Y,X) and BYOND's arctan is (X,Y)
why is it like this? who knows.
TODO: make collision damage use the same function for debounce

### DIFF
--- a/nsv13/code/modules/overmap/physics.dm
+++ b/nsv13/code/modules/overmap/physics.dm
@@ -426,28 +426,29 @@ This proc is to be used when someone gets stuck in an overmap ship, gauss, WHATE
 		var/src_vel_mag = src.velocity.ln()
 		var/other_vel_mag = other.velocity.ln()
 		//I mean, the angle between the two objects is very likely to be the angle of incidence innit
-		var/col_angle = Get_Angle(src, other)
+		var/col_angle = ATAN2(src.position.x - other.position.x, src.position.y - other.position.y)
 
 		// Elastic collision equations
-		var/new_src_vel_x = ((																	\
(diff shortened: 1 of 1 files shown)


Bokkiewokkie (approved):
📈

nsv13/code/modules/overmap/physics.dm
```diff
@@ -426,28 +426,33 @@ This proc is to be used when someone gets stuck in an overmap ship, gauss, WHATE
 		var/src_vel_mag = src.velocity.ln()
 		var/other_vel_mag = other.velocity.ln()
 		//I mean, the angle between the two objects is very likely to be the angle of incidence innit
-		var/col_angle = Get_Angle(src, other)
+		var/col_angle = ATAN2((other.position.x + other.pixel_collision_size_x / 2) - (src.position.x + src.pixel_collision_size_x / 2), (other.position.y + other.pixel_collision_size_y / 2) - (src.position.y + pixel_collision_size_y / 2))
+
+		//Debounce
+		if(((cos(src.velocity.angle() - col_angle) * src_vel_mag) - (cos(other.velocity.angle() - col_angle) * other_vel_mag)) < 0)
```
TheFakeElon, line 432: these cos operations are repeated a lot below, could be worth caching if you want to optimize it :)

PR #2099, opened 2022-10-11
Genpop radio will now use minutes instead of seconds

someone543:
## About The Pull Request
Makes the radio message sent by genpop use minutes instead of seconds

## Why It's Good For The Game
This fixes an issue where the radio message sent after using genpop uses seconds instead of minutes. This caused a 5 minute sentence to be sent as "Incarcerated for 300 minutes". 

## Changelog
:cl:
fix: genpop radio now properly uses minutes instead of seconds
/:cl:

### DIFF
--- a/nsv13/code/modules/Security/genpop.dm
+++ b/nsv13/code/modules/Security/genpop.dm
@@ -234,7 +234,7 @@
 			playsound(loc, 'sound/machines/ping.ogg', 50, 1)
 
 	var/obj/item/card/id/id = new /obj/item/card/id/prisoner(get_turf(src), desired_sentence, desired_crime, desired_name)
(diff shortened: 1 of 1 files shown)


DeltaFire15 (dismissed):
Should either round it to single digit or display as seconds since the current one could give you some wacky numbers depending on how many seconds you choose.

covertcorvid:
the timing system for this is... special, but that's not your fault

someone543:
> Should either round it to single digit or display as seconds since the current one could give you some wacky numbers depending on how many seconds you choose.

This shouldn't be a concern. Genpop can only put in minutes, it's currently impossible to assign any amount of seconds.

Bokkiewokkie (approved):
Locked up for 5 hours in a 3 hour round for your crimes

covertcorvid (approved):

PR #2107, opened 2022-10-15
Nerfs the alien infestation event

covertcorvid:
## About The Pull Request
The player requirement for alien infestation was 10 people. This doubles it.
It also reduces the number of iterations of the events system before announcing the threat.

## Why It's Good For The Game
Aliens are hard and 10 people aren't enough to deal with them, and the announcement comes well after they've gotten out of hand.

## Changelog
:cl:
balance: Increased pop requirement for midround alien infestation from 10 to 20
balance: Reduced minimum iterations before announcement from 400 to 300
/:cl:

### DIFF
--- a/code/modules/events/alien_infestation.dm
+++ b/code/modules/events/alien_infestation.dm
@@ -3,7 +3,7 @@
 	typepath = /datum/round_event/ghost_role/alien_infestation
 	weight = 5
 
-	min_players = 20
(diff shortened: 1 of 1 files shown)


Bokkiewokkie (approved):
Seems alright to me, we can always tweak it more if it turns out to be too little/much.

PR #2164, opened 2022-11-06
Tweaks astrometrics scans

benbot16:
## About The Pull Request

This PR makes system scanning using the astrometrics scanner take 15 seconds, and makes anomaly scanning take 1 minute.

## Why It's Good For The Game

At present astrometrics scanning is both the only way to find hazardous systems and the only way to get reasonable amounts of discovery research. Shortening system scan time allows the scanning computer to be used for locating hazardous systems in a timely fashion, and shortening the anomaly scan time somewhat makes it a bit less tedious to scan them.
This also makes the code for this a bit clearer.

Scanning functions properly.

## Changelog
:cl:
balance: Scanning systems using the astrometrics computer now takes 15 seconds, and scanning anomalies takes 1 minute.
/:cl:

### DIFF
--- a/nsv13/code/modules/research/astrometrics.dm
+++ b/nsv13/code/modules/research/astrometrics.dm
@@ -15,7 +15,9 @@ you build.
 	circuit = /obj/item/circuitboard/computer/astrometrics
 	var/max_range = 40 //In light years, the range at which we can scan systems for data. This is quite short.
 	var/scan_progress = 0
-	var/scan_goal = 2 MINUTES
+	var/scan_goal
+	var/scan_goal_system = 15 SECONDS
+	var/scan_goal_anomaly = 1 MINUTES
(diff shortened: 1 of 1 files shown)


covertcorvid:
Discovery research feels like it's going a little **too** fast. Could you either nudge the anomaly time back up or reduce the anomaly point values?

benbot16:
> Discovery research feels like it's going a little **too** fast. Could you either nudge the anomaly time back up or reduce the anomaly point values?

Think that'll be good?

covertcorvid:
> > Discovery research feels like it's going a little **too** fast. Could you either nudge the anomaly time back up or reduce the anomaly point values?
> 
> Think that'll be good?

Hopefully! We'll see how it feels

covertcorvid (approved):

PR #2168, opened 2022-11-07
Remove Text Macros in strings without interpolated expression

Looreydev:
## About The Pull Request

Removes unnecesary text macros, there are no interpolated expressions in these strings so the text macros will not do anything. Found this while compiling NSV in OpenDream.

## Changelog
:cl: Loorey
fix: removed unnecesary text macros in strings without interpolation
/:cl:

### DIFF
--- a/code/modules/antagonists/clock_cult/mobs/eminence.dm
+++ b/code/modules/antagonists/clock_cult/mobs/eminence.dm
@@ -1,5 +1,5 @@
 /mob/living/simple_animal/eminence
-	name = "\the Eminence"
+	name = "the Eminence"
 	desc = "A glowing ball of light."
 	icon = 'icons/effects/clockwork_effects.dmi'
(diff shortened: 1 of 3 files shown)


covertcorvid:
Core file changes should either be made upstream or have NSV13 comments

ike709:
Worth mentioning that NSV isn't going to work in OpenDream due to auxtools. At the very least atmos will be totally broken, unless you guys removed auxmos.

Looreydev:
> Worth mentioning that NSV isn't going to work in OpenDream due to auxtools. At the very least atmos will be totally broken, unless you guys removed auxmos.

Yeah I know, thanks for mentioning it again. But one can dream, maybe in the future and removing all the yucky small things until then is better than nothing.

Bokkiewokkie (approved):

PR #2214, opened 2022-12-01
Adds damage scaling to NAC shells.

benbot16:
## About The Pull Request

This adds damage and AP scaling to NAC shells based on the powder load at the time of firing.
Damage scales logarithmically on AP shells, and AP scales linearly on base NAC shells.

## Why It's Good For The Game

Shells at 500% powder load and shells at 50% powder load do the same damage and can punch through the same amount of armor, despite their speed differences with shells of the same mass. That doesn't really make sense.

Compiles properly, damage/AP scaling is applied properly.

## Changelog
:cl:
balance: TX-101 shell damage & penetration ability now scales with powder load
/:cl:


Karmic-Skink:
Our shells are considered to contain their own drilling charge.

Some basic research here:
https://en.wikipedia.org/wiki/High-explosive_anti-tank

Also, this is pretty much an unnecessary complication that will likely just cause people to only max load their powder and reduce flexibility in play.

Bokkiewokkie:
Agree with Karmic here, I'd say only make piercing and damage affected by powder load for piercing shells, and to not let it go below the amount of damage they currently do with a low amount of powder.

Bokkiewokkie (approved):

covertcorvid (approved):

PR #2228, opened 2022-12-13
Targeting Overhaul

benbot16:
## About The Pull Request

This PR generally overhauls how guided weapons are handled. I tried to touch as little as possible that wasn't on the base overmap object or projectile object, so hopefully it shouldn't conflict too much.

<details>
<summary>Gameplay Changes</summary>
You can now use DRADIS to lock onto targets. This requires you to press the mode swap button in the top right of the DRADIS panel, which switches from hailing mode to targeting mode and back.
Fired guided projectiles will now always track locked targets, instead of only tracking targets when sprite-clicked. If a locked target isn't selected, they'll track towards the first painted target.
The Tactical console now only displays locked targets, instead of displaying all targets in range. It is also used to select which target is the "locked" target. 
The Datalink is now a bit more in-depth. Clicking on a friendly when in targeting mode in DRADIS allows you to transmit your current locked target to them if they aren't at their max target paints. This allows the mainship to transmit data to fighters, as well as allowing fighters to transmit targeting data to each other.
Fighters also now have a use for the "target lock" button on their control panel. Clicking it will dump all active locks. This can also be done with the unlock hotkey, defaulting to 6.
I also gave ship gunners an optical camera, similar to way back when. This version of the camera is much more limited, only able to view targets out to your base sensor range. However, it also allows you to look across the edges of the overmap Z level, so you can nail that one destroyer that's been constantly corner peeking and glitching out your view. It's still a bit jank where Zs are concerned, but I haven't been able to fix that.
Finally, AI ships will now target incoming missiles with their point defense cannons, and will target ships locking onto them with flak.
</details>

<details>
<summary>Code Changes</summary>
Targets are now handled in two places: target_lock and target_painted. Target_painted is an associative list of the active paints of a ship, listing both the target and the ship that's providing the target with datalink (if applicable).
Torpedoes_to_target is once again used. Torpedoes are added to this list when fired at a ship with homing, and ships will target these when autonomous targeting fires. 
Additionally, most of the handling for homing has been moved off of the fire_projectile proc chain and onto the projectile itself. Projectiles now have a can_home var, which determines whether a gun will try to assign that projectile a homing target. 
The gun camera can be disabled on ships using the no_gun_cam variable, and autotarget functionality can be added to ships by setting autotarget to true. This automatically sets a painted target to be the locked target.
There's also a few new signals and a new proc that gets called when a missile is launched at a ship, which might be useful for other coders.

Most of these changes should be pretty modular, and you can override the targeting procs to add other checks or do specific things.
</details>

Please let me know if there's anything I missed or that I should add.
I was considering adding some functionality to the ATC's console to let them pass datalink targets to fighters, but that'd require mass map changes so I figured I'd hold off on it.

## Why It's Good For The Game

The targeting changes allow you a lot more freedom when launching missiles. You can now launch them at different angles to try and get around flak clouds or confuse point defenses. You can also actually make use of NAC guidance without having to sprite-click. This also gives some QOL to fighter pilots, since right now they need to spriteclick to lock onto other ships.
The datalink changes should also make the datalink system as a whole clearer, and gives pilots/bridgies more options when using it. 
Adding a targeting camera with limited range gets around some of the more annoying parts of edge combat, allowing you to see ships that are just off screen or on the other side of the z-transition. It should have a shorter range than the AI detection range, which should prevent players from just using it to snipe in a corner.
Giving AI the ability to shoot down missiles more easily also nerfs easy standoff attacks, since before this AI ships would just kind of sit there and take any long-range missiles you launched at them.

For these videos the gun camera's range was set to infinity so I could see the target. It doesn't have this much range in practice. If anyone ever wants to do the same for testing and/or adminbus shenanigans VV the ship's DRADIS and set visual_range to something really big. 
<details>

AI PDC:

https://user-images.githubusercontent.com/9423435/207431773-7b06d59c-5331-48eb-ae01-216a8dbba22b.mp4

Target locking:

https://user-images.githubusercontent.com/9423435/207432375-3e215127-727c-4d98-bc82-bfe94544aac2.mp4

Fighter locking:

https://user-images.githubusercontent.com/9423435/207432930-a376d367-0e04-4795-9b15-4472dc111663.mp4

</details>

## Changelog
:cl:
refactor: Refactors how targeting is handled
add: DRADIS now has a targeting mode, cycled using the hailing/targeting button. Clicking on enemies in targeting mode will paint them, and clicking on friendlies will pass your current target lock to them via datalink
tweak: Tweaks how datalink functions. It is no longer automatic, and the sending ship must maintain a lock to maintain the datalink transmission. Targets recieved via datalink are also visible on DRADIS now.
tweak: Painted targets will now be lost if radar contact is not maintained.
add: Readds the gunner's optical camera. This works on large ships and some small ships with TAC consoles, and has a range similar to that of the ship's base sensor range.
balance: AI ships will now target incoming ordnance with PDCs, and will target ships locking them with flak.
balance: All overmap ships are now limited to 3 target paints, except for fighters, which are limited to one paint. Fighters will also automatically lock onto the target they paint.
fix: Fixes some null handling issues in target lists.
code: Homing is now based on the projectile being fired instead of the weapon.
tweak: Non-fighter ghost ships now have access to active radar.
/:cl:

### DIFF
--- a/nsv13/code/modules/munitions/ammunition/missiles/missile_types.dm
+++ b/nsv13/code/modules/munitions/ammunition/missiles/missile_types.dm
@@ -40,6 +40,7 @@
 /obj/item/projectile/guided_munition/missile/dud
 	icon_state = "torpedo_dud"
 	damage = 0
+	can_home = FALSE
 
 /obj/item/ship_weapon/ammunition/missile/georgio
 	name = "\improper Georgio"
--- a/nsv13/code/modules/munitions/ammunition/missiles/missile_types.dm
+++ b/nsv13/code/modules/munitions/ammunition/missiles/missile_types.dm
@@ -42,6 +42,9 @@
 	damage = 0
 	can_home = FALSE
 
+/obj/item/projectile/guided_munition/missile/dud/homing
+	can_home = TRUE
+
 /obj/item/ship_weapon/ammunition/missile/georgio
 	name = "\improper Georgio"
 
--- a/nsv13/code/modules/overmap/radar.dm
+++ b/nsv13/code/modules/overmap/radar.dm
@@ -39,15 +39,16 @@
 	var/radar_delay = MIN_RADAR_DELAY
 	// Whether we use DRADIS-assisted targeting.
 	var/dradis_targeting = FALSE
+	// Whether we can use the radar
+	var/can_use_radar = TRUE
 
 /obj/machinery/computer/ship/dradis/proc/can_radar_pulse()
+	if(!can_use_radar)
+		return FALSE
 	var/obj/structure/overmap/OM = get_overmap()
 	var/next_pulse = OM.last_radar_pulse + radar_delay
 	if(world.time >= next_pulse)
 		return TRUE
-
-/obj/machinery/computer/ship/dradis/internal/can_radar_pulse()
-	return FALSE
 
 /obj/machinery/computer/ship/dradis/internal/awacs/can_radar_pulse()
 	var/obj/structure/overmap/OM = loc
@@ -56,9 +57,6 @@
 	var/next_pulse = OM.last_radar_pulse + radar_delay
 	if(world.time >= next_pulse)
 		return TRUE
-
-/obj/machinery/computer/ship/dradis/minor/can_radar_pulse()
-	return FALSE
 
 
 /*
@@ -127,6 +125,7 @@ Called by add_sensor_profile_penalty if remove_in is used.
 
 /obj/machinery/computer/ship/dradis/minor //Secondary dradis consoles usable by people who arent on the bridge. All secondary dradis consoles should be a subtype of this
 	name = "air traffic control console"
+	can_use_radar = FALSE
 
 /obj/machinery/computer/ship/dradis/minor/cargo //Another dradis like air traffic control, links to cargo torpedo tubes and delivers freight
 	name = "\improper Cargo freight delivery console"
@@ -184,6 +183,7 @@ Called by add_sensor_profile_penalty if remove_in is used.
 	start_with_sound = FALSE
 	base_sensor_range = SENSOR_RANGE_FIGHTER
 	hail_range = 30
+	can_use_radar = FALSE
 
 /obj/machinery/computer/ship/dradis/internal/has_overmap()
 	return linked
@@ -433,6 +433,7 @@ Called by add_sensor_profile_penalty if remove_in is used.
 			send_radar_pulse()
 	else
 		data["can_radar_pulse"] = FALSE
+	data["can_use_radar"] = can_use_radar
 	return data
 
 /datum/asset/simple/overmap_flight
--- a/nsv13/code/modules/overmap/radar.dm
+++ b/nsv13/code/modules/overmap/radar.dm
@@ -18,6 +18,7 @@
 	var/show_asteroids = FALSE //Used so that mining can track what they're supposed to be drilling.
 	var/mining_sensor_tier = 1
 	var/last_ship_count = 0 //Plays a tone when ship count changes
+	var/last_missile_warning = 0 // Anti-spam for missile warning messages
 	//Alpha sliders to let you filter out info you don't want to see.
 	var/showFriendlies = 100
 	var/showEnemies= 100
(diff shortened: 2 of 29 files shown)


Karmic-Skink:
Heckin'

Karmic-Skink (changes requested):
Applying this stuff to the parent DRADIS device/UI is problematic.

One solution:
- Create flag parent dradis device to check for if this feature is enabled
- Set it to FALSE on parent
- Create a DRADIS subtype for WEAPONS active use
- Alternative: Perform a check that it is linked to a device that can even use a torp, and enable flag when present
- Only enable the TGUI element when is being called from a console with flag active
- ~~Cry when ghosts work out they can spam the lock target commands~~

Karmic-Skink:
> > Applying this stuff to the parent DRADIS device/UI is problematic.
> 
> I'm not entirely sure how this could be problematic, since ui_act already checks for whether a user is a pilot or gunner and rejects any attempts to change it if they aren't. Adding checks for a specific console could also cause issues if you somehow open another one as a gunner. I don't know if the latter's still something that happens, but I've heard of it occasionally being an issue a while back.

Because it is UI clutter that isn't relevant to all applications.

benbot16:
> Because it is UI clutter that isn't relevant to all applications.

Ah, thought you were talking about access to the button, not whether it's there for other people. Anyways, added a check to hide that.
Gunner/pilot:
image
Non-gunner/pilot:
image

If you want I can do the same for radar pulsing, since that'd reduce clutter even further.

Bokkiewokkie:
> If you want I can do the same for radar pulsing, since that'd reduce clutter even further.

Yes please, public dradis doesn't need anything other than the Re-focus and zoom buttons

benbot16:
> Yes please, public dradis doesn't need anything other than the Re-focus and zoom buttons

image

That good?

nsv13/code/modules/overmap/radar.dm
```diff
@@ -35,16 +37,19 @@
 	var/obj/item/supplypod_beacon/beacon
 	var/sensor_mode = SENSOR_MODE_PASSIVE
 	var/radar_delay = MIN_RADAR_DELAY
+	// Whether we use DRADIS-assisted targeting.
+	var/dradis_targeting = FALSE
+	// Whether we can use the radar
+	var/can_use_radar = TRUE
```
Bokkiewokkie, line 43: Me when the ship's gunner (the fool) used their dradis manually, setting it as the default for the ship and allowing me to ping using it through the viewscreens because the last touched console is set as the ship default. 
Or anyone else uses a different dradis consoles which removes the gunner's (the fool again, but for *not* using the console this time) ability to lock on and ping.

I know this is just a more general problem with overmap ships but I feel it's still worth mentioning possible problems. Could be 'fixed' by giving viewscreens ~~good code~~ an internal dradis and prioritizing that on use.

nsv13/code/modules/munitions/ammunition/missiles/missile_types.dm
```diff
@@ -40,6 +40,7 @@
 /obj/item/projectile/guided_munition/missile/dud
 	icon_state = "torpedo_dud"
 	damage = 0
+	can_home = FALSE
```
Bokkiewokkie, line 43: I think it would be funny if this updated when you've installed the guidance computer and propulsion system.

Bokkiewokkie (approved):

Karmic-Skink:
@Bokkiewokkie Bokkie, my review here is that it is messing with systems I'm reworking.

Up to you which order they are resolved.

nsv13/code/modules/overmap/radar.dm
benbot16: Pain. I'll see if I can work something out.

benbot16: Aight, did some stuff that should fix viewscreen pinging. It looks like there's already something to prevent minor consoles from becoming the "main" console, though.

I can try reworking it further if you want, but that'd probably be too out of scope for this PR.

nsv13/code/modules/munitions/ammunition/missiles/missile_types.dm
benbot16: Done.

Bokkiewokkie:
> @Bokkiewokkie, my review here is that it is messing with systems I'm reworking.
> 
> Up to you which order they are resolved.
> 
> For instance, I relocated hailing from the DRADIS UI and onto another.

I don't think these conflicts would be to terrible to resolve, and since this is fully ready and fixes other issues we'll have this first.

Bokkiewokkie (approved):

PR #2237, opened 2022-12-18
Medicine/Synthesizer Rebalance

BetaCentari:
## About The Pull Request
Reverts chemical synthesizers from the Beebase to function without matter cartridges
Returns Synthflesh, Styptic Powder, and Silver Sulfadiazine to their original function 

## Why It's Good For The Game
The game is incredibly fast paced, and makes medical treatment need to follow that pace. These changes to the medical system have hindered the speed of medical players and has been the cause of many very strongly voiced concerns. This will quiet these concerns. 

<details>

Don't have any, sorry. 

</details>

## Changelog
:cl:
tweak: Tweaked SF, SP, and SS behavior when applied through patches
balance: Reverts the chemical Synthesizers to function how they used to
/:cl:


Kenionatus:
I personally think the requirement for matter cartridges is neat, so chem factories aren't just set and forget, but that's just my personal opinion. It also encourages interaction with cargo (when med doesn't just walk/tide in and print them themselves).

Karmic-Skink:
>has been the cause of many very strongly voiced concerns

Can we please have some citations of this?

red031000:
Hello, I am one of the complainers, I personally think that both the SF changes and the matter cartridges are bad, and have severely put me off playing medical, where before I used to play it quite a lot.
a SF factory was an easy way to help heal people who had less need of medical at that time, while the doctors focus on the people who had died, and who were in crit. with the pace of NSV, constantly in and out of battles, and very easy to die, I feel that this is important, and these changes made it practically impossible to do this. I asked sophie to change this originally, because it's made playing medical pretty unbearable

TheScienceOfKSP:
1; I agree that needing matter cartridges for the chemfactory is a hassle, if only because getting the cartridges puts a strain on the logistics a ship that already devours materials, and the cartridges themselves are highly prized by engineering for mid and post-combat fixing.

I suggest perhaps making the factories need a different "fuel" that is easier for cargo to fetch (beef broth?) to lower the cases of Medi-Engi feuds.

2; The changes to SF, SP, and SS might be needed/useful on Bee, but on NSV people quite literally get nailed randomly by penetrating rounds, i support reverting the bee-nerf/change on them for NSV.

TL;DR
Magic Factory Bad, But fueled factory running on Engi-tonium worse.
Gib Healz back.

RemervKeberb:
While I'm not a fan of plumbing in the first place. On a lowpop server like ours that level of automation is nice to have because it gives you a tradeoff of setting up during downtime then moving to combat stations to help muni or retrieve the injured, and, as many bee changes. The matter cartridges simply don't add anything but annoyance to those trying to partake in it.

The material cost for them would probably be better off as a power cost just like a chem dispenser. Taking in watts to produce chems at the speed the capacitors allow it to. (Upgraded ones requiring hotwiring to the grid and an increased output from engi if the factory is large enough, for example.)

Even then though, that's a fair bit of effort for not a lot of reward codewise, so I'd advocate for reverting it and forgetting about it.


The patch changes are.. To say the least, a poor attempt to give more depth to the medical system. Combined with the changes to the basic healing chems such as bica and kelotane. It's just an entirely by-passable hurdle which limits treatment options, resulting in a more boring and linear treatment procedure. For actual medical depth chems shouldn't be touched, actual systems like broken bones and fractures from places like bay/paramed would be ideal. Still though, this is a point harping on Bee. For NSV, it doesn't matter. We get injured enough.

One final thing I'd like to note, patches affecting single limbs not only makes them less than ideal. It actually makes them useless because with the same time and effort to craft patches, you could've been crafting bruise packs and ointment which are the already ingame single limb targeting healing medicines.

bluewasabi24:
Perhaps if we had a way to source matter cartridges elsewhere, it could be helpful. Cargo has better things to do than print 20+ cartridges for medical in the first 30 min of the round. If medical can print them in their lathe, that would also save a lot of heartache. Or, if we can treat the dispensers like RCDs where we don't _need_ cartridges and are just able to shove any sort of material (iron sheets, rods, etc) into them in order for them to work, that could also work. 

Especially if you _don't_ have cargo on board and the CO is being pulled left, right, and sideways, medical will most likely have to resort to breaking into cargo, which no one wants, really. 

I also stand by patches applying toward all limbs, not just specific limbs. As mentioned before, we already have bandages and ointment that do the same thing. Patches are almost pointless if ointment and bandages will do it faster and cheaper.

Bokkiewokkie (changes requested):
Can you please remove the "gameplay too fast" comments from the changes? NSV gameplay speed doesn't contribute much to this, and you don't need to justify your changes in the code itself. Only the //NSV13 tags and what was actually changed should be included.

code/modules/plumbing/plumbers/synthesizer.dm
```diff
@@ -19,10 +19,9 @@
 	var/datum/reagent/reagent_id = null
 	///reagent overlay. its the colored pipe thingies. we track this because overlays.Cut() is bad
 	var/image/r_overlay
-	///The amount of reagent dispensable before requiring a refill from a compressed matter cartridge.
-	var/volume_left = 0
-	///The maximum amount of precursor in a synthesizer.
-	var/max_volume = 1000
+
+//NSV 13 Removed the two vars here for matter cartridge requirements. Need speedy heals
```
Karmic-Skink, line 23: You've still got this here.

Karmic-Skink (changes requested):
Please implement a better solution than just the blanket revert/removal of cartridges.

Some suggestions have been offered in comments already.

code/modules/reagents/chemistry/reagents/medicine_reagents.dm
```diff
@@ -254,15 +254,14 @@
 	metabolization_rate = 2.5 * REAGENTS_METABOLISM
 	overdose_threshold = 100
 
-/datum/reagent/medicine/silver_sulfadiazine/reaction_mob(mob/living/M, method=TOUCH, reac_volume, show_message = 1, touch_protection, obj/item/bodypart/affecting)
+/datum/reagent/medicine/silver_sulfadiazine/reaction_mob(mob/living/M, method=TOUCH, reac_volume, show_message = 1) //NSV 13 - reverted to old application
```
Karmic-Skink, line 257: Please note that all of these meds you've changed here are actually topical applications IRL and really do only affect the area they are applied to.

BetaCentari: I am aware, thank you.

code/modules/plumbing/plumbers/synthesizer.dm
BetaCentari: Removed 'need speedy heals' at the end of the comment.

BetaCentari:
> Please implement a better solution than just the blanket revert/removal of cartridges.
> 
> Some suggestions have been offered in comments already.

I've been asked to revert the changes, and that's what's been done for this PR. I don't intend on implementing a new system for plumbing at the moment.

Karmic-Skink:
> > Please implement a better solution than just the blanket revert/removal of cartridges.
> > Some suggestions have been offered in comments already.
> 
> I've been asked to revert the changes, and that's what's been done for this PR. I don't intend on implementing a new system for plumbing at the moment.

This implied narrative clearly isn't shared across the community.

Karmic-Skink:
The tl;dr on this is basically that this PR is addressing symptoms, not the root cause, ie TTK and lethality of ship weapons impacting ship-z and player avatars directly.

Kenionatus:
> "Perhaps if we had a way to source matter cartridges elsewhere"

If peak demand on cargo is indeed an issue then I'd say medical should start with a couple of cartridges to get it going. The quantity of those should probably be just enough make one bottle of oil in an oil factory and an amount of synthflesh patches that runs out after treating one or two people. I think we definitely should start with too few rather than too many because a lack is more likely to get rectified than an excess (because a lack is more noticeable).

> The tl;dr on this is basically that this PR is addressing symptoms, not the root cause, ie TTK and lethality of ship weapons impacting ship-z and player avatars directly.

Can I dl:dr that further to @BetaCentari says medical needs to be fast to keep up with high lethality and you say that lethality needs to come down to the speed of Bee medical? (You don't explicitly state that you think lethality needs to change, but I read that into "addressing symptoms, not root causes".)

Karmic-Skink:
> > "Perhaps if we had a way to source matter cartridges elsewhere"
> 
> If peak demand on cargo is indeed an issue then I'd say medical should start with a couple of cartridges to get it going. The quantity of those should probably be just enough make one bottle of oil in an oil factory and an amount of synthflesh patches that runs out after treating one or two people. I think we definitely should start with too few rather than too many because a lack is more likely to get rectified than an excess (because a lack is more noticeable).
> 
> > The tl;dr on this is basically that this PR is addressing symptoms, not the root cause, ie TTK and lethality of ship weapons impacting ship-z and player avatars directly.
> 
> Can I dl:dr that further to @BetaCentari says medical needs to be fast to keep up with high lethality and you say that lethality needs to come down to the speed of Bee medical? (You don't explicitly state that you think lethality needs to change, but I read that into "addressing symptoms, not root causes".)

The narrative thus far has been "NSV13 too fast", but this hasn't actually been elaborated on.
So at the moment, it just a thread being woven by a few people with little evidence.

NSV13 definitely takes longer to kill people from ship combat than FTL13 (#FTL13 Medical Main), where it was shields down everything exploding within seconds.

code/modules/plumbing/plumbers/synthesizer.dm
```diff
@@ -123,6 +95,8 @@
 	data["current_reagent"] = ckey(initial(reagent_id.name))
 	return data
 
+//NSV13 Removed need for matter cartridges, need speedy heals pls
+
```
Bokkiewokkie, line 99: Can you revert the changes to this file? I think the solution presented in Karmic's PRs regarding this is better and don't want them to conflict during testmerging.

Bokkiewokkie (changes requested):

Bokkiewokkie (approved):

PR #2243, opened 2022-12-26
Incredibly Lazy Chemical Synthesizer Rework

Karmic-Skink:
## About The Pull Request
This PR gives the synth a bespoke restocking item that is available from cargo.

## Why It's Good For The Game
Apparently people fight over RCD cartridges.

<details>

image

</details>

## Changelog
:cl:
add: Added chem synth refill cartridge
balance: Reworked chem synth to use its own bespoke cartridge from cargo
/:cl:

### DIFF
--- a/code/modules/plumbing/plumbers/synthesizer.dm
+++ b/code/modules/plumbing/plumbers/synthesizer.dm
@@ -71,24 +71,24 @@
 	volume_left = max(volume_left - amount*delta_time*0.5, 0)
 
 /obj/machinery/plumbing/synthesizer/attackby(obj/item/O, mob/user, params)
-	if(!istype(O, /obj/item/rcd_ammo))
+	if(!istype(O, /obj/item/synthesizer_cartridge))
 		return ..()
-	var/obj/item/rcd_ammo/R = O
-	if(!R.ammoamt)
+	var/obj/item/synthesizer_cartridge/R = O
+	if(!R.chems)
 		to_chat(user, "<span class='warning'>The [R.name] doesn't have any reagent left!</span>")
 		return ..()
 	var/added_volume = -volume_left //For the difference calculation
-	volume_left = min(volume_left+R.ammoamt*10, src.max_volume) //400 per cartridge
+	volume_left = min(volume_left+R.chems*10, src.max_volume) //400 per cartridge
 	added_volume = added_volume+volume_left
-	R.ammoamt -= added_volume/10
-	if(R.ammoamt <= 0) //Emptied
+	R.chems -= added_volume/10
+	if(R.chems <= 0) //Emptied
 		to_chat(user, "<span class='notice'>You refill the chemical synthesizer with the [R.name], emptying it completely!</span>")
 		qdel(R)
 		return
 	if(added_volume == 0) //No change
 		to_chat(user, "<span class='notice'>The chemical synthesizer is full!</span>")
 		return
-	to_chat(user, "<span class='notice'>You refill the chemical synthesizer with the [R.name], leaving [R.ammoamt*10] units in it.</span>")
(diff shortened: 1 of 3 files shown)


Pontenerd:
I fail to see how this PR actually reponds to any of the concerns raised in #2237.
image
This is the closest suggestion, however I do not see how this hits on the "easier for cargo" part, as it has turned what was a draw on materials into a (much more severe) draw on the budget directly. Furthermore, I fail to see how this actually makes any aspect of the game more fun to play, as it stands, the community have mostly been against the cartrage change for plumbing, as it just adds more tedium to the game

Karmic-Skink:
> I fail to see how this PR actually reponds to any of the concerns raised in #2237. image This is the closest suggestion, however I do not see how this hits on the "easier for cargo" part, as it has turned what was a draw on materials into a (much more severe) draw on the budget directly. Furthermore, I fail to see how this actually makes any aspect of the game more fun to play, as it stands, the community have mostly been against the cartrage change for plumbing, as it just adds more tedium to the game

Upstream Issue

red031000:
not an upstream issue, as upstream has a different gameplay cycle, therefore having compressed matter works for upstream, NSV is different in this regard

covertcorvid:
If we're going this route, a design disk (we could spawn it at trader stations!) would be nice.

Karmic-Skink:
> not an upstream issue, as upstream has a different gameplay cycle, therefore having compressed matter works for upstream, NSV is different in this regard

99% the same.

red031000:
well I would argue this is the 1% then

having the fuel available from cargo and cargo only, at 120 for 1000 will be a huge drain on the budget for any projects med has (and well, SF being the prime example, that's a massive drain, likely will require thousands of chems)

the thing is, in NSV, we already have a massive drain on cargo budget: muni

muni requires constant restocking, and will be competing with medical for the cargo budget to get what they need, and well, muni will pretty much always win, making this change potentially even worse than using compressed matter

Karmic-Skink:
> well I would argue this is the 1% then
> 
> having the fuel available from cargo and cargo only, at 120 for 1000 will be a huge drain on the budget for any projects med has (and well, SF being the prime example, that's a massive drain, likely will require thousands of chems)
> 
> the thing is, in NSV, we already have a massive drain on cargo budget: muni
> 
> muni requires constant restocking, and will be competing with medical for the cargo budget to get what they need, and well, muni will pretty much always win, making this change potentially even worse than using compressed matter

No, you don't argue the 1%, you actually elaborate on the details of it. This isn't a fight unless you want it to be. 

As indicated by the title and the clock, I spent very little time on the actual numbers here, so yes, they would be subject to change, hence make a case about that.

Munitions competing for cargo budget with Medical is an entirely separate issue, and it is extremely foolhardy to roll it into this.

By definition mechanically we have systems for separate budgets that are active. And this is a game about active decisions and teamwork.
 - If people are unaware of the budget system, then it may be time to spread the word
 - If cargo is using the general discretionary budget on munitions, then that is their decision
 - If the bridge bagels are preferring to use sheer volume of fire rather than being economical, then that is their decision
 - If munitions aren't using their own budget to purchase munitions, that is their decision
 - If medical is choosing to not use their own budget for medical supplies, then that is their decision
 - If the budgets are insufficient for the particular departments, then that can be adjusted independently (PS, one of the things I do while I'm sitting on the server is actually monitoring these budgets and seeing how they are used, you'd be surprised how often the department budgets are untouched and have hundreds of thousands of credits in them [Side note x3: I have been considering reworking the budgets, including cargo console UI/UX, for a long while now])
 - If admemes decide to do the funny because someone has left all the budget cards for every department in one place, that is their decision
 
 But anyway, this is distracting me from Tomcat Tuesday aka Fighters 4.0

covertcorvid:
A different rework got merged

PR #2245, opened 2022-12-28
Synths Retro Redux

Karmic-Skink:
## About The Pull Request
Inside you there are two wolves.
- One that wants the game to be easier in the name of Fun.
- The other wants the game to be harder in the name of Fun.

## Why It's Good For The Game
Adds an optional active maintenance step to synthesizers, that can be ignored for a reduced rate.
Please try it before screaming about balance numbers, tia.

<details>

</details>

## Changelog
:cl:
balance: Doubles synthesizer capacity and restock rate for same resources
balance: Synthesizers now tick at 1/4th speed if not stocked with resources
/:cl:

### DIFF
--- a/code/modules/plumbing/plumbers/synthesizer.dm
+++ b/code/modules/plumbing/plumbers/synthesizer.dm
@@ -22,7 +22,9 @@
 	///The amount of reagent dispensable before requiring a refill from a compressed matter cartridge.
 	var/volume_left = 0
 	///The maximum amount of precursor in a synthesizer.
-	var/max_volume = 1000
+	var/max_volume = 2000 //NSV13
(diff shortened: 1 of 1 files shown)


red031000:
this actually seems fine to me, I would be happy if this was the case, I brought up something with corvid about splitting the output when two outputs are connected, it's not strictly necessary so can be implemented later if not possible now, but I think it would help with the slower rate, as both outputs are fed with chems

Bokkiewokkie (approved):

covertcorvid:
interestingly, plumbing seems to operate on a query/response system instead of publish/subscribe, which explains why it fills one at a time

PR #2320, opened 2023-02-20
Ethereal Lightbulb

SerynEngi:
## About The Pull Request

This PR doubles ethereal light radius from a radius of 1, to a radius of 2, so instead of it being a highlighter of where the ethereal is, it can actually be used by the ethereal.

## Why It's Good For The Game

Ethereals are lightbulbs, not glowsticks. Provides actually semi-useful light radius from their only unique species trait.

## Changelog
:cl:
tweak: Doubled Ethereal light radius.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/species_types/ethereal.dm
+++ b/code/modules/mob/living/carbon/human/species_types/ethereal.dm
@@ -96,7 +96,7 @@
 		var/healthpercent = max(H.health, 0) / 100
 		if(!emageffect)
 			current_color = rgb(r2 + ((r1-r2)*healthpercent), g2 + ((g1-g2)*healthpercent), b2 + ((b1-b2)*healthpercent))
-		ethereal_light.set_light_range_power_color(2 + (3 * healthpercent), 2 + (2 * healthpercent), current_color) //NSV13- Its not a performance issue here, and we're theoretically balanced for PVE.
+		ethereal_light.set_light_range_power_color(2 + (3 * healthpercent), 2 + (2 * healthpercent), current_color) //NSV13- Makes the base brightness a radius of two, and reduces the rate of brightness decline from damage.
 		ethereal_light.set_light_on(TRUE)
 		fixed_mut_color = copytext_char(current_color, 2)
 	else
--- a/code/modules/mob/living/carbon/human/species_types/ethereal.dm
+++ b/code/modules/mob/living/carbon/human/species_types/ethereal.dm
@@ -96,7 +96,7 @@
 		var/healthpercent = max(H.health, 0) / 100
 		if(!emageffect)
 			current_color = rgb(r2 + ((r1-r2)*healthpercent), g2 + ((g1-g2)*healthpercent), b2 + ((b1-b2)*healthpercent))
-		ethereal_light.set_light_range_power_color(2 + (3 * healthpercent), 2 + (2 * healthpercent), current_color)
+		ethereal_light.set_light_range_power_color(2 + (3 * healthpercent), 2 + (2 * healthpercent), current_color) //NSV13- Its not a performance issue here, and we're theoretically balanced for PVE.
 		ethereal_light.set_light_on(TRUE)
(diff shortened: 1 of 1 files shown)


Karmic-Skink (changes requested):
This requires an //NSV13 tag due to being a core file

SerynEngi:
> This requires an //NSV13 tag due to being a core file.

Mission accomplished.

code/modules/mob/living/carbon/human/species_types/ethereal.dm
```diff
@@ -96,7 +96,7 @@
 		var/healthpercent = max(H.health, 0) / 100
 		if(!emageffect)
 			current_color = rgb(r2 + ((r1-r2)*healthpercent), g2 + ((g1-g2)*healthpercent), b2 + ((b1-b2)*healthpercent))
-		ethereal_light.set_light_range_power_color(1 + (2 * healthpercent), 1 + (1 * healthpercent), current_color)
+		ethereal_light.set_light_range_power_color(2 + (3 * healthpercent), 2 + (2 * healthpercent), current_color) //NSV13- Its not a performance issue here, and we're theoretically balanced for PVE.
```
Karmic-Skink, line 99: Remove the commentary.

Karmic-Skink (changes requested):
We don't need the justification as a comment here, I just asked for an //NSV13 tag

Altangy:
Karmic, it's just a comment and it fulfills the requirement you set out. It should be fine

code/modules/mob/living/carbon/human/species_types/ethereal.dm
SerynEngi: The commentary is so that when some inevitable bee-base change comes along, whomever does the change can look and see why it was increased. 

Namely that we do not have (and to my knowledge, never have had) sufficient ethereal players to where this is a performance issue, and we're balanced for PVE so its not a balance issue.

I do not understand the need to remove a comment providing said detail. 

Does it make the code run less efficiently? Should we remove other comments that provide details? 

I found a literal poem commented out while I was hunting through the wrong place to find the variables I was looking to modify. Should I find that again and delete it?

I will change it if there is good reason to do so. If you still don't like the fact that I added said detail but do not have a logical, code-based reason to do so, close the PR.

Karmic-Skink: Yes, and the tl;dr above belongs in the PR "Why It's Good For The Game" section which gets hooked and referenced through plugins like GitLense so we can all see exactly which PR made changes and why.

However if you want to make it a fight, please do battle with the close PR button yourself?

covertcorvid: Neither of these is a good reason for closing a PR so let's all quit daring each other to do it, okay?

What would be more helpful to me when beebasing if there's going to be a comment is describing the function of the change (e.g. "made ethereals brighter") in case the block of code gets moved, but we've got plenty of uncommented tags lying around too.

SerynEngi: Comment changed to desired descriptive details. Let me know if you want it to be more succinct, corvid.

covertcorvid: Can't see it yet, make sure you push

Bokkiewokkie (approved):
Looks good I was worried darker colors of ethereal wouldn't work well with this but the light brightness increase made it better than it was.
I also recommend you port this to beestation, to potentially avoid having more conflicts to resolve in a beebase.

covertcorvid (approved):

SerynEngi:
> Looks good I was worried darker colors of ethereal wouldn't work well with this but the light brightness increase made it better than it was. I also recommend you port this to beestation, to potentially avoid having more conflicts to resolve in a beebase.

They were the ones who reduced it in the first place, so I don't think that would go over well.

PR #2323, opened 2023-02-24
Fixes ship brakes

benbot16:
## About The Pull Request

Fixes a bug introduced in #2315 that broke ship brake keybinds. It wasn't testmerged so I have to do it in a separate PR.

As it turns out, a change I made to the vehicle toggle_brakes didn't get commited properly, so overmap/toggle_brakes and vehicle/toggle_brakes ended up with the same name keys, and vehicle/toggle_brakes overrode overmap/toggle_brakes.

It has a different key now, so it should work.

## Why It's Good For The Game

Bugs are bad.

<details>

image

image

</details>

## Changelog
:cl:
fix: Fixes the overmap brakes hotkey.
/:cl:

### DIFF
--- a/nsv13/code/datums/keybinding/vehicle.dm
+++ b/nsv13/code/datums/keybinding/vehicle.dm
@@ -7,7 +7,7 @@
 // Toggling brakes for tugs/similar vehicles
 /datum/keybinding/vehicle/toggle_brakes
 	key = "Alt"
-	name = "toggle_brakes"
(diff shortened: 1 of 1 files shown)


Bokkiewokkie (approved):

benbot16:
> there should probably be a check for key clashing, but i'm not sure it wasn't intended this way

The keybind system supports having the same key bound to multiple things using the priority system. That's why we can have WASD keybinds for ships.

PR #2327, opened 2023-03-01
Shrike adjustments

DuskLight9978:
## About The Pull Request
The PR Allows the shrike to construct machines that originally could not be constructed, The cargo chems remain as an option to continue the challenge but if they wish to exit the challenge they can now construct the dispensers, silo, and Ore redemption machine. None are present round start but the boards are now in the tech storage.

## Why It's Good For The Game

Brings the shrike more inline with expectations, a ship that has its gimicks and challenges but allows one to upgrade it by rebuilding desired machines.

<details>

</details>

## Changelog
:cl:Dusklight
tweak: Shrike can now build chemdispensers, ORM, And Silo, but do not come round start.
/:cl:


Bokkiewokkie (changes requested):
The timing of this PR isn't the best. Karmic is currently working on making a new chemistry machine to replace the bottles on the Shrike.
On top of that, I don't see the problem with the removal of the ORM and Ore Silo, besides the fact they normally have unlimited storage capacity. The Shrike already has a requisitions type of cargo (with just a protolathe for all your printing needs), which removes the need to use the remote linking.
I wouldn't be against you making a type of ore silo which acts as... a silo, being just a big storage machine to avoid having to leave your ores in crates (or the smelter) for optimal storage. But as it is now, this PR will be denied.

red031000:
I would like to add that most of the player base genuinely dislikes/outright hates shrike, and will wait for the next round instead of playing, this has the effect of killing pop if it is chosen, badly

red031000:
I would personally recommend that if you would rather wait for the rework than fix the issues now, shrike be removed from rotation until the rework is done

Bokkiewokkie (commented):
I feel like my original points about the ore silo and such still apply, we have another PR which increases the protolathe and improves the ore furnace open right now which applies basically everything I talked about

PR #2334, opened 2023-03-06
Refillable chem dispenser (clean edititon)

covertcorvid:
## About The Pull Request
Adds a machine that stores separated chemicals loaded by the user. Basic chems are always displayed (greyed out if none are available), but custom chems can be loaded as well, which allows for precision mixing to reduce waste.
In addition, these machines do not require the presence of a beaker to create recipes, and will try to determine whether a recipe is able to be completed before processing requests - if completion is not possible, the user will be prompted for confirmation before proceeding.

## Why It's Good For The Game
Part of the promised rework

<details>

(https://streamable.com/fyuck1)

</details>

## Changelog
:cl: covertcorvid, Karmic-Skink
add: Added a non-power-based variant of the chem dispenser and out it on the Shrike
/:cl:


red031000:
will basic chems always be available?

covertcorvid:
> will basic chems always be available? somewhat difficult to tell

No, the buttons will always be present but they will be greyed out if there is none of that chemical in the storage.

red031000:
also I genuinely don't like that you are hard blocked from building certain machines on the shrike, if you manage to get the board for it you should be able to build it, a better method would be to make the board harder to get

red031000:
> > will basic chems always be available? somewhat difficult to tell
> 
> No, the buttons will always be present but they will be greyed out if there is none of that chemical in the storage.

how will you get more of the basic chems then?

covertcorvid:
> also I genuinely don't like that you are hard blocked from building certain machines on the shrike, if you manage to get the board for it you should be able to build it, a better method would be to make the board harder to get

I understand and acknowledge that you don't like this. The ability to acquire things not normally available to you may become available at a later date but would require an overhaul of some of the larger systems to do it the way we want.

> how will you get more of the basic chems then?
By purchasing their crates through cargo.

red031000:
so there would be one crate per chem? or multiple chems in each crate? current that's not implemented now

covertcorvid:
> so there would be one crate per chem? or multiple chems in each crate? current that's not implemented now

There are multiple chems in each crate and this is already implemented on the Shrike. We can add the crates to the other maps if you like but I don't think anyone would buy them without Shrike's constraints.

red031000:
thanks for explaining

re: the hard block from building, if I wanted to, for some reason, to spawn a circuit board as an admin, I would be incredibly frustrated to find that it wouldn't work on the shrike, admittedly I could spawn the machine directly but, yeah

covertcorvid:
> thanks for explaining
> 
> re: the hard block from building, if I wanted to, for some reason, to spawn a circuit board as an admin, I would be incredibly frustrated to find that it wouldn't work on the shrike, admittedly I could spawn the machine directly but, yeah

Happy to clarify things.

I have no sane way to made admin-spawned items different from the ones available to players without introducing special subtypes, and honestly that defeats the point. We are trying to advertise more clearly that these machines can't be built in order to get players to use the cargo system (having cargo play a larger role in supply requisitions was in fact also a request from a player) so that this doesn't surprise people, but I will not be removing the block as part of this PR.

Bokkiewokkie (dismissed):
Mapping approved

Bokkiewokkie:
> thanks for explaining
> 
> re: the hard block from building, if I wanted to, for some reason, to spawn a circuit board as an admin, I would be incredibly frustrated to find that it wouldn't work on the shrike, admittedly I could spawn the machine directly but, yeah

It is actually still possible and fairly easy to override the block on the machine building.
You can spawn the circuitboard, which will have its build path set to `null` upon initializing, but changing it back to the original build path will make it possible to build again just fine.

Bokkiewokkie (approved):

PR #2340, opened 2023-03-09
Hammer the Hammerhead Hangar

SerynEngi:
## About The Pull Request

This PR hammers the hammerhead hangar into a functional state that doesn't require the ATC and actually lets pilots get fighters out.

## Why It's Good For The Game

Fighters are the only heavy armament the hammerhead has left, so we probably should make it so that they can take off without disassembling half of the hangar and all of the firelocks.

Works on my machine. There is actually a screenshot there, but I forgot to take a before. Its better, I promise.

https://cdn.discordapp.com/attachments/933214711112163380/1083216334504083496/Unfucked_Hammerhead_Hangar.PNG

## Changelog
:cl:
Fix: Hammerhead fighter launch tubes now functional at round-start due to the addition of blast-door buttons and removal of extra blast-doors, heavy firelocks and unused atmospherics piping.
/:cl:

### DIFF
--- a/_maps/map_files/Hammerhead/Hammerhead.dmm
+++ b/_maps/map_files/Hammerhead/Hammerhead.dmm
@@ -268,7 +268,6 @@
 	name = "Launch tubes 1 and 2"
 	})
 "aaY" = (
-/obj/machinery/atmospherics/pipe/simple/general/visible,
 /turf/open/floor/plating,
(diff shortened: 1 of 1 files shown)


Karmic-Skink (dismissed):
Update the changelog to be descriptive of the work done and the changes made.

PR #2364, opened 2023-03-26
Makes catwalks behave like floors for mouse-related purposes

covertcorvid:
## About The Pull Request
Prevents turfs with a wire and a catwalk from being considered for mouse migration. Prevents mice from chewing wires that are covered by catwalks.

## Why It's Good For The Game
These are supposed to be better than exposed wires

<details>

</details>

## Changelog
:cl:
tweak: Mice will no longer spawn or eat on catwalk wires
/:cl:


Bokkiewokkie (approved):
This technically stops mice from eating wires on catwalks in space as well but I think the mouse has bigger problems at that point already

covertcorvid:
> This technically stops mice from eating wires on catwalks in space as well but I think the mouse has bigger problems at that point already

Would it be better if I only look for the NSV kind?

PR #2311, opened 2023-02-08
[Upstream] UPDATES, UPDATES EVERYWHERE

Bobbanz1:
## About The Pull Request
Pulls the following PRs from Upstream:
* https://github.com/BeeStation/BeeStation-Hornet/pull/8414
* https://github.com/BeeStation/BeeStation-Hornet/pull/8406
* https://github.com/BeeStation/BeeStation-Hornet/pull/8304
* https://github.com/BeeStation/BeeStation-Hornet/pull/8398
* https://github.com/BeeStation/BeeStation-Hornet/pull/8358
* https://github.com/BeeStation/BeeStation-Hornet/pull/8278
* https://github.com/BeeStation/BeeStation-Hornet/pull/8437
* https://github.com/BeeStation/BeeStation-Hornet/pull/8436
* https://github.com/BeeStation/BeeStation-Hornet/pull/8439
* https://github.com/BeeStation/BeeStation-Hornet/pull/8438
* https://github.com/BeeStation/BeeStation-Hornet/pull/8444
* https://github.com/BeeStation/BeeStation-Hornet/pull/8440
* https://github.com/BeeStation/BeeStation-Hornet/pull/8405

## Why It's Good For The Game
UPDATES
UPDATES EVERYWHERE!!

## Testing Photographs and Procedure

<details>
<summary>Screenshots&Videos</summary>

nedladdning
</details>

## Changelog
:cl:EvilDragonfiend, Ivniinvi, Zytolg, RKz, CydiaLamiales, Shiraizawa, itsmeow
admin: welding/unwelding an airlock will be logged
code: changed a long var declaration from a single line into multiple lines for better readability
code: timers are put in the closest future bucket, as a precaution
tweak: medibots have more skins based on which first aid kit was used when it's made. These currently have no speicla ability by its shape which means these are just skins.
imageadd: more medibot skins (advanced, rad, syndi, doctors roundstart kit)
tweak: Phobias can now be weighted
tweak: The phobias of Robots and Doctors are slightly less likely to occur
tweak: The phobia of Space is significantly less likely to occur
add: Lizard speech now has a variation on the use of X in normal conversation. Things like "Xenomorph" will sound the same as before as it is phonetically similiar, however uses of "x-ray" will become "ecks-ray".
add: Added a line of text showing the Security Level in the current round
fix: fixed anomalies spawning inside walls
tweak: Instead of being dusted, plasmaman helmets now have their lamps broken when hit with the Light Eater
[...]

PR #2375, opened 2023-04-01
Turns the expanded lizard hiss regex from #2311 into a character pref

DeltaFire15:
## About The Pull Request
Title. Now you have a character-pref level option to toggle between s-only hissing (Legacy) and x- and s hissing (Expanded)
Defaults to the expanded version.
This also applies retroactively if you happen to turn into a lizard during a round for whatever reason.

Guh, DB prefs are scary, I sure hope this works and doesn't break anything!

## Why It's Good For The Game
Options good, especially when people tend to have different ideas of how their lizard's speech may show itself.

No.

## Changelog
:cl:
add: Lizard hiss is now handled by a character preference, permitting choice between and new- and old styles.
/:cl:

### DIFF
--- a/SQL/beestation_schema.sql
+++ b/SQL/beestation_schema.sql
@@ -117,7 +117,7 @@ CREATE TABLE IF NOT EXISTS `SS13_characters` (
 	`preferred_squad` VARCHAR(32) NOT NULL COLLATE 'utf8mb4_general_ci',
 	`preferred_pilot_role` VARCHAR(32) NOT NULL COLLATE 'utf8mb4_general_ci',
 	`flavor_text` MEDIUMTEXT NOT NULL COLLATE 'utf8mb4_general_ci',
-  `lizard_hiss_style` VARCHAR(32) NOT NULL COLLATE 'utf8mb4_general_ci',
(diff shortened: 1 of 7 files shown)


covertcorvid (approved):

Crossedfall:
You should probably add an entry into the database changelog.

https://github.com/BeeStation/NSV13/blob/master/SQL/database_changelog.txt

DeltaFire15:
> You should probably add an entry into the database changelog.
> 
> https://github.com/BeeStation/NSV13/blob/master/SQL/database_changelog.txt

Changelog talks about 5.9, has entries up to including 6.0 and the current NSV version before my changes is 6.1. Seems to lack attention in general.

Bokkiewokkie:
> > You should probably add an entry into the database changelog.
> > https://github.com/BeeStation/NSV13/blob/master/SQL/database_changelog.txt
> 
> Changelog talks about 5.9, has entries up to including 6.0 and the current NSV version before my changes is 6.1. Seems to lack attention in general.

Did not know there was a changelog, definitely should be updated yeah.

V6.1 added the flavor text field to the ss13_characters table if you want to include that in there

Bokkiewokkie (approved):
We can do the changelog on the other pref PR

PR #2383, opened 2023-04-07
No EMP for Lord Singuloth

SerynEngi:
## About The Pull Request

This PR does one thing, removes the EMP proc for the Singularity.

## Why It's Good For The Game

This enables Lord Singuloth as a viable powersource within hull on a ship. While it is not a default power source on the map, it should still be allowed to be viable without EMPing every room around it, making it as viable as the Tesla as a backup power-supply.

## Changelog
:cl:
del: Removed EMP effects from the Black Hole singularity.
/:cl:


Karmic-Skink (dismissed):
As this is a core file, it requires //NSV13 tags

SerynEngi:
> As this is a core file, it requires //NSV13 tags

Fair point, and I changed it to commenting out the EMP as opposed to a deletion as it makes a little more sense to do that, and I probably should have done it that way in the first place.

code/modules/power/singularity/singularity.dm
```diff
@@ -359,9 +359,9 @@
 						"<span class='userdanger'>You look directly into the [src.name] and feel weak.</span>")
 	return
 
-
-/obj/anomaly/singularity/proc/emp_area()
-	empulse(src, 8, 10)
+//NSV13
+//obj/anomaly/singularity/proc/emp_area()  //NSV13
+//	empulse(src, 8, 10)  //NSV13
```
Bokkiewokkie, line 364: Commenting this out isn't needed, it's only the actual implementation that's important.

SerynEngi: So its okay to leave a proc just in the code with no trigger? 
I was under the impression that such was bad, or at least messy.
I modified/commented out the thing that would trigger the emp_area() proc, so I also commented out the proc itself.

Bokkiewokkie: Normally it's messy, but this isn't new code, and since the functionality has not changed it's worth keeping up to date in case we want to put it back later. It's just not worth adding another merge conflict to the pile.
This also allows admins to still trigger it manually if they want to.

Bokkiewokkie (approved):

PR #2403, opened 2023-04-22
Fixes mouse aim not working while dragging the mouse

benbot16:
## About The Pull Request

This PR makes it so that a pilot dragging their mouse with mouse move enabled will update the ship's target angle. This also moves some code around in onMouseDrag to hopefully make it a tiny bit more performant. 

## Why It's Good For The Game

This makes it more practical to use autofire as a pilot instead of spam-clicking.

<details>

https://user-images.githubusercontent.com/9423435/233801988-f84befab-4352-4d8d-8fd4-a5c9f6a4cce1.mp4

</details>

## Changelog
:cl:
fix: Ship mouse aim now works while you're dragging the mouse.
/:cl:

### DIFF
--- a/code/_onclick/drag_drop.dm
+++ b/code/_onclick/drag_drop.dm
@@ -74,7 +74,7 @@
 	var/obj/item/H = get_active_held_item()
 	if(H)
 		. = H.canItemMouseDown(object, location, params)
-	else if(src.overmap_ship && (src.overmap_ship.gunner == src || src.overmap_ship.pilot == src) || (GetComponent(/datum/component/overmap_gunning))) //NSV13 - let us mouse-down if we're a ship operator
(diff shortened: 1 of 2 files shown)


Kenionatus:
I'm not 100 percent sure, but it looks like the currently testmerged version has the issue that it uses the helm's mouse clicks to fire tac guns. Maybe it's only right after you switch to the weapon or any time afterwards, not sure.

Bokkiewokkie:
> That should be fixed in the last commit, I added some extra checks to mousedown and mouseup.

Can confirm that this is still an issue, when the helm clicks it fires the tactical console's selected weapons as well

breslichb:
> Can confirm that this is still an issue, when the helm clicks it fires the tactical console's selected weapons as well

in that case it's probably also an issue with other TMs, then. I can't replicate that on local.

Bokkiewokkie (approved):

PR #2405, opened 2023-04-23
Adds safety checks for things that reference the global overmap list

covertcorvid:
## About The Pull Request
How we keep getting nulls in there I don't know, but this will possibly make them mess things up less

## Why It's Good For The Game
Bugs bad

<details>

</details>

## Changelog
:cl:
code: Added checks for nulls to every place we use the global overmap list
/:cl:

### DIFF
--- a/code/game/machinery/porta_turret/portable_turret.dm
+++ b/code/game/machinery/porta_turret/portable_turret.dm
@@ -458,7 +458,7 @@
 		//Nsv13 start
 		for(var/A in GLOB.overmap_objects) //Has to go through global list due to occuring on a ship's z, not the overmap.
 			var/obj/structure/overmap/target = A
-			if((get_dist(A, base) < scan_range) && can_see(base, A, scan_range) && istype(A, /obj/structure/overmap) && target.z == z)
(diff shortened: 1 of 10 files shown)


DeltaFire15:
Something something performance though..
There is only exactly two things that even do ops on the list (one add and one remove) so it should in theory not be too difficult to look for the root instead.

covertcorvid:
> Something something performance though.. There is only exactly two things that even do ops on the list (one add and one remove) so it should in theory not be too difficult to look for the root instead.

You would think that it wouldn't be too difficult to find the problem, but since we already check to make sure we're not adding nulls to the list and everything is supposed to remove itself from the list during its Destroy (), you would be wrong

Bokkiewokkie (approved):
Surely this will be enough
It does feel a bit excessive to both have qdeleted checks and listclearnulls in there, but there's not too many jump_end calls so it's probably fine

covertcorvid:
> Surely this will be enough It does feel a bit excessive to both have qdeleted checks and listclearnulls in there, but there's not too many jump_end calls so it's probably fine

Actually I should take the listclearnulls out so that we can see if DeltaFire's fix works!

DeltaFire15:
> Actually I should take the listclearnulls out so that we can see if DeltaFire's fix works!

Just unTM this PR, handling add / del properly should resolve it entirely and if it shows up again, can just swap around testmerges.

PR #2406, opened 2023-04-27
Moves overmap object list add to Init instead of LateInit

DeltaFire15:
## About The Pull Request
Alternate PR to #2405
My theory for the FTL issues is that it bricks when an overmap object is created and immediately deleted (e.g. a Carrier creating a fighter on a bullet's tile) since the list add is only done on LateInit, so this moves it to the earliest possible point in Initialize() itself instead.
Hopefully should tackle the FTL bricking if this is indeed the problem, and not wacky things going on in Destroy().

## Why It's Good For The Game
Fix man good.

No.

## Changelog
:cl:
fix: Hopefully resolves some FTL issues by moving the list add somewhere more sane.
/:cl:


covertcorvid:
that makes sense, I hope this is the problem

covertcorvid (approved):

Bokkiewokkie (approved):

PR #2408, opened 2023-04-28
Buffs fighter ghost ships + Increases ghost ship probability

someone543:
## About The Pull Request
Any time a fighter is selected as a ghost ship, its stats are heavily buffed to be similar to a standard light fighter (Speed, armor, hull, ammo)
Also increases probability of a ghost ship from 10% to 20%

## Why It's Good For The Game
Fighter PVP is pretty fun, so this PR attempts to replicate that experience with ghost ship fighters.
Also, dying in 5-6 PDC shots and only having 15 cannon ammo isn't fun.
Probability is increased because for a ghost role that is extremely outmatched by the NSV and usually dies within a minute, 10% is a bit too rare

<details>

TBA

</details>

## Changelog
:cl:
tweak: Ghost ship probability increased to 20% from 10%
balance: Ghost ship fighters buffed
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/overmap_ghosts.dm
+++ b/nsv13/code/modules/overmap/overmap_ghosts.dm
@@ -52,18 +52,8 @@
 	if(gunner)
 		QDEL_NULL(gunner)
 
-	//Buffs fighters so dogfights are more fun. 
-	if(mass == MASS_TINY)
-		max_armour = 200
-		obj_integrity = 200
-		max_integrity = 200 //About as squishy, and fast, as a light fighter
-		forward_maxthrust = 3.5
-		backward_maxthrust = 3.5
-		side_maxthrust = 4
-		integrity_failure = 200
-		max_angular_acceleration = 200
-		speed_limit = 10
-		shots_left = 500
+	//Buff the ships
+	spec_ghostship_changes()
 
 	//Insert the extra machines
 	if(!dradis)
@@ -111,3 +101,6 @@
 
 	else //Try again later
 		addtimer(CALLBACK(src, PROC_REF(ghost_key_check), ghost), 1 SECONDS)
+
+/obj/structure/overmap/proc/spec_ghostship_changes() //Proc to buff ghost ships. Must be overridden to apply buffs
+	return
--- a/nsv13/code/modules/overmap/overmap_ghosts.dm
+++ b/nsv13/code/modules/overmap/overmap_ghosts.dm
@@ -53,7 +53,8 @@
 		QDEL_NULL(gunner)
 
 	//Buff the ships
-	spec_ghostship_changes()
+	if(src.spec_ghostship == TRUE)
+		spec_ghostship_changes()
 
 	//Insert the extra machines
 	if(!dradis)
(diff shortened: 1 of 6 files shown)


nsv13/code/modules/overmap/overmap_ghosts.dm
```diff
@@ -52,6 +52,19 @@
 	if(gunner)
 		QDEL_NULL(gunner)
 
+	//Buffs fighters so dogfights are more fun.
```
Karmic-Skink, line 55: This needs to be modular based on the type of fighter, rather than hardcoded.

Karmic-Skink (changes requested):

nsv13/code/modules/overmap/overmap_ghosts.dm
DeltaFire15: Agree with karmic, try doing something like adding a spec_ghostship_changes() proc ../overmap that gets called when a ghost ship is created and have it called in place of your code, then have fighters override the proc to do what you want them to change.

someone543: Done

nsv13/code/modules/overmap/overmap_ghosts.dm
```diff
@@ -52,6 +52,9 @@
 	if(gunner)
 		QDEL_NULL(gunner)
 
+	//Buff the ships
+	spec_ghostship_changes()
```
Karmic-Skink, line 56: There is no if here so it will always be called.

nsv13/code/modules/overmap/overmap_ghosts.dm
```diff
@@ -98,3 +101,6 @@
 
 	else //Try again later
 		addtimer(CALLBACK(src, PROC_REF(ghost_key_check), ghost), 1 SECONDS)
+
+/obj/structure/overmap/proc/spec_ghostship_changes() //Proc to buff ghost ships. Must be overridden to apply buffs
+	return
```
Karmic-Skink, line 106: But why call an empty proc?

nsv13/code/modules/overmap/types/syndicate.dm
```diff
@@ -619,3 +631,15 @@
 /obj/structure/overmap/syndicate/ai/bomber/apply_weapons()
 	weapon_types[FIRE_MODE_ANTI_AIR] = new/datum/ship_weapon/light_cannon(src)
 	weapon_types[FIRE_MODE_TORPEDO] = new/datum/ship_weapon/torpedo_launcher(src)
+
+/obj/structure/overmap/syndicate/ai/bomber/spec_ghostship_changes() //Buffs fighters so dogfights are more fun.
```
Karmic-Skink, line 635: Why is the bomber almost exactly the same as the fighter?

Karmic-Skink (changes requested):
Calling an empty proc is inefficient.

Also having this be a look up table whereby each fighter is exactly the same is kinda bad.
tbh I'd rather have blanket multipliers rather than carbon copy tables.

Also Also: this is missing the solgov fighter.

nsv13/code/modules/overmap/overmap_ghosts.dm
DeltaFire15: Standard practice for OOP (basetype proc interface that just returns without doing anything)
Proccall overhead for calling empty proc should be comparable to running an if to check for said empty proc and this solution is the cleaner one inheritance-wise.

DeltaFire15 (commented):
As outlined in my comment on karmic's question about the empty proc, I believe the previous empty return for basetype implementation was better. Proccall overhead should be negligible when compared to having to do a check instead, and it keeps overmaps from having *another* variable on every single one of them (the var bloat on them has gotten very very bad over time since we keep adding more)

DeltaFire15:
Also if you want to reduce copy spaghetti, you could in theory have procs on the overmap level designed for specific classes (well, probably only fighter for now since thats what the PR is for), 
then call that proc in the fighter-type specific overrides of the spec_ghostship-whatever proc.
Should reduce copypasta to a single line and make it easy to make blanket adjustments.

someone543:
> Also if you want to reduce copy spaghetti, you could in theory have procs on the overmap level designed for specific classes (well, probably only fighter for now since thats what the PR is for), 
> 
> then call that proc in the fighter-type specific overrides of the spec_ghostship-whatever proc.
> 
> Should reduce copypasta (in the individual overrides for the bunch of fighter types) to a single line each and make it easy to make blanket adjustments.

Sorry, but I'm confused on what you want me to do. If you want me to apply buffs depending on the type of ship, isn't that what's currently going on with if(mass == MASS_TINY)?

Also sorry for the late reply

Bokkiewokkie (approved):

PR #2412, opened 2023-04-30
Minor weather improvement

covertcorvid:
## About The Pull Request
This isn't the full rework because trying to write our own event handling system is hurting my soul and there's a lot of other stuff going on.
Instead, this is a tweak to the way system events currently work. This makes events no longer trigger every time the ship enters a system. Instead, it biases normal round event selection towards those events that relate to the system type.

## Why It's Good For The Game
Variety

## Changelog
:cl:
tweak: Removed immediate event trigger from FTL jump end. Added chance for normal event subsystem to pick from system-related events instead of full event list.
/:cl:


code/controllers/subsystem/events.dm
```diff
@@ -64,7 +64,16 @@ SUBSYSTEM_DEF(events)
 	// Only alive, non-AFK human players count towards this.
 
 	var/sum_of_weights = 0
-	for(var/datum/round_event_control/E in control)
+
+	//NSV13 - are we using our event list or the full one?
+	var/obj/structure/overmap/mainship = SSstar_system.find_main_overmap()
+	var/list/possible_events
+	if(mainship.current_system && length(mainship.current_system.possible_events) && prob(50))
+		possible_events = mainship.current_system.possible_events
```
Bokkiewokkie, line 72: This also makes it really easy to trigger events for other ships as well when we change events to be compatible with that (thinking abt. PvP or boarding ships)

Bokkiewokkie (approved):
:+1: Not getting harassed every jump but still running the event timer seems like a good idea. ~~it also means we can make more devastating rad events again~~

PR #2451, opened 2023-06-15
Skinsuit temperature resistance upgrade

SerynEngi:
## About The Pull Request

This PR gives the emergency skinsuit proper thermal protection against space.

## Why It's Good For The Game

Unlike standard SS13, we spend considerably larger amounts of time in space, both willingly and unwillingly. This change will enable everything from non-engineering repairs on the hull to munitions techs not having to hunt up public access suits to actually work on the gun that just got blown up.

## Changelog
:cl:
tweak: Changed the skinsuit to properly protect you from cold.
/:cl:


Bokkiewokkie:
After playing with this I think it makes skinsuits too powerful, because there are no other negative effects such as slowdowns (which is the major downside to skinsuits) in zero-g (space).

SerynEngi:
> After playing with this I think it makes skinsuits too powerful, because there are no other negative effects such as slowdowns (which is the major downside to skinsuits) in zero-g (space).

Fair enough. Tossed it up here to see how it played out. Didn't play out as well as it might.

PR #2459, opened 2023-06-25
Personal Armor Bullet/Laser rebalance

SerynEngi:
## About The Pull Request

This PR specifically brings up bullet armor values and brings down laser armor values across the board.

## Why It's Good For The Game

Most enemies do not use lasers, and bullet protection was subpar. This corrects this issue so those wearing the appropriate armor can take more bullets than the standard graytider. This also up-armors the warden's coat on par with the also up-armored HOS greatcoat. 

## Changelog
:cl:
balance: Increased bullet armor values of all standard armor and armored hardsuits.
balance: Decreased laser armor values of all standard armor and armored hardsuits.
/:cl:

### DIFF
--- a/code/modules/clothing/spacesuits/hardsuit.dm
+++ b/code/modules/clothing/spacesuits/hardsuit.dm
@@ -684,7 +684,7 @@
 	icon_state = "hardsuit0-sec"
 	item_state = "sec_helm"
 	hardsuit_type = "sec"
-	armor = list("melee" = 35, "bullet" = 15, "laser" = 30,"energy" = 40, "bomb" = 10, "bio" = 100, "rad" = 50, "fire" = 75, "acid" = 75, "stamina" = 50)
+	armor = list("melee" = 35, "bullet" = 45, "laser" = 15,"energy" = 40, "bomb" = 10, "bio" = 100, "rad" = 50, "fire" = 75, "acid" = 75, "stamina" = 50)
 
 
 /obj/item/clothing/suit/space/hardsuit/security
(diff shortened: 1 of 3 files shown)


Karmic-Skink:
If the concern is bullets v lasers, why are these numbers not just reversed?

SerynEngi:
> If the concern is bullets v lasers, why are these numbers not just reversed?

This was actually my original intention.

However, after some discussion with bokkie and dusk, I elected to bring down the laser resistance as a whole to a minimal level (most obvious with the captains equipment changes) for the reason that lasers are just not common.

The reason the bullet armor was actually brought up, again after some discussion, was for sec personnel to actually properly fight boarders. They'll now take roughly twice the number of rounds to take down when compared to our unarmored ship-side personnel, thus encouraging people to either actually utilize the security gear, or let sec actually handle it.

BlueHNT:
My concern is that SWAT will be completely overlooked due to the huge slowdown it has compared to its mostly mediocre values for the stuff it's intended to do. Got any plans for it, or will you just give it this minor boost and toss it?

SerynEngi:
> My concern is that SWAT will be completely overlooked due to the huge slowdown it has compared to its mostly mediocre values for the stuff it's intended to do. Got any plans for it, or will you just give it this minor boost and toss it?

Baseline security equipment is really intended to be better, as a general rule, than the basic Swat Mk1 equipment that can be ordered, at least for combat. The captain's hardsuit locker utilizes the Swat Mk. 2 which is in fact better. The reason the HOS hardsuit has the strongest full body protection is that they should be leading the charge against boarders.

That said, the face that it actually has full protection over the limbs (as it is not a resistance, but actual armor values designed to cover all protected limbs) is its advantage over something like a bulletproof vest. Sure, its slow, but it will also let you take a ton of hits in any area as well as store things. Its a trade off, as intended.

Bokkiewokkie (approved):

PR #2460, opened 2023-06-26
Сrate sprites port from Paradise

Blundir:
## About The Pull Request

Resprites the crates with the sprites from the https://github.com/ParadiseSS13/Paradise
Also took a few crate sprites from tg (they no longer use these, after their recent resprite).
Soulful pack respectful to previous sprites, while making them in 3/4 perspective, also making them look better in general.
Had to remove the animation tho, due to jank of vertical animation.

## Why It's Good For The Game

Here is a comparison:
Before:
dreammaker_6Sj4CkyfRe

After:
dreammaker_fXykEiUA9m

<details>

dreamseeker_rAEaNoftBo

</details>

## Changelog
:cl:
imageadd: crate resprite with sprites from Paradise
/:cl:


red031000:
ok imo these need two things before they should even be *considered*
1. to have animations that work (and look good)
2. an attept (at least) at a bee upstream port (which I don't honestly think will work but it's much better to at least do it that way in terms of code)

ofc I'm not a coder so in the end it's up to them, and this is my personal opinion

Blundir:
> ok imo these need two things before they should even be _considered_
> 
> 1. to have animations that work (and look good)
> 2. an attempt (at least) at a bee upstream port (which I don't honestly think will work but it's much better to at least do it that way in terms of code)
> 
> ofc I'm not a coder so in the end it's up to them, and this is my personal opinion

Upstream is very unlikely to approve this, also i did pr completely modular, so no need for it really, about the animation, i think i know how to redo it, but i'd only do that if sprites would be approved.

DeltaFire15:
Current maintainer consensus on this PR is to not merge it.

PR #2462, opened 2023-06-27
SMP-II

BlueHNT:
## About The Pull Request
I felt like there is a distinct lack of RnD ballistic weaponry, so I made a gun. It's a 45acp SMG with a 2-round quick burst and normal speed full auto. The only real distinct thing about it is the 2-speed burst which is slightly faster, roughly on par with Glock.

## Why It's Good For The Game
Weapon variety for R&D. Has not been implemented there as of this PR.

<details>

image

</details>

## Changelog
:cl:
add: Added an experimental SMP-II SMG
add: Added 45 caliber 30-round lethal magazine
add: Added 45 caliber 30-round rubber magazine
add: Added 45 caliber 30-round armour-piercing magazine
/:cl:


BlueHNT:
updated in hands (right, mirrored left)
image
image

smaller SMG
image
image

Bokkiewokkie:
I'm gonna be honest, this isn't what NSV needs right now.
After playing security myself for a while now I disagree with the idea that sec needs any new equipment, since they're already the most heavily armed force that isn't an ERT, nuke ops team, or deathsquad. (If anything, I would consider locking equipment which is *already* accessible roundstart right now behind research)

Even if we did want or need a new gun, I wouldn't want it to be this. As it stands this is just a c20r but for Sec, even firing the exact same ammo (but without a bayonet lug :<). On top of that the sprite doesn't match our other guns, and making a sprite bigger than 32x32 just gets in the way most of the time.
You also added the research designs to "radioactive weapons" which seems badly categorized unless these bullets are made of polonium (which doesn't seem to be the case)

If you really want to add this to the game you should port it to Beestation, but I don't think they'll accept it like this either.

PR #2493, opened 2023-07-29
Overmap Ship normal damage type immunity

SerynEngi:
## About The Pull Request

This PR adds armor to overmap ships of all types for all non-shipboard overmap armor types.

## Why It's Good For The Game

Reduces/prevents the rare overmap bug exploitation from killing ships.

## Changelog
:cl:
fix: Added normal armor to all overmap ships.
/:cl:

### DIFF
--- a/nsv13/code/__DEFINES/overmap.dm
+++ b/nsv13/code/__DEFINES/overmap.dm
@@ -13,7 +13,9 @@
 #define FIRE_MODE_BROADSIDE 9
 #define FIRE_MODE_PHORON 10
 
+//Base Armor Values
 
+#define OM_ARMOR "melee" = 100, "bullet" = 100, "laser" = 100, "energy" = 100, "bomb" = 80, "bio" = 100, "rad" = 100, "acid" = 100, "stamina" = 100
 
 //Deprecated / legacy weapons.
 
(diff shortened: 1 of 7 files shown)


Bokkiewokkie (changes requested):
They shouldn't have 100 bomb protection because explosions on the overmap are funny (they can still have a lot, however).
Fire damage is also applied by incendiary weapons and shouldn't have any.

DeltaFire15 (changes requested):
This is so much copypasta, maybe make a OVERMAP_BASE_RESISTANCES define and put the copypasted part of the armor list in there? Since defines are replaced at compile time it should be fine to use in that list.

Bokkiewokkie:
> This is so much copypasta, maybe make a OVERMAP_BASE_ARMOR_VALUES define and put the copypasted part of the armor list in there? Since defines are replaced at compile time it should be fine to use in that list.

Agree with this, it also makes it easier to adjust globally (since I see you still missed a couple of fire armour values)

Karmic-Skink:
There is a proc to add armour values that be added to late initialize etc

Bokkiewokkie:
> There is a proc to add armour values that be added to late initialize etc

@SerynEngi can you actually do this instead, you only have to add it to the /obj/structure/overmap type once

Bokkiewokkie (changes requested):
see comment above

Bokkiewokkie (approved):
Yes excellent commit :+1:

PR #2496, opened 2023-07-30
Upped the firerate on the mech LBX AC 10 "Scattershot"

Brillcrafter:
## About The Pull Request

This is a simple value change, from 20 cooldown to 10 cooldown. Now to 15 cooldown after feedback
## Why It's Good For The Game

When I was using the scattershot in a round, the ROF was too slow for it to be really useful compared to the other mech weapons, the ROF has been increased to make it on par with the other weapons.

## Changelog
:cl:
Balance: increased the ROF from 0.5 RPM to 0.75 RPM, decreased from 1 RPM to 0.75 after feedback.
/:cl:

### DIFF
--- a/code/game/mecha/equipment/weapons/weapons.dm
+++ b/code/game/mecha/equipment/weapons/weapons.dm
@@ -293,7 +293,7 @@
 	name = "\improper LBX AC 10 \"Scattershot\""
 	desc = "A weapon for combat exosuits. Shoots a spread of pellets."
 	icon_state = "mecha_scatter"
-	equip_cooldown = 10
+	equip_cooldown = 10  //NSV13 - halved cooldown
 	projectile = /obj/item/projectile/bullet/scattershot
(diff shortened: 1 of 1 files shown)


Bokkiewokkie (changes requested):
Since this is a core file change it should either be ported to [our upstream](https://github.com/BeeStation/BeeStation-Hornet) first or given a tag like `//NSV13 - halved cooldown`.

DeltaFire15:
A single burst from this thing applies a concentrated 72 damage and can also just be weaved with other weapons during cooldown. This seems unneccessary.

Brillcrafter:
But the spread is incredibly high, if you are about 2-3 blocks away about half the pellets will miss. Which would equal about 30 damage Also swapping weapons quickly is very hard due to the button being in the action bar.

Bokkiewokkie (approved):
We can just revert this if we think it's too much

PR #2499, opened 2023-07-31
Makes MP16 not obsolete

BlueHNT:
## About The Pull Request
Gives it a 2 round controlled burst and new sprites.

Please do leave comments and opinions how to further improve it.

## Why It's Good For The Game
There are so many guns in the game, yet all of 'em are just... decorative pieces from 2015.

<details>

image
image
image
image

</details>

## Changelog
:cl:
balance: rebalanced MP16 burst by giving it a burst
sprite: Replaced the MP16 sprites
/:cl:


Bokkiewokkie (approved):
I think if we do anything else to it we should make it more unique rather than just balancing every gun to be closer to the c20r, but the sprites are better :+1:

BlueHNT:
It's a 9mm so the only real thing you can give such a small caliber smg is a fast firerate if anything.. Make it more like MP5 while the CR20 stays more... UMP like.

Bokkiewokkie:
> It's a 9mm so the only real thing you can give such a small caliber smg is a fast firerate if anything.. Make it more like MP5 while the CR20 stays more... UMP like.
> 
> Issue with that is if you do put it 1 notch up in firerate it goes up A LOT.

Yeah every 1 fire rate is 0.1 seconds of fire delay less

PR #2505, opened 2023-08-05
only send mail 20% of the time

Kenionatus:
## About The Pull Request
Makes mail crates have a 20% chance to be on each cargo shuttle instead of always beign there.

## Why It's Good For The Game
Due to a frequent lack of cargo staff and the often quite busy gameplay loop of NSV, mail tends to pile up in cargo. Only sending a mail crate some of the time should reduce that without flat out removing mail, which would be a horrible thing for bluewasabi, who put a bunch of work into making custom mail for NSV.

Too lazy to do that for a two line change. Real chads test on production anyway.
<details>

</details>

## Changelog
:cl:
balance: Each cargo shuttle now only has a 20% chance of carrying a mail crate.
/:cl:

### DIFF
--- a/code/modules/shuttle/supply.dm
+++ b/code/modules/shuttle/supply.dm
@@ -71,7 +71,8 @@ GLOBAL_LIST_INIT(blacklisted_cargo_types, typecacheof(list(
 
 /obj/docking_port/mobile/supply/initiate_docking()
 	if(getDockedId() == "supply_away") // Buy when we leave home.
-		create_mail()
+		if(prob(20)) //NSV13 Only send mail 1/5 of the time
+			create_mail()
 		buy()
 	. = ..() // Fly/enter transit.
 	if(. != DOCKING_SUCCESS)


Kenionatus:
The number is of course open for discussion.

Bokkiewokkie:
I feel like this is a map issue more than anything (Atlas moment) and the ideal solution would just be giving Atlas a bigger cargo shuttle (maybe where the arrivals shuttle used to be since we removed it?).
Making this fun mechanic completely random seems like a waste, so I would suggest making it an option to get mail or not on the cargo console.

Kenionatus:
I don't think it's a map issue since I dislike mail piling up in any size of cargo area. The mail switch idea definitely makes sense from a gameplay standpoint, although I don't really like it from a thematic/narrative standpoint. Independently from that, I don't see myself implementing such a switch in the near future cause I'm too lazy to learn TGUI.

PR #2511, opened 2023-08-10
APNW slider can be set to zero now

### LINKED ISSUE #2321: APNW Power Allocation Problem
## Testmerges:
None
Although this section of code hasn’t been touched since the PR that introduced it:
* https://github.com/BeeStation/NSV13/pull/774

## Reproduction:
Currently there exists an issue with the APNW Power Allocation slider, the problem being that you can’t lower it back down to 0 after moving it from its starting position.

This issue is due to how it looks in the code itself

```
    var/adjust = text2num(params["adjust"])
    if(action == "power_allocation")
        if(adjust && isnum(adjust))
            power_allocation = adjust
            if(power_allocation > maximum_power_allocation)
                power_allocation = maximum_power_allocation
                return
            if(power_allocation < 0)
                power_allocation = 0
                return
```

The problem is the fact that var/adjust is equal to text2num of the slider value...and the following conditional statement checks if adjust is true, or exists, alongside checking if adjust is a number.
`if(adjust && isnum(adjust))`
So if you set the Armor Pump Well to accept a grand total of 0 watts, it will not alter it because adjust is now equal to 0

Which is a number, however it’s also the logic meaning for FALSE, so the conditional statement is then set to the following
`if(FALSE && TRUE)`
Which in logic is equivalent to FALSE as 0 && 1 is always 0

bruhlookatthisdood:
## About The Pull Request

fixes #2321 
tbh i just looked at it and said that looked weird and removed it, hopefully I didn't fuck something else up :

## Why It's Good For The Game

it's nice to be able to do that

## Changelog
:cl:
fix: APNW can be set to zero now.
/:cl:


nsv13/code/modules/overmap/armour/nano_well.dm
```diff
@@ -358,7 +358,7 @@ Starting Materials
 		return
 	var/adjust = text2num(params["adjust"])
 	if(action == "power_allocation")
-		if(adjust && isnum(adjust))
+		if(isnum(adjust))
```
DeltaFire15, line 361: while at it, may aswell just CLAMP the value in the lines below (and then return) instead of doing some unneccessary if filtering

Bokkiewokkie (approved):

PR #2516, opened 2023-08-13
Delta's "five minute" pylon adventure

DeltaFire15:
## About The Pull Request
What started as a "fun quick PR" to display the nucleium use of FTL pylons to the main console UI quickly started deteriorating when I went to test and started to notice issue upon issue which pylons have had ever since they were merged (I seem to have a sense for running into a billion unfixed issues). 
So. Instead of just PRing this little thing, I've went through pylon code and refurbished as much as I could without losing my last bits of sanity from realizing the issues in the first place.
This results in hopefully more stable pylons that do not shut down from a single tick of power hickup (power code do be like that at times), makes a good chunk of their code work in the first place, and does some mild balancing to round it off.
Oh. And of course the very small feature that I opened up the file for in the first place. That one did take only five minutes.

## Why It's Good For The Game
Fix man good.

My suffering will have to suffice.

## Changelog
:cl:
fix: FTL Pylons are now not doomed to run out of nucleium even when supplied.
fix: FTL Pylons with closed shielding now actually have an increasing nucleium demand..
balance: .. due to the nucleium increase from shielding actually working, power drain scaling when shielded has been drastically cut.
tweak: FTL Pylons need multiple cycles of failed power to shut down (to prevent power code oddities)
balance: FTL Pylons now also output plasma when they are running, instead of only when booting (why did they not?)
code: Improves various other bits and pieces of code related to FTL Pylons that do not need individual mention
qol: Nucleium use of FTL Pylons is now visible on the main console.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/FTL/components/drive.dm
+++ b/nsv13/code/modules/overmap/FTL/components/drive.dm
@@ -311,6 +311,7 @@ Preset classes of FTL drive with pre-programmed behaviours
 		pylon_info["gyro"] = round(P.gyro_speed / P.req_gyro_speed, 0.01)
 		pylon_info["capacitor"] = round(P.capacitor / P.req_capacitor, 0.01)
 		pylon_info["draw"] = display_power(P.power_draw)
+		pylon_info["nucleium"] = round(P.get_nucleium_use() / 2, 0.01) //Converted into mol / second, SSmachines runs every 2 seconds.
 		pylon_info["shielded"] = P.shielded
 		all_pylon_info[++all_pylon_info.len] = pylon_info // Unfortunately, this is currently the best way to embed lists
 	data["pylons"] = all_pylon_info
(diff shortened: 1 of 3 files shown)


Bokkiewokkie (approved):

DeltaFire15:
> Sometimes drive pylons enter a voodoo state where they can't be toggled and are off, but they provide ftl charge rate.

Odd. Shouldn't be caused by my PR since I haven't touched how exactly shutdown works, only its conditions, but I'll see if I can reproduce it.

Bokkiewokkie:
> > Sometimes drive pylons enter a voodoo state where they can't be toggled and are off, but they provide ftl charge rate.
> 
> Odd. Shouldn't be caused by my PR since I haven't touched how exactly shutdown works, only its conditions, but I'll see if I can reproduce it.

Assuming you want to add this to the PR then, you can merge it yourself if you'd rather make a separate one

PR #2525, opened 2023-08-18
Adds a overmap z failsafe

DeltaFire15:
## About The Pull Request
Basically, we sometimes have ships with people end up in nullspace that really shouldn't be there.
This adds a failsafe that makes any ship with a certain overmap flag (which all ships that used instance_overmap have), or those with any player inside that might notice, reaquire / create an overmap z if they somehow lose theirs when they arrive at a system.
This should mitigate effects from the variety of things that might cause this to happen (worst case admins should be able to manually trigger the failsafe by using add_ship, if it somehow race conditions and breaks)

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Overmap objects with players should be a bit less likely to get stuck in nullspace.
/:cl:


Bokkiewokkie (dismissed):

Bokkiewokkie (approved):
Haven't seen many things break on the overmap after this

PR #2572, opened 2023-10-29
Round extension and enemy spawning tweaks

Bokkiewokkie:
## About The Pull Request

When the round extension mission is given, it will now check if Rubicon is already cleared and assign Dolos as the objective instead. I've also raised the requirement for these missions to 10, since four players will usually have a harder time getting a successful Rubicon mission.
On top of this I've automatically made the clear system objective unhide the target system, made enemy fleet spawning only happen in sector 2 (So no more empty starmaps because all of the ships spawned in the Syndicate sector) and adjusted the way briefings are generated to make it clear which paper is the extension objective and to maintain more consistency between the papers.
Also adds a hide/unhide button to the starsystem manager, and a label for systems that are hidden (see picture for preview)

## Why It's Good For The Game

These are quality of life features for admins, and increases the challenge for people who want to keep fighting after a successful mission.

I don't have a video but I did actually test this
Preview of hidden system on the starsystem manager:
image

## Changelog
:cl:
add: Added the option to get another ticket objective if Rubicon is already cleared
tweak: Increased Rubicon objective player requirement to 10
tweak: Mission briefing names will now be consistent in a round
balance: Fleets will now be stronger after the main mission is done
admin: Added button to hide/unhide a system on the starsystem manager
admin: Automatically unhides systems selected for the "clear system" objective
/:cl:

### DIFF
--- a/nsv13/code/game/gamemodes/overmap/objectives/rubicon.dm
+++ b/nsv13/code/game/gamemodes/overmap/objectives/rubicon.dm
@@ -26,6 +26,7 @@
 	desc = "Defeat all enemies in the [system_name] system"
 	brief = desc
 	target_system = SSstar_system.system_by_id(system_name)
+	target_system.hidden = FALSE
 	RegisterSignal(SSstar_system.find_main_overmap(), COMSIG_SHIP_KILLED_FLEET, PROC_REF(check_completion))
 
 /datum/overmap_objective/clear_system/check_completion()
@@ -38,8 +39,9 @@
 /datum/overmap_objective/clear_system/rubicon
 	system_name = "Rubicon"
 	extension_supported = TRUE
-	required_players = 4
+	required_players = 10
 
 /datum/overmap_objective/clear_system/dolos
 	system_name = "Dolos Remnants"
-	required_players = 4
+	extension_supported = TRUE //Only if Rubicon is not available
+	required_players = 10
--- a/nsv13/code/controllers/subsystem/overmap_mode.dm
+++ b/nsv13/code/controllers/subsystem/overmap_mode.dm
@@ -330,11 +330,8 @@ SUBSYSTEM_DEF(overmap_mode)
 	*/
 
 	var/datum/star_system/rubicon = SSstar_system.system_by_id("Rubicon")
-	if(get_active_player_count(TRUE,TRUE,FALSE) > 10)
-		if(length(rubicon.enemies_in_system))
-			mode.objectives += new /datum/overmap_objective/clear_system/rubicon
-		else //If Rubicon is empty we go to Dolos
-			mode.objectives += new /datum/overmap_objective/clear_system/dolos
+	if(get_active_player_count(TRUE,TRUE,FALSE) > 10 && length(rubicon.enemies_in_system)) //Make sure there are enemies to fight
+		mode.objectives += new /datum/overmap_objective/clear_system/rubicon
 	else
 		mode.objectives += new /datum/overmap_objective/tickets
 		for(var/datum/faction/F in SSstar_system.factions)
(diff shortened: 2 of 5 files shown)


DeltaFire15:
>made enemy fleet spawning only happen in sector 2 (So no more empty starmaps because all of the ships spawned in the Syndicate sector) 

All fleets that spawn in sector 3 via faction ticks (aka time) should also already navigate to sector 2 since a long time ago.

Bokkiewokkie:
> > made enemy fleet spawning only happen in sector 2 (So no more empty starmaps because all of the ships spawned in the Syndicate sector)
> 
> All fleets that spawn in sector 3 via faction ticks (aka time) should also already navigate to sector 2 since a long time ago.

This just means they end up piling up in Rubicon

DeltaFire15:
>This just means they end up piling up in Rubicon, which makes the fight unnecessarily difficult
They navigate to sector 2 *individual systems*, not rubicon or the sector itself, though they may be there momentarily while passing through. (Rubicon itself is also not a valid system for said navigation)

Bokkiewokkie:
> > This just means they end up piling up in Rubicon, which makes the fight unnecessarily difficult
> 
> They navigate to sector 2 _individual systems_, not rubicon or the sector itself, though they may be there momentarily while passing through. (Rubicon itself is also not a valid destination for said navigation)

Alright then, I think spawning them only in sector 2 is just easier regardless though

DuskLight9978:
> > This just means they end up piling up in Rubicon, which makes the fight unnecessarily difficult
> 
> They navigate to sector 2 _individual systems_, not Rubicon or the sector itself, though they may be there momentarily while passing through. (Rubicon itself is also not a valid destination for said navigation)

Fleets normally do not travel much in my experience, from feeling they only maneuver based on their assigned 'objective' IE: patrol and defend fleets only moving to friendly syndicate systems; This search for objective points normally only goes out one system out from their current system UNLESS its the Hunt players objective or a go to certain system. Otherwise, if a fleet tasked with taking neutral systems is surrounded by syndicate systems.. It does not move, I am not certain about moving from 3>2 as I don't normally look at sector 3, but I rarely see things leaving Rubicon. This also fits with what I was told in voice that they have a numerical representation of an objective that needs to be overridden or they do their own thing... if its in the one jump distance range. 

It may be that Skynet borked, but that has been the experience.

DeltaFire15:
>or a go to certain system
Which is exactly what that does. The entirity of non-random fleet navigation is my code. It hooks into that maybe a dozen lines after the line modified in this PR in faction fleet spawning (l. 107+ in factions.dm)

DuskLight9978:
> > or a go to certain system
> 
> Which is exactly what that does. The entirity of non-random fleet navigation is my code. It hooks into that maybe a dozen lines after the line modified in this PR in faction fleet spawning (aka l. 107+ in factions.dm)

I only see specific fleets have a 'go to system' in their variable, the rest do not activate it from the last time I messed with it, Such as the solbuster fleet, could it be their not setting their go to system properly?

nsv13/code/game/gamemodes/overmap/objectives/rubicon.dm
```diff
@@ -38,8 +39,9 @@
 /datum/overmap_objective/clear_system/rubicon
 	system_name = "Rubicon"
 	extension_supported = TRUE
-	required_players = 4
+	required_players = 10
 
 /datum/overmap_objective/clear_system/dolos
 	system_name = "Dolos Remnants"
-	required_players = 4
+	extension_supported = TRUE //Only if Rubicon is not available
```
DeltaFire15, line 46: This var doesn't actually do anything as the only section using it is commented out.

nsv13/code/game/gamemodes/overmap/objectives/rubicon.dm
```diff
@@ -38,8 +39,9 @@
...
+	required_players = 10
```
DeltaFire15, line 47: This system doesn't actually contain any permanent garrison fleets and is fairly likely to end up being a free objective besides for the very minor random ships spawning or a fleet stopping by.

nsv13/code/game/gamemodes/overmap/objectives/rubicon.dm
Bokkiewokkie: Yeah I've just set it because the variable exists and I want it to be included in a future system that uses it.

nsv13/code/game/gamemodes/overmap/objectives/rubicon.dm
Bokkiewokkie: Oh you're right, I assumed the remnant was there all the time.

PR #2573, opened 2023-11-02
Adds special offers to overmap traders

Bokkiewokkie:
## About The Pull Request

Adds the special offer category of trader item, notably including **firing electronics** and full repairs.
I've also made PDC mount boards available without a specific limit, we can change this later if it's a problem.
These offers will require a certain threshold of faction points to be reached, after which they become available in their respective stores. Special offer prices will have less fluctuation, but are generally more expensive by default.

## Why It's Good For The Game

This has been a plan for a while now, though some of the gun dynamics will obviously change with 2.0 which will likely include a partial rework of the balance of this as well.

image

## Changelog
:cl:
add: Added special offers to traders, which have to be unlocked with faction ticket points
balance: Made gun parts available for purchase at Whiterapids' Munitions
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/traders_items.dm
+++ b/nsv13/code/modules/overmap/traders_items.dm
@@ -188,7 +188,7 @@
 	desc = "Essential electronics for building most modern naval weapons."
 	price = 20000
 	stock = 2
-	unlock_path = /obj/item/circuitboard/machine/vls
+	unlock_path = /obj/item/ship_weapon/parts/firing_electronics
 	special_requirement = 500 //Kick their ass!!!
 
 /datum/trader_item/pdc_circuit
--- a/nsv13/code/modules/overmap/traders_items.dm
+++ b/nsv13/code/modules/overmap/traders_items.dm
@@ -77,6 +77,15 @@
 	repair_amount = 50
 	stock = 2
 
+/datum/trader_item/ship_repair/tier3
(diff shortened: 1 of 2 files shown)


covertcorvid:
Do we want to make these no longer undeletable since we're adding ways of getting more?

Bokkiewokkie:
> Do we want to make these no longer undeletable since we're adding ways of getting more?

I think we should do that seperately, as they are critical ship parts and not available before having done combat

nsv13/code/modules/overmap/traders_items.dm
```diff
@@ -166,17 +175,40 @@
...
+	stock = 10
+	unlock_path = /obj/item/circuitboard/machine/vls
+	special_requirement = 100 //At least put some effort into it
+
+/datum/trader_item/firing_electronics
+	name = "Firing Electronics"
+	desc = "Essential electronics for building most modern naval weapons."
+	price = 20000
+	stock = 2
+	unlock_path = /obj/item/circuitboard/machine/vls
```
DeltaFire15, line 191: This should probably be firing electronics, not a VLS board like the sold item above.

Bokkiewokkie: Yep... that's pretty essential

Bokkiewokkie:
PR label bot error?

PR #2588, opened 2023-12-02
Torp Factory 1.5

DeltaFire15:
## About The Pull Request
Lets face it, the munitions torp automation right now has a tendency to be more trouble than it's worth. The machines are very clunky to handle, only work half the time (which neccessitated slower belts despite having been thrown into faster processing), and are incredibly processing inefficient.

This PR reworks almost everything that was related to the torp / missile assemblers, cleaning up code, making it magnitudes less expensive to idle (seriously the processing & iteration was horrible)
The machines now do not process at all, instead operating on their watched turf being entered by something they care about, and only then springing to action. This reduces their dry load to 0, and makes them only have any kind of performance requirement when things actually appear, which they also only have to iterate on once as opposed to every process() even if invalid.

They also now pick up components of the type last manually (by hand) inserted, should they pass the tile the assembler is watching.

https://github.com/BeeStation/NSV13/assets/46569814/3837f8a9-0829-4421-8ff5-1af86797b8ce

It is recommended to only keep a single type of component per assembler as there is no reason to have more than one right now, but if there is more than one, the one currently in first place (that will be applied next) is the one it will grab parts of.

In general, they also react MUCH faster thanks to being signal-based (superfast) instead of processing (guh)

Look at it go!

https://github.com/BeeStation/NSV13/assets/46569814/1bfa1df0-0a4e-4e6a-bee0-da9af8523e6a

It could run on normal belts now instead of the slowed belt subtype (hell you can throw components past it and the assembler will catch them), however I didn't feel like editing the maps.

Also fixes two random bugs, one with the FTL Thirring Computer soundloop that has been broken since it was added, and a random GC failure with clients_by_z which did not account for mob deletions.

## Why It's Good For The Game
Torp automation has been left somewhat neglected and lacking effective usefulness. This improves functionality, making a well-supplied assembly line a torpedo (or missile) powerhouse.

Also fix man good.

Already in the PR desc body itself.

## Changelog
:cl:
refactor: The Torp and Missile factory automation machinery has been rewritten almost entirely to make it less clunky and more fun.
add: Automated Assemblers will now pick up components of the last manually inserted type if they pass in front of them!
fix: The Thirring Drive Core sound loop you have never heard before works now!
fix: The client_by_z list shouldn't be getting flooded with nulls anymore.
/:cl:

### DIFF
--- a/code/modules/mob/living/living.dm
+++ b/code/modules/mob/living/living.dm
@@ -32,6 +32,11 @@
 				qdel(S)
 			else
 				S.be_replaced()
+	//NSV13 - stop runtiming
+	if(registered_z)
+		SSmobs.clients_by_zlevel[registered_z] -= src
+		registered_z = null
+	//NSV13 end
 	if(ranged_ability)
 		ranged_ability.remove_ranged_ability(src)
 	if(buckled)
--- a/code/modules/recycling/conveyor.dm
+++ b/code/modules/recycling/conveyor.dm
@@ -195,6 +195,11 @@ GLOBAL_LIST_EMPTY(conveyors_by_id)
 	if(!operating)
 		SSmove_manager.stop_looping(convayable, SSconveyors)
 		return
+	//NSV13 - account for potentially changing location between sig send and reaching this.
+	if(convayable.loc != source)
+		SSmove_manager.stop_looping(convayable, SSconveyors)
+		return
+	//NSV13 end.
 	start_conveying(convayable)
 
 /obj/machinery/conveyor/proc/conveyable_exit(datum/source, atom/convayable, direction)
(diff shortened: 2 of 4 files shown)


Bokkiewokkie (dismissed):
Finally, factory without process

Bokkiewokkie (dismissed):
Why would they call that convayable... I had to double check that wasn't a typo

DeltaFire15:
Sound now has a cooldown so you shouldn't get your ears factory blasted & the picked up item is now decided by the last manually inserted one instead of the one first in the current list
(both not in with the current compile though)

Bokkiewokkie (approved):

PR #2628, opened 2024-04-21
Relayed BSA is more 'fun' and also random fixes.

DeltaFire15:
## About The Pull Request
While the (true) BSA has a lot of oomph on the overmap itself, its actual effects when relayed are somewhat less interesting as it's effectively just a hitscan-penning NAC that leaves no trace once it passed. This gives it a large explosion and some rather unpleasant effects on direct hits instead.
I considered letting it actually truncate ships & spacing a line, but given we do have pvp content on the overmap sometimes (or if some bus contains a foe with a BSA), that'd be a bit rude . Also is similar to the effects when *someone* fires the thing with the shutters closed.

Also fixes a lot of random things that runtime since I test with pause on runtime enabled and also look at logs sometimes. Sidetracking my beloved.

## Why It's Good For The Game
Explosions are nice.

Also fix man good.

No.

## Changelog
:cl:
balance: The BSA now goes kaboom when relayed instead of being essentially a ghost NAC shell.
fix: The Eclipse no longer spawns a borg module that should never exist outside of mobs.
fix: DNA no longer has some super weird code for flavor text, instead copying the one of the source mob.
tweak: You can no longer stack interactions on walls. This is to prevent a runtime from actions that modify the type of them (deconstruction mainly).
fix: Paper no longer runtimes if you do not have a pen in hand (???).
fix: Traders no longer do some cursed things to create their trade items, instead using the all-powerful initial().
fix: The stormdrive no longer runtimes if it has zero gas in it.
fix: Some squad assignment stuff shouldn't try to use a list as datum anymore.
/:cl:

### DIFF
--- a/code/modules/paperwork/paper.dm
+++ b/code/modules/paperwork/paper.dm
@@ -474,7 +474,12 @@
 		if(clipboard.pen)
 			holding = clipboard.pen
 
-	data["held_item_details"] = holding?.get_writing_implement_details()
+	//NSV13 - Old part here runtimes.
+	if(holding)
+		data["held_item_details"] = holding.get_writing_implement_details()
+	else
+		data["held_item_details"] = null //Overriding old data, otherwise data would not get changed.
+	//NSV13 end.
 
 	// If the paper is on an unwritable noticeboard, clear the held item details so it's read-only.
 	if(istype(loc, /obj/structure/noticeboard))
--- a/_maps/map_files/Eclipse/Eclipse1.dmm
+++ b/_maps/map_files/Eclipse/Eclipse1.dmm
@@ -262,7 +262,7 @@
 "bp" = (
 /obj/structure/closet/crate/large,
 /obj/effect/spawner/lootdrop/maintenance/four,
-/obj/item/robot_module/borgi,
+/obj/item/borg/upgrade/transform/borgi,
 /turf/open/floor/plasteel/dark/side,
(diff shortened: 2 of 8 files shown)


code/modules/paperwork/paper.dm
```diff
@@ -474,7 +474,12 @@
 		if(clipboard.pen)
 			holding = clipboard.pen
 
-	data["held_item_details"] = holding?.get_writing_implement_details()
+	//NSV13 - Old part here runtimes.
+	if(holding)
+		data["held_item_details"] = holding.get_writing_implement_details()
+	else
+		data["held_item_details"] = null //Overriding old data, otherwise data would not get changed.
+	//NSV13 end.
```
Bokkiewokkie, line 482: Could make this a ternary like:
```suggestion
	data["held_item_details"] = holding ? holding.get_writing_implement_details() : null //NSV13 Overriding old data, otherwise data would not get changed.
```
But yours is also valid

Bokkiewokkie (approved):
Epic fix PR

code/modules/paperwork/paper.dm
DeltaFire15: I think I prefer this as an elif to emphasize the "passing null does something specific" for whoever has to do a beebase that touches this.

Bokkiewokkie (approved):

PR #2652, opened 2024-05-14
Added a cross-server comms cooldown failure message

Pockets-byte:
… comms also being cross-time

basically what it says. there's nothing more to it

## Why It's Good For The Game

casual time travel is casual

## Changelog
:cl:
tweak: added a rejection message to the comms console
/:cl:


williamcll:
Remind me again how many years are there between bee and nsv?

code/game/machinery/computer/communications.dm
```diff
@@ -246,6 +246,7 @@
 			if (!can_send_messages_to_other_sectors(usr))
 				return
 			if (!COOLDOWN_FINISHED(src, important_action_cooldown))
+				to_chat(user, "<span class='alert'>Temporal communication array recharging. Please stand by.</span>")
```
DeltaFire15, line 249: usr instead of user.

DeltaFire15 (dismissed):
Not a big fan of this going for "time travel".

I prefer the implicit relative time adjustment from simple "interstation comms", aka you here pretend the beestation you are talking to is the one of the current year, and beestation pretends the patrol ship they're getting a call from is in their time.

Otherwise the entire "people talk to eachother like they know eachother", which obviously happens, makes zero sense (besides the whole thing about there not being any known means of time travel in the scenario [Temporal Agents are not real]).

code/game/machinery/computer/communications.dm
Pockets-byte: ack

Pockets-byte:
> Not a big fan of this going for "time travel".
> 
> I prefer the implicit relative time adjustment from simple "interstation comms", aka you here pretend the beestation you are talking to is the one of the current year, and beestation pretends the patrol ship they're getting a call from is in their time.
> 
> Otherwise the entire "people talk to eachother like they know eachother", which obviously happens, makes zero sense (besides the whole thing about there not being any known means of time travel in the scenario [Temporal Agents are not real]).

I think it's more in line with the setting to have casual time travel that's used for inane reasons - reminder that chrono legionaries are not the only time effects that can be seen- things like wizards or sepia slime stuff. and I mean- people could go it the long way around- and possibly imply that this can be done elsewhere- it's a plot hole either way, and I personally think this is a better(and funner) way of handling it than not addressing it at all- and in addition- this is time communication, not time travel

williamcll:
> Not a big fan of this going for "time travel".
> 
> I prefer the implicit relative time adjustment from simple "interstation comms", aka you here pretend the beestation you are talking to is the one of the current year, and beestation pretends the patrol ship they're getting a call from is in their time.
> 
> Otherwise the entire "people talk to eachother like they know eachother", which obviously happens, makes zero sense (besides the whole thing about there not being any known means of time travel in the scenario [Temporal Agents are not real]).

normally when someone mentions this in-game I say it's "time dilation"

Pockets-byte:
changed it to something a little more vague

Bokkiewokkie (approved):
Please change your PR title and description and re-request a review when you make a relevant change

PR #2663, opened 2024-05-25
Removes integrity failure value from all overmap entities

DeltaFire15:
## About The Pull Request
Title. We are setting this despite it not being used for any overmap stuff , and if it did the values would all be wrong. The basetype does it right with having it set to 0, so that is what remains.

## Why It's Good For The Game
Fix man good.
(Technically it's also a micro performance increase because this stops an empty proc from being called every time a ship takes damage)

## Changelog
:cl:
code: Integrity_failure variable settings removed from all overmaps. (This does nothing)
/:cl:


Bokkiewokkie (dismissed):

Bokkiewokkie (approved):
Merge conflicts my beloathed

PR #2668, opened 2024-05-28
Updates the armour pumps to account for lag properly

Bokkiewokkie:
## About The Pull Request

Adds delta_time to the armour pumps' process(), to avoid lag affecting the pump performance.

## Why It's Good For The Game

Now the pumps should actually work at the same speed all the time, most importantly not slowing down when there is game lag from things like big fleet battles and atmospherics processing during a firetorping.

## Changelog
:cl:
fix: Fixed armour pumps not accounting for game lag in their processing
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/armour/nano_pump.dm
+++ b/nsv13/code/modules/overmap/armour/nano_pump.dm
@@ -58,7 +58,7 @@
 	if(stress_shutdown)
 		. += "<span class='warning'>The system overload lights are flashing</span>"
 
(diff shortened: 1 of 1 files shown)


DeltaFire15 (changes requested):
deltat needs to be divided by 2 due to the machinery processing ticks being 2 seconds apart, yet deltat being passed as delay / 10, or you will buff APNP by a factor of 2x

DeltaFire15 (approved):

PR #2680, opened 2024-06-08
adds a warning noise when ships with shields take damage to the hull

Pockets-byte:
## About The Pull Request

there was an unused shield failure noise in the game files. it's used now

<details>

</details>

## Changelog
:cl:
add: added a warning sound for hull hits when the shields fail
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/weapons/damage.dm
+++ b/nsv13/code/modules/overmap/weapons/damage.dm
@@ -82,7 +82,10 @@ Bullet reactions
 		blocked = TRUE
 		damage_sound = pick('nsv13/sound/effects/ship/damage/shield_hit.ogg', 'nsv13/sound/effects/ship/damage/shield_hit2.ogg')
 		if(!impact_sound_cooldown)
-			add_overlay(new /obj/effect/temp_visual/overmap_shield_hit(get_turf(src), src))
+			var/obj/effect/temp_visual/overmap_shield_hit/E = new /obj/effect/temp_visual/overmap_shield_hit(get_turf(src), src)
+			add_overlay(E)
+			E.pixel_y = pixel_y
+			E.pixel_x = pixel_x
 	if(!impact_sound_cooldown && damage_sound)
 		relay(damage_sound)
 		if(damage_amount >= 15) //Flak begone
--- a/nsv13/code/modules/overmap/weapons/damage.dm
+++ b/nsv13/code/modules/overmap/weapons/damage.dm
@@ -43,6 +43,11 @@ Bullet reactions
 				return BULLET_ACT_FORCE_PIERCE // :)
 			else
 				return FALSE //Shields absorbed the hit, so don't relay the projectile
+		if(shield_result == SHIELD_NOEFFECT && !failure_sound_cooldown)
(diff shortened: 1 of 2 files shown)


SerynEngi:
I have nothing to say pertaining to whether or not we use this effect.

I've explained before that this variety of shield generator doesn't really have a down state except when it loses power. You already know that the shield is insufficient for the hit when you start taking the hits on the hull, so a "down" state is something I view as unneeded.

_What I will say is that the effect isn't unused_. This sound is straight from the PDSR.

PDSR

Pockets-byte:
> I have nothing to say pertaining to whether or not we use this effect.
> 
> I've explained before that this variety of shield generator doesn't really have a down state except when it loses power. You already know that the shield is insufficient for the hit when you start taking the hits on the hull, so a "down" state is something I view as unneeded.
> 
> _What I will say is that the effect isn't unused_. This sound is straight from the PDSR.
> 
> PDSR

I didn't actually check to see if it was used which is my bad, in this case all it does is play this sound when a ship with shields has the shields not stop something,  with a fairly long cooldown(for sounds)- this would probably happen with the pdsr as well

nsv13/code/modules/overmap/weapons/damage.dm
```diff
@@ -43,6 +43,11 @@ Bullet reactions
 				return BULLET_ACT_FORCE_PIERCE // :)
 			else
 				return FALSE //Shields absorbed the hit, so don't relay the projectile
+		if(shield_result == SHIELD_NOEFFECT && !failure_sound_cooldown)
```
SerynEngi, line 46: Okay, _now_ I have a comment on the code. If I'm reading this right, it looks like it will play the shields down effect at the same time it would play the projectile hitting sound with a cooldown of 5 seconds. 

Now, I'm no audio engineer, but that sounds like adding another sound that doesn't tell you something new (the sound of the projectile striking the hull is very obvious, and in a lot of cases, excessively loud), is just a bad idea.

Pockets-byte: I personally think it's nice to have the little extra warning- it can be hard to tell sometimes, but I could look into moving the trigger into when the shield's charge is dropped to zero- but that'll play even when you aren't actually taking damage

SerynEngi: I have nearly 300hrs in engineering and probably a solid 20hrs of that standing at that shield generator. It is _never_ at zero for more than a moment unless the power requested is higher than the power provided. 

_It is also extremely obvious to everyone when the shield are down because the shield deflect sounds stop playing and the hull hit effects start._

If you want to add in a warning effect for when the shields are at 33% of effective strength or less with an appropriate cooldown, that would be useful- and would not be attached to actual projectile hits. _**It would belong in the shield generator dm file and not in damage.dm**_

SerynEngi: Even more importantly, this sound effect will also trigger with the PDSR, as this is damage.dm and the PDSR uses this sound for another reason. So you're going to have that PDSR sound playing for 2 separate reasons on the Galactica.

Pockets-byte: fine whatever. I'll just close this then

PR #2681, opened 2024-06-08
Makes shields hits only spawn one visual (which is actually aligned with the ship)

Bokkiewokkie:
## About The Pull Request

Title, fixes an age old bug

## Why It's Good For The Game
Intended behaviour

I'm not making a recording for 5 seconds of shield footage sorry
Update: At this point I've tested it so much I did honestly consider making an "Aetherwhisp getting attacked by pirates ASMR" video

## Changelog
:cl:
fix: Fixed shield sprites leaving doubles when the ship gets hit
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/shieldgen.dm
+++ b/nsv13/code/modules/overmap/shieldgen.dm
@@ -338,7 +338,7 @@
 
 /obj/effect/temp_visual/overmap_shield_hit
 	name = "Shield hit"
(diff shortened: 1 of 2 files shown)


Pockets-byte:
doesn't quite work right on the server

https://github.com/BeeStation/NSV13/assets/55299415/292f0204-a518-4489-b9f7-8db9d5437b9f
see above

PR #2683, opened 2024-06-15
Stormdrive Radiation Buff

SerynEngi:
## About The Pull Request

This PR amplifies the native radiation output of the Stormdrive in a manner directly relative to its output.

## Why It's Good For The Game

In addition to encouraging engineers to make pluoxium for usage in experimental stormdrive mixes, this also makes the stormdrive relatively dangerous to approach needlessly after it has been started in the same manner that a similar PR made it unwise to stand in front of a PA.

Radiation testing was done using an 18.5MW mix at full power
Before- 1tile away: 225 counts
After- 1 tile away: 4.5k counts/ 5 tiles away (through open air): 150 counts

## Changelog
:cl:
tweak: Increased the radiation output of the Stormdrive
/:cl:

### DIFF
--- a/_maps/map_files/Atlas/atlas.dmm
+++ b/_maps/map_files/Atlas/atlas.dmm
@@ -1461,7 +1461,10 @@
 /turf/open/floor/monotile/steel,
 /area/nsv/weapons/gauss/battery/deck2/port)
 "eN" = (
-/obj/machinery/atmospherics/components/unary/vent_pump/on/layer2,
+/obj/effect/landmark/start/station_engineer,
+/obj/machinery/atmospherics/components/unary/vent_pump/on/layer2{
+	dir = 4
+	},
 /turf/open/floor/carpet/orange,
 /area/engine/engine_room)
 "eO" = (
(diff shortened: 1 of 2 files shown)


jet-pack-cat:
Good to see plasmamen buffs. 

Was skeptical at first, tested it, the radiation is lethal, not a challenge to deal with, on atlas at least. And it generates pluoxium just as well as any supermatter shard, generating about half the radiation when set up properly. All while being significantly faster and easier to set up (takes about 20 mins at most, probably way less once you learn it), plus it doesn't kill the crew. 

Got around 23mw, 26mw with additional rad collectors.
33% constricted plasma
33% nitrogen
33% pluoxium
1000kpa input pressure
66% pluoxium on the mixer to adjust for temperature difference of the pluoxium, no need for cooling or direct contact heat exchange like with tritium mixes, temperature is mostly stable. Adds a reason to actually monitor the storm drive since the mix will very slowly drift.
1.3% rod insertion (simulated minimum rod insertion before thermal runaway, should theoretically generate 25mw but in practice does less)

Only concern is that the ftl drive room may start to glow. Gives lethal a lethal dose to anyone without rad suits in engineering, even with the weaker mixes.

SerynEngi:
> Only concern is that the ftl drive room may start to glow. Gives a lethal dose to anyone without rad suits in engineering, even with the weaker mixes.

Hmmm... it would be easy enough for me to slightly alter the room to give it that extra distance (and another wall) on the atlas, specifically because is a non-reactor controls area that by virtue of the design has to be relatively close.

Part of the consideration for this was the fact that the regular engineering hardsuit is resistant enough to deal with the few casualties you actually have to enter the room for, (shattered pipes most often) even if it downs you afterwards.

Fun fact, its actually not as potent, radioactively, as a CO2 SM, its somewhere between 1/3 and 1/2 as potent.

PR #2684, opened 2024-06-15
[Semi-Experimental] The coldbloodening

DeltaFire15:
## About The Pull Request
Lizards in our lore are coldblooded. Those in our game are not, and have some fairly noninteractive traits (digitigrade is optional, so the only noteworthy drawback is cold damage mod which rarely comes up).

This replaces their basic "cold mod" modifier with the coldblooded property.
What being cold-blooded does is that it negates automatic body temperature stabilization, which means temperature adjustment is done purely by environmental factors.
This means that having ones body at non-standard temperatures becomes more of an active consideration, as an area with desired temperatures becomes the main way of temperature adjustment (besides varied methods such as chemicals, nanites, xenobio shenanegans, etc).
They do still preserve their resilience to direct heat-based damage.

In some talks regarding the matter, it was mentioned some other servers having heat sharing done via hugging (help intent), however while I liked that concept, I did not like it being on a trivial & spammable action (that would also need spamming).
Instead, this adds new functionality to "aggressive" grabs, which makes them share temperature between grabber and grabee while maintained. This is affected by insulation of both targets, albeit weighted somewhat towards transmissibility, unless either target has total insulation in which case it does not occur.
This in my opinion is a more fun and interactive take on the previously mentioned hug temperature sharing, and aggressive grabs fit that closeness more accurately in my opinion.

**Feel free to suggest interesting things regarding this change.** This does not have to be its final form.

### Revision 1

Added thermogenesis to ectotherms. Essentially, some cold-blooded creatures in nature can somewhat counteract the cold by repeatedly contracting & relaxing (basically vibrating) their muscles, as the energy usage in muscles releases some heat.

This adds an equivalent of this to lizards. If they remain at temperatures significantly below normal (<~15°C currently) for an amount of time, they will subconsciously increase muscle activity in an attempt to heat up (aka shiver), even more so if *very* cold. This is inefficient, requires nutrition if active, can be a bit slow to respond, and doesn't feel good (mood debuff).

### Revision 1.5

PR now also contains some random kneecapping of cult level mood buffs from simple drink sources within the NSV modular mood events. Who knew all moths were blood cultists?

Lizards now have a "comfy" bodytemperature range of ~30-60°C, in which they gain a mood buff as opposed to the debuff for being cold. Fairly difficult to maintain since they lack any kind of natural stabilization and the ship is 20°C unless you HE pipe or cook the department.

Also fixes a long-standing bug with heat protection calculations that would allow you to bleed or gain negligible amounts of heat with full protection due to decimal accuracy and the protection value being split into 11 small values being added together.
Usually body temperature stabilization would be bigger than the values added by this, but guess who doesn't do that!

## Why It's Good For The Game
Species uniqueness is fun.

### Revision 1
Some demonstrations added.
#### Avoid these if using mobile data (video files do be big)

<details> 
<summary>Thermogenesis demo</summary>
(semi-outdated - messages related to it got changed, but functionality is the same)

https://github.com/BeeStation/NSV13/assets/46569814/212c7f7c-3437-4174-85f1-14a2c0e60df7

</details>

<details>
<summary>Grab temperature exchange demo</summary>

https://github.com/BeeStation/NSV13/assets/46569814/6e673235-598b-4bff-ae30-cc8bc7459fe8

</details>

## Changelog
:cl:
balance: Lizards are now coldblooded (they have no natural bodytemperature stabilization besides the environment).
balance: Coldblooded creatures also have an inefficient form of thermogenesis they can use to heat up.
balance: Lizards dislike the cold (<15°C / when you start shivering) and like warmth (30°C-60°C), which can affect mood.
balance: Lizards no longer have a 1.5 cold damage modifier.
balance: Having somebody in an aggressive grab will share temperature between the grabber and grabbed target over time (respecting insulation).
balance: Drinking blood as a moth now has a weaker mood buff.
balance: Naval coffee now has a weaker mood buff.
fix: Temperature protection values returned by the getter are no longer very very slightly off the value they should be. Thanks decimal precision.
/:cl:


Lauwma:
Shouldn't that be a upstream change ?

DeltaFire15:
> Shouldn't that be a upstream change ?

I am not interested in maintaining this change upstream on a server I am not knowledgeable about regarding its balancing & design-style.

code/modules/mob/living/carbon/life.dm
```diff
@@ -547,6 +547,10 @@ GLOBAL_LIST_INIT(ballmer_windows_me_msg, list("Yo man, what if, we like, uh, put
 
 //used in human and monkey handle_environment()
 /mob/living/carbon/proc/natural_bodytemperature_stabilization()
+	//NSV13 - coldblooded beings do not naturally stabilize.
+	if(HAS_TRAIT(src, TRAIT_COLDBLOODED))
+		return 0
+	//NSV13 end.
```
Bokkiewokkie, line 553:
```suggestion
	//NSV13 - coldblooded beings do not naturally stabilize.
	if(HAS_TRAIT(src, TRAIT_COLDBLOODED))
		return 0
	//NSV13 end.
```
You could also simulate this by just lowering the metabolism efficiency a lot. Cold-blooded animals still generate *some* metabolic heat.

DeltaFire15: Ectothermism is the inability to *control* body-temperature. Metabolism is essentially a constant factor for heat (albeit potentially slightly higher or lower depending on activity).

Emulating the effect of metabolism on body heat would be a constant heat gain over time for (almost) all species, which would be much more punishing than this is, especially on NSV where having to suit up is common.

DeltaFire15: Active thermogenesis via means such as muscle activity for the sake of exotherm energy usage could be considered I suppose, while ignoring the default metabolism inherent to most beings for sake of not causing things to melt in any insulated equipment.

This would delta exclusively upwards, be slow, and require additional nutrition to maintain (while below a certain temperature)

DeltaFire15: My take on this sort of behavior implemented. See testing section for a demo.

nsv13/code/datums/mood_events/nsv_events.dm
```diff
@@ -1,6 +1,6 @@
 /datum/mood_event/moth_drink_blood
 	description = "<span class='nicegreen'>That hit the spot!</span>\n"
-	mood_change = 10
+	mood_change = 10 //Why is a moth drinking blood a mood buff the level of being a cultist?? And for *ten minutes*???
 	timeout = 10 MINUTES
```
Bokkiewokkie, line 4:
```suggestion
	mood_change = 3
	timeout = 7 MINUTES
```
You can also just suggest a change, code comments aren't really the place for complaints.

nsv13/code/datums/mood_events/nsv_events.dm
```diff
@@ -52,3 +52,7 @@
 	description = "<span class='nicegreen'>Cheers! ¡Salud! Kanpai! Prost! Skål! Santé! Sláinte! Saúde!</span>\n"
 	mood_change = 3
 	timeout = 30 SECONDS
+
+/datum/mood_event/lizard_vibrations
+	description = "<span class='warning'>I know my muscles are shaking to warm me up, but it still feels distracting..</span>\n"
+	mood_change = -2
```
Bokkiewokkie, line 58:
```suggestion

/datum/mood_event/lizard_vibrations
	description = "<span class='warning'>I'm shivering!</span>\n"
	mood_change = -2
```
I think it would be better to word this less technically, you could also consider calling it shivering instead of lizard vibrations, then, for a potential more general application

Bokkiewokkie (changes requested):
Code itself looks good! And the shivers are a good solution to make coldbloodedness less devastating in space. Just a few notes

nsv13/code/datums/mood_events/nsv_events.dm
DeltaFire15: I usually leave these sorts of comments as pointer to myself or others to change this in the future so it doesn't go forgotten, since sometimes I *do* manage to contain the urge to resolve every single bit of misbehavior in the same PR.

nsv13/code/datums/mood_events/nsv_events.dm
DeltaFire15: I suppose this *is* essentially just shivering. Bit carried away on terminology. I do want to leave an indicator in the moodlet that this sort of shivering is because of the environmental cold.

DeltaFire15: Adjusted. Still lizard specific since this is just three lines if someone wants to replicate it.

nsv13/code/datums/mood_events/nsv_events.dm
Bokkiewokkie: I think the new values I chose are more fitting for the type of moodlet it is (based on other drink related buffs). I'm fine with just having that in this PR since it's not too big of a thing. I can also open it as a new PR but then please remove the comment to avoid a conflict.

Bokkiewokkie (approved):
Adding "ate without table moodlet" after this

PR #2688, opened 2024-06-29
Insulation tweak

DeltaFire15:
## About The Pull Request
As we all know, if you are hot, insulation is bad for you. It keeps in the heat and makes you stay---
Actually that isn't the case. For whichever reason, being warm has additional insulation bypassing calculations that make you bleed off heat like crazy, despite you already having a mechanism for that, aka `natural_bodytemperature_stabilization()`.

This means that you can even bleed  *double* the heat you actually should if your insulation is already low, since it applies whether or not you are insulated.
(In addition to your active temperature control via stabilization)

It irks me. A lot. Insulation doesn't care whether you want to be warm or not; it is up to the one using it (or not using it if trying to use environmental temperature) to handle that conundrum.

This PR removes the additional calculation from temperature adjustment calculation, making it function just as it should without magically tunneling heat through your liquid-nitrogen-proof suit.

## Why It's Good For The Game
There is absolutely no reason for this to exist. If you are hot, and trying to cool down via the air, the natural response is to remove insulation. There is no such special case for being cold (thankfully) and there should be none for being hot either.

## Changelog
:cl:
balance: Being warm no longer bypasses insulation (sort of, it was worse than that but I'm not elaborating here) when in a colder environment than you.
/:cl:


Ikalpo:
does this make it worse to deal with spicy air? because if so I don't want it.

DeltaFire15:
> does this make it worse to deal with spicy air? because if so I don't want it.

That depends entirely on how you define "deal with".
Suited in something like an atmos suit? No change.
Currently being cooked by the area? No change.
Left the area and had no insulation? No change. (excluding a certain specific bodytemperature range where temperature bleed would be higher than it should be before)
Wearing something really good at retaining heat and left the area after being cooked? You should take it off to cool down faster. This is the only case where the change should be noticable, which is the same behavior as if you got, say, tempgunned on freeze mode while wearing a fire suit without this change.

Bokkiewokkie:
FYI, I've been waiting with this until the auxmos update is fullmerged

Bokkiewokkie (approved):

PR #2695, opened 2024-07-13
Fixes broadside cannon direction recognition

### LINKED ISSUE #2648: Broadsides shooting only one side
## Round ID & Round Date:
12.05.2024

## Testmerges:
N/A

## Reproduction:
Major issue on both Snake and Hammerhead as broadsides currently only shoot the direction they decide to shoot in. Tends to be left. I am pretty sure this is most likely a code a issue as it just entirely ignores 2 of the NACs on board further nerfing the already aged and unsupported (currently) hammerhead. Snake is cool, but isn't played due to the same reason of Broadsides annoying and bad. I love the broadsides as a design, but currently they just... aren't fun.

## Picture Included

image

DeltaFire15:
## About The Pull Request
The proc deciding which broadside cannon to fire assumes things about the angle variable that aren't true.
This makes the proc more reliable, which makes the correct cannon fire all the time instead of only until angle goes above 360 or below 0.

Fixes #2648 

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: Broadside cannons should now always fire on the correct side.
/:cl:

### DIFF
--- a/nsv13/code/__HELPERS/misc.dm
+++ b/nsv13/code/__HELPERS/misc.dm
@@ -14,7 +14,8 @@
 
 ///Whether the angle is on the port or starboard side of the ship (facing north or south on the map)
 /proc/angle2dir_ship(angle)
-	if(0 < angle && angle < 180)
+	var/modulated_angle = (((angle % 360) + 360) % 360)
+	if(modulated_angle <= 180)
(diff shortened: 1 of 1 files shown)


Kenionatus:
Should fix #2648 if it works.

PR #2697, opened 2024-07-13
Fixes the "You are Dead" tgui window popup

DeltaFire15:
## About The Pull Request
Title.
At the moment, it shows up but is never properly cleared because usually client is already null when the ghost's logout is called, and even if it wasn't that would be a bad place to clear it.
This moves the procs to clear that message to revival, ghost grabbing, and the mob Login() proc, the latter two only being triggered if the mob is actually alive, which should be much more reliable.

## Why It's Good For The Game
Fix man good.

## Changelog
:cl:
fix: The "You are Dead" chat window popup should now be properly removed when you are not dead.
/:cl:


Bokkiewokkie (approved):
Same opinion on porting as mentioned in the dir resting PR

PR #2693, opened 2024-07-05
Space mines, Syndicate stealth attacks

Bokkiewokkie:
## About The Pull Request

This PR adds space mines, a new weapon invented by combining Syndicate cloaking technology with high-yield explosives. Currently Aspala-class submarines, Mako flak frigates, and combat carriers will be able to drop them when fleeing or patrolling.

## Why It's Good For The Game

More fun stuff to ~die to~, it also makes Syndicate systems more dangerous by spawning a few mines in them at roundstart.

## Testing Photographs and Procedure
I tested it. Only shooting doesn't work nicely (you need to actually click on a mine to hit it right now)

## Changelog
:cl:
add: Added space mines
/:cl:

PR #2710, opened 2024-09-08
Some more overmap fixes

### LINKED ISSUE #2701: Broken black hole
Issue reported from  Round ID: 7665 (NSV 13 | Official | MRP | 98% Free Lag | Spaceship Combat Server)

Reporting client version: 515.1641

On round 7663 in NSV Snake I made a flight to Veneto and into the black hole. The result? Pirates and our ship just pass through the hole harmlessly, the screen wasn't darkened and no gravitational pull, only disabling the inertia dampeners.

It worked properly before.

DeltaFire15:
## About The Pull Request
#2693 adjusts singularities to use the system contents to track affected, but fails to account for the fact not all contents of `system_contents` are overmap objects (as opposed to `GLOB.overmap_objects`).
This PR does similar but safer, while also reinforcing some sanity a bit (current_system was not 100% reliable at being set for anomalies and spawn_ship flooded system_contents with typepaths)

Also makes overmaps firing projectiles a bit less likely to runtime if they are near mapedge or were firing while being moved to nulspace.

This fixes #2701 (Well technically it is a testmerge breaking it but this does that part except without breaking).

## Why It's Good For The Game
Fix man good.

This is tested on local.

## Changelog
:cl:
code: Singularities now use system contents in a safe way.
fix: Spawn_ship no longer floods system_contents with typepaths.
fix: Overmap firing should be a bit more resilient to turf-related oddities.
/:cl:

### DIFF
--- a/nsv13/code/controllers/subsystem/starsystem.dm
+++ b/nsv13/code/controllers/subsystem/starsystem.dm
@@ -271,7 +271,6 @@ Returns a faction datum by its name (case insensitive!)
 	return system
 
 /datum/controller/subsystem/star_system/proc/spawn_ship(obj/structure/overmap/OM, datum/star_system/target_sys, center=FALSE)//Select Ship to Spawn and Location via Z-Trait
-	target_sys.system_contents += OM
 	if(target_sys.occupying_z)
 		var/turf/destination = null
 		if(center)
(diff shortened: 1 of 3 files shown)


Bokkiewokkie (approved):
At some point we probably should just make an add_ship() that actually handles different types of objects. But this works for now

PR #2714, opened 2024-10-11
Various KNPC fixes and tweaks

Bokkiewokkie:
## About The Pull Request

"Improves" KNPC code and pathfinding a bit, the mainly being how climbable objects are handled. It also stops them from just leaving their leaving guns on the ground after they've climbed a table or something else.

## Why It's Good For The Game
More challenging "smarter" npc behaviour can be more fun, and to balance it out making them a bit slower at shooting etc. to avoid a squad of them instantly gunning people down.

## Changelog
:cl:
add: Added knpcs to the ghost orbit menu under the npc section
balance: made knpcs take actions a bit slower
fix: fixed inflatable walls not having a range check on their deployment
code: made knpcs use human procs instead of forcemoving things around constantly
refactor: refactored knpc climbing
/:cl:

### DIFF
--- a/code/__HELPERS/path.dm
+++ b/code/__HELPERS/path.dm
@@ -143,8 +143,8 @@
 	if(!start || !end)
 		stack_trace("Invalid A* start or destination")
 		return
-	// if(start.z != end.z || start == end ) //no pathfinding between z levels
-	//	return //NSV13 we DO want pathfinding between z levels.
+	if(start == end) //no pathfinding between z levels //NSV13 we DO want pathfinding between z levels. removed "start.z != end.z || " from the check
+		return
 	if(max_distance && (max_distance < get_dist(start, end))) //if start turf is farther than max_distance from end turf, no need to do anything
 		return
 
--- a/code/__HELPERS/path.dm
+++ b/code/__HELPERS/path.dm
@@ -143,8 +143,8 @@
 	if(!start || !end)
 		stack_trace("Invalid A* start or destination")
 		return
-	if(start.z != end.z || start == end ) //no pathfinding between z levels
-		return
+	// if(start.z != end.z || start == end ) //no pathfinding between z levels
+	//	return //NSV13 we DO want pathfinding between z levels.
 	if(max_distance && (max_distance < get_dist(start, end))) //if start turf is farther than max_distance from end turf, no need to do anything
 		return
 
--- a/code/__HELPERS/path.dm
+++ b/code/__HELPERS/path.dm
@@ -337,16 +337,21 @@
  * * ID: An ID card that decides if we can gain access to doors that would otherwise block a turf
  * * simulated_only: Do we only worry about turfs with simulated atmos, most notably things that aren't space?
 */
+//NSV13 refactored this to use a typecache for directional objects in src
 /turf/proc/LinkBlockedWithAccess(turf/destination_turf, caller, ID)
 	var/actual_dir = get_dir(src, destination_turf)
+	var/static/list/directionals = typecacheof(list(
+		/obj/structure/window,
+		/obj/machinery/door/window,
+		/obj/structure/railing,
+		/obj/structure/barricade,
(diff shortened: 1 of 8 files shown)


nsv13/code/modules/overmap/knpc.dm
```diff
@@ -154,11 +161,11 @@ GLOBAL_LIST_EMPTY(knpcs)
 				continue
 			if(!blocking_firelock.density || blocking_firelock.operating)
 				continue
-			if(blocking_firelock.welded)
+			if((blocking_firelock.welded))
 				break	//If at least one firedoor in our way is welded shut, welp!
 			blocking_firelock.open()	//Open one firelock per tile per try.
 			break
-		for(var/obj/machinery/door/firedoor/blocking_firelock in this_turf)
+		for(var/obj/machinery/door/firedoor/blocking_firelock as() in this_turf)
```
DeltaFire15, line 168: Unsafe call, this_turf's contents is not exclusively firelocks.

code/__HELPERS/path.dm
```diff
@@ -143,8 +143,8 @@
 	if(!start || !end)
 		stack_trace("Invalid A* start or destination")
 		return
-	if(start.z != end.z || start == end ) //no pathfinding between z levels
-		return
+	// if(start.z != end.z || start == end ) //no pathfinding between z levels
```
DeltaFire15, line 146: Should still not permit cases of start = end.

Bokkiewokkie: true!

nsv13/code/modules/overmap/knpc.dm
Bokkiewokkie: Oh idk why that's there, I guess I got distracted and forgot to remove it when undoing some refactors :p

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -376,7 +406,13 @@ This is to account for sec Ju-Jitsuing boarding commandos.
...
 	var/obj/item/storage/S = H.back
 	var/obj/item/gun/target_item = null
-	//Okay first off, is the gun already on our person?
+	//We must have lost our gun somehow, get it from the floor if we simply dropped it.
+	if(istype(H.loc, /turf))
+		var/turf/T = H.loc
+		target_item = locate(/obj/item/gun) in T.contents
+		if(H.put_in_hands(target_item))
+			return TRUE
+	//Otherwise, is there a gun already on our person?
```
DeltaFire15, line 415: Potentially loops forever if we just intentionally dropped our gun (e.g. because out of ammo) assuming there is another valid weapon somewhere on the ground nearby.

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -353,9 +382,10 @@ This is to account for sec Ju-Jitsuing boarding commandos.
...
 	var/mob/living/carbon/human/H = HA.parent
-	var/obj/item/gun/G = H.get_active_held_item()
+	var/obj/A = H.get_active_held_item()
+	var/obj/B = H.get_inactive_held_item() //check your other hand, just in case.
 	//We already have a gun
-	if(G && istype(G))
+	if((A && istype(A, /obj/item/gun)) || (B && istype(B, /obj/item/gun)))
 		return 0
 	var/obj/item/gun/G_New = locate(/obj/item/gun) in oview(HA.view_range, H)
 	if(G_New && gun_suitable(H, G_New))
```
DeltaFire15, line 391: On a side note despite you not actively editing this line, this should probably iterate through all guns it finds in the range instead of just locate()ing one out of that list, since it will just return exactly one gun and judge based on that single gun, even if there are five suitable ones nearby.

Bokkiewokkie: Yeah so far there's been a lot more code I've *wanted* to refactor but not had the time for 😅

PR #2715, opened 2024-10-16
[Port] Hardsuit Tails

DeltaFire15:
## About The Pull Request
Some talk on NSV led me to remember I saw armored tails on hardsuits somewhere somewhen. In turn, I did manage to remember where I remembered that from, tracked down where they ported it from, and, in turn, ported that here.

This essentially ports https://github.com/Skyrat-SS13/Skyrat-tg/pull/2183, albeit without Vulpine sprites (We don't have Foxes. Except Renault.), aswell as doing some mildly "funky" implementation stuff because our mutant bodypart overlay proc (and especially accessory backend) vary, even relative to the point of time this is from.

For now has the same color setups as the base PR (luckily, as opposed to bee, most of our hardsuits still have "old" sprites).
However, that also means that some NSV-unique suits do currently lack this, like Pilot Hardsuits.
I've spent enough of today's brain power on this for now though, so we'll see at some point.

See screenshots section for included suits.

The Felinid tails are also kind of, big, compared to how they look without a suit, but eh better than them being like three pixels wide (maybe the suit is just thick!). Do tell if you prefer them this way or crammed into the suit, no big feelings in either direction (except it not being very comfy to have a tail in there aswell as the rest of you).

## Why It's Good For The Game
Bluespace Tail Storage is expensive.

Having your tail just vanish when you put on a hardsuit has been a minor recurring annoyance across servers for me. Now they do not, and you have some more distinction even in hardsuits.

<details>

Base suits (minus RD and HOS because I oopsied):
tailsuits_standard_ANN

RD & HOS suit (because I oopsied):
OOPS_ANN

Special suits:
tailsuits_special_ANN

</details>

## Changelog
:cl: NotRanged, Azarak, DeltaFire15
add: Tails should now be visible in hardsuits that support them.
imageadd: Tails in hardsuits.
/:cl:


Bokkiewokkie (changes requested):
Since this is NSV13, can we also add tail color combinations for our own suits? They should all be in nsv13/icons/mob/suit.dmi, if you need a color reference. I feel like the skinsuit, radsuit, and regular EVA suit (and pirate eva suit) would all also make sense (even more so than the actual hardsuits) to give tails.
On top of this, the current "colored" part of the ERT suit tails seems a bit oversaturated, maybe you can just pick a color from the actual suit for those (I assume the colors are different because Skyrat has some more vibrant suits)

The rest looks fine, originally I was surprised to see code instead of sprites, but it's executed quite nicely here.

DeltaFire15:
> Since this is NSV13, can we also add tail color combinations for our own suits? 

That is the intent, I just spent hours thinking about how to translate that implementation from their more bespoke overlay 
handling here and have zero energy left to see about cooking overlay values.

>I feel like the skinsuit, radsuit, and regular EVA suit (and pirate eva suit) would all also make sense

Note to self.

>On top of this, the current "colored" part of the ERT suit tails seems a bit oversaturated, maybe you can just pick a color from the actual suit for those (I assume the colors are different because Skyrat has some more vibrant suits).

We can certainly adjust values if needed since the handling was made for that.

> The rest looks fine, originally I was surprised to see code instead of sprites, but it's executed quite nicely here.

It certainly is a useful implementation since it makes it easy for compat for new suits, albeit at some slightly lower variance, but it suffices unless someone decides to sprite tails for days.

DeltaFire15:
Current Flightsuit take (kinda dark? Unfortunately staring at 32x32 sprite colors for too long makes my head hurt but feel free to try your own palettes)
image

(The rogue pixel on the cat is NOT me, that is a Flightsuit sprite issue that also occurs on master)

Bokkiewokkie:
Merging this, we can add other color variations in a new PR

PR #2721, opened 2024-12-08
Keycard authenticators now authenticate keycards

DeltaFire15:
## About The Pull Request
Keycard auth devices currently do not check whether the target using them has an ID card, despite being card auth devices (implying some physical component to the authentication).
They also do not ensure the two keycards being used are not the same, which allows a single user to perform this activity.

This PR makes keycard authenticators require an ID to swipe (This affects Silicons since they fall into the "has access but does not have IDs" category. They shouldn't really have access to these as opposed to humanoid crew, which is the main point of this requirement).

Additionally, the IDs used for triggering & confirming the requested action must be different. This is a "soft" requirement in a way, as it does not prevent you from making two IDs and switching between them while also managing to reach both devices, intended as a weak point while also not being as easy to achieve as the original bypassing methods.

Mirrored upstream at https://github.com/BeeStation/BeeStation-Hornet/pull/12002

I remembered I made this change elsewhere ages ago when I saw two authenticators reachable from the *same tile* on the Tycoon, so I decided to remake that old PR here (and mirror it to bee so I don't have to //nsv13 but if it gets closed I'll just do that)

## Why It's Good For The Game
Keycard Authentication Devices should authenticate keycards.
Their main intent was to create a two-step auth process akin to submarine movies, but at this time can easily be bypassed via either silicons (who are not humanoid nor command and thus NT wouldn't trust with these), or via quick movement methods / two auths being near eachother.
This rectifies the former, while also making it somewhat more difficult to "solo auth", incentivizing command cooperation.

<details>

(Testing was repeated here but screenshots are from bee testing since I don't feel like taking new ones :) )

<details>
<summary>Silicon response</summary>

image

</details>

<details>
<summary>Evil captain trying to solo auth</summary>

image

</details>

</details>

## Changelog
:cl:
balance: keycard authentication devices require an ID to use.
balance: IDs used to initiate / confirm keycard auth events must not be the same ID.
/:cl:


Bokkiewokkie (approved):

DeltaFire15:
Upstream PR merged.

Bokkiewokkie (approved):

PR #2722, opened 2024-12-09
Emergency Jumps

DeltaFire15:
## About The Pull Request
Have you found yourself in a horrible situation? Do you have a bit of a deathwish? Do not fret, for the shipyard engineers have managed some new "safe" and "reliable" drive modifications.
This allows for a vessel to, at high risk to itself, perform a (uncontrolled) blind jump to some sector gravity well, bypassing any known forms of drive disruption (*including* attempts to abort the jump past its final confirmation).

This is intended as a last-resort sort of move where the only other real options would be to blow up or call the shuttle. This adds a third one. Its risks are weighted around that.

How to fry your drive:

1. Disable its safeties via the keycard authentication option.
2. Ensure the FTL is at least 25% prepared (Most of the spooling process is drivespace calculations, which are discarded for this jump type)
3. Attempt to jump to *any* system via a Navigation console.
4. Confirm the Manual Override via the FTL itself.

> [!Warning]
> Jumping without trajectory calculations is extremely hazardous, and may result in currently unknown phenomena due to drive field instabilities and inconsistencies in drive-space. Both short- and long-term effects are extremely likely.

> [!Note]
> The Drive Emergency Capacitor requires 35 minutes to recharge after brute-forcing a vessel's path (does not affect normal travel). Drive Safety Protocols must also manually be reengaged, regardless of whether a jump was performed or not.

This should be merged after https://github.com/BeeStation/NSV13/pull/2721 due to some keycard auth backend changes I'll have to merge conflict resolve with myself.

As always nothing is infinitely final.
(Cooldown period is subject to change and I am considering whether it should be a one-time option)

## Why It's Good For The Game
Options are fun. This is one that permits escape from a lot of situations, but has equal (or perhaps even worse) drawbacks associated with it. I don't see it used particularly often (for the crew's own good), but it should lead to some interesting outcomes.

<details>

Find out ~

</details>

## Changelog
:cl:
add: Emergency Jumps, a highly hazardous get-out-of-jail card.
soundadd: Sounds related to emergency jumps.
fix: relay() no longer tries to play null sounds if no sound is provided to it (something done quite a lot)
/:cl:


Pockets-byte:
This: this is amazing. Dear god this is amazing. 
I’m absolutely psyched to see it- what’s the plan for strong interdiction, because I’m guessing that weak currently applies to all of the existing interdictors- 
Also a couple suggestions- phase ghost got me thinking, bay’s bluespace jump effects are insanely cool, it might be cool to have something similar- and second, it puts me in mind of starsector where phase stuff is cool and creepy- perhaps later this could actually tie into some sort of antag / aggressor that you attract by doing this?

Bokkiewokkie (changes requested):

Bokkiewokkie (approved):

PR #2726, opened 2024-12-12
Batch of Fixes (1/? ?)

DeltaFire15:
## About The Pull Request
Soo a while ago a certain person sent me *several paragraphs* worth of bug reports **via** ***byond dms***.
This fixes some of them. There's still a LOT more, but for now I tried to nail the simpler ones first.

Also rewrites a bunch of things on how deckgun linkages work because I went insane while fixing the bugs related to those. (Deck gun turret cores were the basetype instead of a seperate type, leading to the complete uselessness of istype checks when trying to filter for cores since ALL machines related to deck guns would pass the check.)

Also improves a bunch of documentation (a lot of /proc definitions mostly, those that I noticed anyways)

ALSO managed to nail an ANCIENT AI mecha remote control bug that made its vision not update quite right.
WOOOO (I remember having issues with it as far back as 2019 on another server, haha!) (This is related to one of the reported bugs, but deserves a special place in here)

Attendum 1:
After some discussion, having mech remote control forcibly severed by a mech's destruction now causes damage equal to 33% of maximum integrity to the AI itself. This is to somewhat counteract the potential of wave-tactics especially in cross z combat scenarios, as well as create a small level of involvement to making the decision to disconnect (potentially leaving the mech in enemy hands).

## Why It's Good For The Game
Fix man good.

I have done extensive testing on most of these, but as always something might evade my claws.

## Changelog
:cl:
fix: Foam no longer persists infinitely when combined with hulljuice / firefoam (diminishing returns)
balance: Cooking Oil contact damage is now lower / more reasonable (I consider that one a bug for reasons I am not going into detail on)
fix: Love potions do not work on knpcs. (They feel no love)
fix: Several repair sources that bypassed the point of no return no longer do so.
fix: Several AI-related overmap checks should now be more sane (especially not being weird if not on an overmap).
fix: AIs no longer die if the exosuit that is being remote controlled is destroyed. They flash-sever the connection instead. (Assuming core exists and you used a beacon, not the malf override)
balance: To balance out AIs no longer immediately dying if they do not disconnect from a mech early, they now take 33% of maximum integrity in feedback damage if forcibly disconnected from a mech (via destruction). This should preserve at least some involvement in making the decision to disconnect.
fix: AI camera networks should behave a LOT better while remote controlling mechs, ESPECIALLY when taking them to other z levels, fixing a plethora of related bugs (mostly related to static, equipment usage, and so on).
fix: You can no longer set ships to autopilot which never were AI controlled in the first place.
fix: Sabres & Fighters can't buy things from stations anymore (Do it from a big ship instead)
fix: You cannot make Sabres / Fighters within another small craft. Use your time more wisely.
fix: The astrometrics sensors no longer pierce bluespace into other sectors (You'll have to be there to scan those)
fix: Pool ladders can no longer be built infinitely because somebody forgot their doafter mistrust.
code: Reworked a bunch of NAC code, especially around linkages, bad typing, unsafe stuff.. and so on.
fix: NAC cores no longer preserve linkage when unanchored.
fix: NAC machines no longer preserve linkage when unanchored (at least most of them).
fix: NAC machines can no longer be single-tile stacked (and work) (at least most of them).
fix: NAC machines can no longer be multilinked to several NACs (at least most of them).
fix: NAC machines should now behave better when being destroyed (code-wise).
tweak: NAC cores now tell you what is wrong during linkage.
tweak: A bunch of maps had their munitions cores changed due to the retyping of them I did because their type was very very bad.
code: take_quadrant_damage doesn't discard damage anymore if the ship has no quadrants (and just goes directly to take_damage)
/:cl:

### DIFF
--- a/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/deck_guns.dm
+++ b/nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/deck_guns.dm
@@ -24,7 +24,7 @@
 	maintainable = FALSE //This just makes them brick.
 	load_sound = 'nsv13/sound/effects/ship/freespace2/crane_short.ogg'
 	interaction_flags_machine = INTERACT_MACHINE_OPEN | INTERACT_MACHINE_OFFLINE | INTERACT_MACHINE_WIRES_IF_OPEN | INTERACT_MACHINE_ALLOW_SILICON | INTERACT_MACHINE_OPEN_SILICON | INTERACT_MACHINE_SET_MACHINE
-	var/obj/machinery/deck_turret/core
+	var/obj/machinery/deck_turret/core/core
 	var/id = null //N.B. This is NOT intended to allow them to manual link deck guns. This is for certain boarding maps and is thus a UNIQUE CONSTRAINT for this one case. ~KMC
 	circuit = /obj/item/circuitboard/machine/deck_turret
 
@@ -61,6 +61,9 @@
 /obj/machinery/ship_weapon/deck_turret/Destroy()
 	for( var/obj/item/ship_weapon/ammunition/naval_artillery/shell in ammo )
 		shell.speed = initial( shell.speed ) // Reset on turret destruction
+	if(core)
+		core.turret = null
+		core = null
 	return ..()
 
 /obj/machinery/ship_weapon/deck_turret/spawn_frame(disassembled)
@@ -93,7 +96,9 @@
 
 /obj/machinery/ship_weapon/deck_turret/multitool_act(mob/living/user, obj/item/I)
 	. = ..()
-	core = locate(/obj/machinery/deck_turret) in SSmapping.get_turf_below(src)
+	var/obj/machinery/deck_turret/core/maybe_core = locate(/obj/machinery/deck_turret/core) in SSmapping.get_turf_below(src)
+	if(!maybe_core.anchored)
+		return
 	if(!core)
 		link_via_id()
 	if(!core)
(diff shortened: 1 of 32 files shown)


Pockets-byte:
noooo- this gets rid of revolver NACs

nsv13/code/modules/munitions/ship_weapons/ballistic_weapons/deck_guns.dm
```diff
@@ -267,16 +286,107 @@
...
+				break
+			if(!payload_gate && istype(maybe_machine, /obj/machinery/deck_turret/payload_gate))
+				var/obj/machinery/deck_turret/payload_gate/maybe_payload_gate = maybe_machine
+				if(maybe_payload_gate.core)
+					continue
+				if(!maybe_payload_gate.anchored)
+					continue
+				payload_gate = maybe_payload_gate
+				maybe_payload_gate.core = src
+				break
```
Bokkiewokkie, line 388:
```suggestion
		for(var/obj/machinery/maybe_machine in T)
			switch(maybe_machine.type)
				if(/obj/machinery/deck_turret/powder_gate)
					var/obj/machinery/deck_turret/powder_gate/maybe_powder_gate = maybe_machine
					if(maybe_powder_gate.core)
						continue
					if(!maybe_powder_gate.anchored)
						continue
					powder_gates += maybe_powder_gate
					maybe_powder_gate.core = src
					break
				if(/obj/machinery/computer/deckgun)
					if(computer)
						continue
					var/obj/machinery/computer/deckgun/guncomputer = maybe_machine
					if(guncomputer.core)
						continue
					computer = guncomputer
					guncomputer.core = src
					break
				if(/obj/machinery/deck_turret/payload_gate)
					if(payload_gate)
						continue
					var/obj/machinery/deck_turret/payload_gate/maybe_payload_gate = maybe_machine
					if(maybe_payload_gate.core)
						continue
					if(!maybe_payload_gate.anchored)
						continue
					payload_gate = maybe_payload_gate
					maybe_payload_gate.core = src
					break
```
Could use a switch but I guess you do lose out on checking the gate/computer before the type in that.

Bokkiewokkie (approved):
Not required but recommended change:

Bokkiewokkie (approved):

PR #2730, opened 2024-12-17
[Experimental] Large ship durability increase

SerynEngi:
## About The Pull Request

This PR moderately increases the health and substantially increases the armor of the Tycoon, Gladius, and Galactica. This is entirely a buff, and does not touch other ships at all.

## Why It's Good For The Game

Larger ships should be able to take a lot more punishment. Ideally, this will encourage more stand-and-fight type engagements with the ships that are large enough to actually take said punishment in the first place.

Cruisers will now have 50% more medium armor and 400% more heavy armor than Atlas class vessels, enabling them to actually tank hits vice needing to somehow avoid them all with a much larger and slower hull.

Our sole dreadnaught will largely be immune to munitions that are not torpedoes or heavy munitions, enabling it to play like the dreadnaught it actually is.

## Changelog
:cl:
balance: Increased the armor and health of Tycoon, Gladius, and Galactica
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/types/nanotrasen.dm
+++ b/nsv13/code/modules/overmap/types/nanotrasen.dm
@@ -226,18 +226,18 @@
 
 /obj/structure/overmap/nanotrasen/battlecruiser/starter //Currently assigned to Tycoon and Gladius
 	role = MAIN_OVERMAP
-	obj_integrity = 1400
-	max_integrity = 1400
+	obj_integrity = 1600
+	max_integrity = 1600
 	starting_system = "Staging" //Required for all player ships
-	armor = list("overmap_light" = 95, "overmap_medium" = 60, "overmap_heavy" = 20)
+	armor = list("overmap_light" = 95, "overmap_medium" = 75, "overmap_heavy" = 50) //Cruisers should be able to take substantially more damage than Frigates and destroyers
 	overmap_deletion_traits = DAMAGE_STARTS_COUNTDOWN
 
 /obj/structure/overmap/nanotrasen/battleship/starter //Galactica
 	role = MAIN_OVERMAP //Player controlled variant
-	obj_integrity = 2150
-	max_integrity = 2150
+	obj_integrity = 2400
+	max_integrity = 2400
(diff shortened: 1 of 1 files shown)


Pockets-byte:
slightly worried about power creep, the galactica is already the tankiest ship in the game, with the high armor against anything that doesn't deal heavy, and the PDSR, that's limited immunity to every damage type in the game - in addition to a higher total hitpoint pool(x2 because armor)- I won't say that this is bad, because it isn't. but I also don't feel completely certain that it'll not ramp up the durability way too high

SerynEngi:
So, as I understand your comment. You functionally said: "I am concerned that this may make the Galactica too durable."

That's a fair comment! Now, keep in mind that I do not feel that balance should be done off of what a ship is theoretically capable of, but what is actually happening on a regular basis. Basically, you should not punish skill unless that skill required is mediocre.

This is labeled an experimental PR by intention. It is intended to be test-merged to see what happens, and tweaked accordingly. The intention is to create a gameplay style where larger ships are more willing to accept punishment in favor of kills. That starts by making it reasonably possible for that to happen.

someone543:
Shouldn't this also apply to the hammerhead? It is also a cruiser that uses the exact same hull and armor values as the current tycoon and gladius.

SerynEngi:
> Shouldn't this also apply to the hammerhead? It is also a cruiser that uses the exact same hull and armor values as the current tycoon and gladius.

Good question. I could, and I'll do so if the project leads request it, though I personally consider the Hammerhead to functionally be a destroyer. That said, it will probably help to have a "control" ship as it were that does not have the increased values while still having the same maneuverability.

To continue, the original intent is to check the most frequently selected ships of the larger sizes: the Gladius uses the same values as the Tycoon, or it would not have been selected. We'll use feedback to adjust these numbers.

Pockets-byte:
Valid, I'm looking forwards to it

DeltaFire15:
Please update this PR from master to have the CI tests run properly.

Bokkiewokkie (approved):

PR #2618, opened 2024-03-21
[UPSTREAM] TGUI Say

Bobbanz1:
## About The Pull Request
This PR ports TGUI Say from Beestation

It accomplishes this by pulling from upstream the following PR: 
* https://github.com/BeeStation/BeeStation-Hornet/pull/8404

That PR in turn ports TGUI Say from TG Station, those are in particular the following PRs:
* https://github.com/tgstation/tgstation/pull/67116
* https://github.com/tgstation/tgstation/pull/67971
* https://github.com/tgstation/tgstation/pull/67137
* https://github.com/tgstation/tgstation/pull/67265

What the Beestation port does is it implements TGUI Say. However it does it in such a way that allows the player to choose between either using the TGUI Say interface or the original interface. 

The PR also introduces the ability to let the player choose between using TGUI Inputs, such as tgui buttons/tgui lists and the original interface.

Additional Changes ported from Beestations TGUI Say PR:
* Added explo (although we don't use it), uplink, `:l`, `:r` and `:h` channel handling
* Added support for `.channel` prefixes
* Colors were tweaked
* The entire new window is draggable, excluding the textbox.
* The border has been tweaked slightly, it is not animated but has a minor gradient on both the top and bottom
* Added a border around the text input part of the box.
* Added LOOC channel (Quickly accessible by pressing L on your keyboard)
* Legacy support pref, with typing indicators and all
* Typing `;` in Say mode will convert the radio prefix to `;`, and removing it will go back to Say.
* Shift+TAB and right clicking the button will go back in channels.
* Re-opening the window on a different channel will no longer break the button contents.
* Added "show prefix" mode which treats `:s` as text in the input box while still updating the color and button.
* Pressing ESC on the game window with the say unfocused will exit the say.

SQL Update that needs to be done before test-merge and/or full merge
```sql
UPDATE SS13_preferences SET preference_value = preference_value | (1<<14) | (1<<15) | (1<<16) | (1<<17) WHERE preference_tag = 2;
```

Todo:
* [ ] Figure out how to record this stuff in action without reaching the github's file size limit

## Why It's Good For The Game
Keeps the codebase somewhat up to date with Beestation. 
This also makes it easier to import a future pull request from Beestation if we have the preferences and code required for it already in the codebase.

## Testing Photographs and Procedure

<details>
<summary>Screenshots&Videos</summary>

**TGUI Say Enabled:**
TGUISayPreferences
withTGUISayEnabled

**TGUI Say Disabled:**
TGUISayDisabled
[...]

PR #2741, opened 2025-01-06
Minor Nitpicks

francinum:
```
Fixes default/Legacy control
> This does *technically* mean that legacy can't use TGUI Say, but nobody using Legacy is going to care.

Adds macrotest debug tools

Adds Dream Daemon VSCode debug target
```

## About The Pull Request

Default input has been broken for 8 months since #2618, due to it directly binding printable keys on legacy users.

The only safe keys to hard-bind on legacy input users is in this random list from 2008.
https://www.byond.com/docs/notes/macro.html

This also comes with a set of janky tools for directly poking at the client skin. They aren't production safe and are behind a compile flag, `MACROTEST`

You can now compile targeting Dream Daemon, as god intended. (This boots faster and means your client doesn't try to crash itself when you hit a breakpoint)

## Why It's Good For The Game

I got nerd-sniped. I entered the VC as a cheap joke.

<details>

Nobody but me cares about this anyways.

</details>

## Changelog
:cl:
fix: Default/Legacy input now properly relays keys.
del: TGUI Say and Legacy Input are incompatible, Legacy will take precedence.
/:cl:

### DIFF
--- a/code/modules/keybindings/setup.dm
+++ b/code/modules/keybindings/setup.dm
@@ -36,6 +36,11 @@
 
 	erase_all_macros()
 
+	/// Is this client using Chat Relay/Legacy input mode. If so,
+	/// we need to be **VERY CAREFUL** about what keys we bind,
+	/// and we can't bind anything printable. -Francinum
+	var/using_chat_relay = (prefs.toggles2 & PREFTOGGLE_2_HOTKEYS)
+
 	var/list/macro_sets = SSinput.macro_sets
 	var/use_tgui_say = !prefs || (prefs.toggles2 & PREFTOGGLE_2_TGUI_SAY)
 	var/say = use_tgui_say ? tgui_say_create_open_command(SAY_CHANNEL) : "\".winset \\\"command=\\\".start_typing say\\\";command=.init_say;saywindow.is-visible=true;saywindow.input.focus=true\\\"\""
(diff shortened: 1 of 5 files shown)


Bokkiewokkie (approved):
Checks out

covertcorvid (dismissed):

code/modules/keybindings/setup.dm
```diff
@@ -52,13 +57,16 @@
...
-		winset(src, "[setname]-say", "parent=[setname];name=T;command=[say]")
-		winset(src, "[setname]-me", "parent=[setname];name=M;command=[me]")
-		winset(src, "[setname]-ooc", "parent=[setname];name=O;command=[ooc]")
-		if(use_tgui_say)
-			winset(src, "[setname]-radio", "parent=[setname];name=Y;command=[radio]")
-			winset(src, "[setname]-looc", "parent=[setname];name=L;command=[looc]") // NSV13 - Moves LOOC keybind to L from U
-			winset(src, "[setname]-close-tgui-say", "parent=[setname];name=Escape;command=[tgui_say_create_close_command()]")
+		// If we bind these, we're going to disable input relay.
+		// This *does* silently break TGUI-Say, but I doubt players who want to use this mode care. -Francinum
+		if(!using_chat_relay)
```
Bokkiewokkie, line 62:
```suggestion
		if(using_chat_relay)
```
This does the opposite right now (the `preftoggle hotkeys` var is used for people who are in *hotkey mode* and is enabled by default).
You may want to rename the variable (or invert it when checked initially) but I will push a change to edit the ! out myself now to fix this on live.

Bokkiewokkie (changes requested):

code/modules/keybindings/setup.dm
```diff
@@ -36,6 +36,11 @@
 
 	erase_all_macros()
 
+	/// Is this client using Chat Relay/Legacy input mode. If so,
+	/// we need to be **VERY CAREFUL** about what keys we bind,
+	/// and we can't bind anything printable. -Francinum
+	var/using_chat_relay = (prefs.toggles2 & PREFTOGGLE_2_HOTKEYS)
```
francinum, line 42: Fair enough, I'd prefer this fix instead to keep the var name correct.
```suggestion
	var/using_chat_relay = !(prefs.toggles2 & PREFTOGGLE_2_HOTKEYS)
```

code/modules/keybindings/setup.dm
```diff
@@ -52,13 +57,16 @@
...
-		winset(src, "[setname]-say", "parent=[setname];name=T;command=[say]")
-		winset(src, "[setname]-me", "parent=[setname];name=M;command=[me]")
-		winset(src, "[setname]-ooc", "parent=[setname];name=O;command=[ooc]")
-		if(use_tgui_say)
-			winset(src, "[setname]-radio", "parent=[setname];name=Y;command=[radio]")
-			winset(src, "[setname]-looc", "parent=[setname];name=L;command=[looc]") // NSV13 - Moves LOOC keybind to L from U
-			winset(src, "[setname]-close-tgui-say", "parent=[setname];name=Escape;command=[tgui_say_create_close_command()]")
+		// If we bind these, we're going to disable input relay.
+		// This *does* silently break TGUI-Say, but I doubt players who want to use this mode care. -Francinum
+		if(using_chat_relay)
```
francinum, line 62:
```suggestion
		// If we bind these, we're going to break the default command bar input relay behaviour.
		// This *does* mean we outright ignore the tgui-say pref, but I doubt players who want to use this mode care. -Francinum
		if(!using_chat_relay)
```

Bokkiewokkie (approved):
All good

PR #2740, opened 2025-01-03
Various bugfixes

Bokkiewokkie:
## About The Pull Request

Trivial

Fixes #2729 
Fixes #2500
Fixes #2263
Fixes #2554 
Fixes #2705 

## Why It's Good For The Game

Bunch of annoying bugs get fixed

## Testing
Tested all of this twice

## Changelog
:cl:
fix: Fixed durasteel tiles not stacking properly with different types
fix: Fixed tugs having "too much oxygen" to breathe
fix: Fixed dropship walls and entrances being deconstructable
fix: Fixed gauss racks letting you load empty crates
fix: Fixed stormdrive dividing by 0 when all control rods are removed
/:cl:

PR #2743, opened 2025-01-06
5% Entropy

Karmic-Skink:
## About The Pull Request
This PR upgrades the Stormdrive to have a little bit of randomness in gas interactions determined by each individual 'drive instance.

Note 1: This is achieved using variables and isn't hardcoded, allowing for subtype variations and admin trickery
Note 2: This includes the divide by zero fix in #2740 to avoid conflicts

## Why It's Good For The Game
More engaging engineering gameplay.

<details>

</details>

## Changelog
:cl:
tweak: Stormdrive gas modifiers are now subtly altered per instance
/:cl:

### DIFF
--- a/nsv13/code/modules/power/stormdrive.dm
+++ b/nsv13/code/modules/power/stormdrive.dm
@@ -152,6 +152,23 @@ Control Rods
 	var/reactor_end_times = FALSE //This is the end times for this reactor, we know this because the third error has been made
 	var/repairing = FALSE //Flag for SD repairs to stop duplicate actions
 
+	//Gas Variables Here
+	var/gas_modifier = 5 //Initially % modifier
+	var/constricted_plasma_modifier = 1
+	var/plasma_modifier = 1
+	var/tritium_modifier = 1
+	var/o2_modifier = 1
+	var/n2_modifier = 1
+	var/co2_modifier = 1
+	var/h2o_modifier = 1
+	var/nob_modifier = 1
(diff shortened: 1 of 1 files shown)


Bokkiewokkie (approved):

jet-pack-cat:
Could be intentional but [rand\(\) returns an integer](https://www.byond.com/docs/guide/chap16.html), so the variation is fixed to being either 100%, 99%, or 101% of the initial value.  

image
^ thats what that looks like in practice

You at most just need to do is to add x% more of n2, and or nitrous.

1% for this version, 2% if you have really really tight margins.. (There is a TINY bit of difference due to funny math). 
image
modified to match original (left) original (right)

This is already has a 17c margin for meltdown anyways.

Most mixes already have absurd coolant and reinforcement margins anyways. 
(wiki mix has a coolant margin of _600%_ for its normal operating temperature)

Wouldn't call it engaging personally, its just a simple multiplication for very very very optimized mixes :p

Bokkiewokkie:
> Could be intentional but [rand() returns an integer](https://www.byond.com/docs/guide/chap16.html), so the variation is fixed to being either 100%, 99%, or 101% of the initial value.
>

Yeah it's picking between integers of 95 and 105 which is a variation of 10%p 🤷‍♀️

DeltaFire15 (approved):
Seems to be fine and only affects externally overengineered mixes.

PR #2751, opened 2025-01-25
Lets space vines climb ladders and staircases

Bokkiewokkie:
## About The Pull Request

Does what it says, vines can now climb ladders and through open spaces
Mirrored in the upstream PR [#12192](https://github.com/BeeStation/BeeStation-Hornet/pull/12192)

## Why It's Good For The Game

Lets vines spread in a more logical way, also adding a bit more challenge to combating them :3c

<details>
I tested this but some pictures of vines growing aren't all that exciting
</details>

## Changelog
:cl:
balance: added the ability for vines to traverse multi-z through ladders and open space
/:cl:

### DIFF
--- a/code/modules/events/spacevine.dm
+++ b/code/modules/events/spacevine.dm
@@ -541,8 +541,20 @@
 		buckle_mob(V, 1)
 
 /obj/structure/spacevine/proc/spread()
-	var/direction = pick(GLOB.cardinals)
-	var/turf/stepturf = get_step(src,direction)
+	var/turf/startturf = get_turf(src)
+	var/direction = pick(GLOB.cardinals_multiz)
+	var/turf/stepturf = get_step_multiz(src,direction)
+	if(direction == UP || direction == DOWN)
+		var/ladder = FALSE
+		for(var/obj/structure/ladder/L in startturf.contents)
+			if(!L.up && direction == UP)
+				continue
+			else if(!L.down && direction == DOWN)
+				continue
+			ladder = TRUE
+		if(!stepturf?.zPassIn(src, direction, startturf) && !ladder) //We can't go up, choose another direction
+			direction = pick(GLOB.cardinals)
+			stepturf = get_step(src,direction)
 	if(locate(/obj/structure, stepturf) || locate(/obj/machinery, stepturf))//if we can't grow into a turf, we'll start digging into it
 		for(var/obj/structure/S in stepturf)
 			if(S.density && !istype(S, /obj/structure/alien/resin/flower_bud_enemy) && !istype(S, /obj/structure/reagent_dispensers/fueltank)) //don't breach the station!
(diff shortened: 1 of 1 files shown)


Pockets-byte:
yeah so vines now grow through the ceiling
vines attacking everywhere
*scream

Bokkiewokkie:
> yeah so vines now grow through the ceiling vines attacking everywhere *scream

Strange, they even check if there's a direct path up

Pockets-byte:
> > yeah so vines now grow through the ceiling vines attacking everywhere *scream
> 
> Strange, they even check if there's a direct path up

Nono- I wasn’t clear enough- they only ignore floors, they work going up

Bokkiewokkie:
> > > yeah so vines now grow through the ceiling vines attacking everywhere *scream
> > 
> > 
> > Strange, they even check if there's a direct path up
> 
> Nono- I wasn’t clear enough- they only ignore floors, they work going up

Oh I see, thanks

code/modules/events/spacevine.dm
```diff
@@ -541,8 +541,20 @@
...
+	var/turf/stepturf = get_step_multiz(src,direction)
+	if(direction == UP || direction == DOWN)
+		var/ladder = FALSE
+		for(var/obj/structure/ladder/L in startturf.contents)
+			if(!L.up && direction == UP)
+				continue
+			else if(!L.down && direction == DOWN)
+				continue
+			ladder = TRUE
+		if(!stepturf?.zPassIn(src, direction, startturf) && !startturf.zPassOut(src, direction, stepturf) && !ladder) //We can't zmove, choose another direction
```
DeltaFire15, line 555: Should probably fail if either doesn't allow passage, not neccessarily both?
```suggestion
		if((!stepturf?.zPassIn(src, direction, startturf) || !startturf.zPassOut(src, direction, stepturf)) && !ladder) //We can't zmove, choose another direction
```

DeltaFire15 (changes requested):

code/modules/events/spacevine.dm
Bokkiewokkie: No, because this is direction-dependent. You want to check both if the turf you are entering allows it from your direction, and if the turf you are exiting also allows you to leave in that direction

Bokkiewokkie: Of course, for turfs like floors they always let you leave and enter from above them, but imagine a vine trying to travel down openspace into a wall below

DeltaFire15: But.. this is the check that if true *prevents* movement is it not?

So right now if you can move out of the current tile, but cannot enter the one you want to go to... the proc is false and allows movement, aka the exact case you describe here, right?
Which means you want to intercept if *either* blocking case if present, and block movement if it is except if a ladder is there, no?

Bokkiewokkie: Apparently bee did not realize this in my fix and just merged it in 3 days >.>

DeltaFire15 (approved):
code looks fine; content scares me.

PR #2755, opened 2025-01-30
Adds the ability to construct AI cores on Syndicate ships and make Syndicate helmet cams

Bokkiewokkie:
## About The Pull Request

The title pretty much sums it up, right now an AI that is *constructed* (not spawned) on a Syndicate ship will get a law 0 to stay loyal to the Syndicate, and be linked to the Syndie camera net and radio channel.
I've also included the ability to make Syndicate helmet cameras, because with an AI those will have more purpose, too. Boarders will spawn with cameras in their helmets by default.
> [!Note]
> All AI technology on Syndicate ships is designated to immediately be transferred to Minervasoft upon arrival in a Syndicate shipyard. Withholding transfer or staying quiet about possessing AI technology will have severe repercussions.

## Why It's Good For The Game

Adds support for more shenanigans during PVP, and in general.

<details>
I tested it, trust
</details>

## Changelog
:cl:
add: Added Syndicate AI core construction
add: Added the ability for Syndies to install syndicate cameras to helmets
balance: Syndicate boarder helmet now have cameras
refactor: Refactored helmet camera code a bit
/:cl:

### DIFF
--- a/code/game/objects/structures/ai_core.dm
+++ b/code/game/objects/structures/ai_core.dm
@@ -244,7 +244,7 @@
 							A = new /mob/living/silicon/ai(loc, laws, brain.brainmob)
 
 						if(get_overmap()?.faction == "syndicate") //NSV13 start - Syndicate ship radio & traitor setting
-							A.set_zeroth_law("Accomplish your ship's objectives at all costs.", "Accomplish your ship's objectives at all costs.")
+							A.set_zeroth_law("Accomplish the Syndicate's objectives at all costs.", "Accomplish your Syndicate's objectives at all costs.")
 							A.set_syndie_radio()
 							A.faction |= "Syndicate" //Why is this capitalized
 							A.network = list("syndicate")
--- a/code/game/objects/structures/ai_core.dm
+++ b/code/game/objects/structures/ai_core.dm
@@ -243,6 +243,13 @@
 						else
 							A = new /mob/living/silicon/ai(loc, laws, brain.brainmob)
 
+						if(get_overmap()?.faction == "syndicate") //NSV13 start - Syndicate ship radio & traitor setting
+							A.set_zeroth_law("Accomplish your ship's objectives at all costs.", "Accomplish your ship's objectives at all costs.")
+							A.set_syndie_radio()
+							A.faction |= "Syndicate" //Why is this capitalized
+							A.network = list("syndicate")
+							A.radio.recalculateChannels() //NSV13 end
+
 						if(brain.force_replace_ai_name)
 							A.fully_replace_character_name(A.name, brain.replacement_ai_name())
 						SSblackbox.record_feedback("amount", "ais_created", 1)
--- a/nsv13/code/modules/clothing/helmet_camera.dm
+++ b/nsv13/code/modules/clothing/helmet_camera.dm
@@ -6,7 +6,8 @@
 		if(src == user.get_item_by_slot(ITEM_SLOT_HEAD)) //Make sure the player is not wearing the suit before applying the upgrade.
 			to_chat(user, "<span class='warning'>You cannot install the upgrade to [src] while wearing it.</span>")
 			return FALSE
-		if(do_after(install_camera(I, src), 5, target = src))
+		if(do_after(user, 5, target = src))
+			install_camera(I, src)
 			to_chat(user, "<span class='notice'>You successfully attach the camera to [src].</span>")
 			return TRUE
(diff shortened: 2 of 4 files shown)


nsv13/code/modules/clothing/helmet_camera.dm
```diff
@@ -64,3 +59,26 @@
...
+		if(faction == "Syndicate")
+			builtInCamera.network = list("syndicate")
+		else
+			builtInCamera.network = list("ss13")
+			builtInCamera.internal_light = FALSE
+		return TRUE
+	user.transferItemToLoc(I, src)
+	builtInCamera = new /obj/machinery/camera(src)
+	builtInCamera.c_tag = "Helmet Cam #[rand(0,999)]"
+	if(user.faction & "Syndicate" || faction == "Syndicate") //If a syndicate agent puts it on, it's a syndie camera now
```
DeltaFire15, line 78: This won't work as expected, as first of all `&` binds very weakly so it should be contained within parentheses, and additionally, `&` when used with a list always gives back a list, which means the condition is always true (as an empty list evaluates as true within if). You will have to either `length()` check the result, or use `in` or a comparable operation instead.

nsv13/code/modules/clothing/helmet_camera.dm
```diff
@@ -2,30 +2,25 @@
...
 			to_chat(user, "<span class='warning'>You cannot install the upgrade to [src] while wearing it.</span>")
-			return
-		if(user.transferItemToLoc(I, src))
-			builtInCamera = new /obj/machinery/camera(src)
-			builtInCamera.c_tag = "Helmet Cam #[rand(0,999)]"
-			builtInCamera.network = list("headcam", "ss13")
-			builtInCamera.internal_light = FALSE
-			QDEL_NULL(I)
+			return FALSE
+		if(do_after(install_camera(I, src), 5, target = src))
```
DeltaFire15, line 9: The first argument of `do_after` is the user. The current call will immediately install the camera regardless of whether the do_after succeeds (and will probably runtime because the user is a `TRUE`).

code/game/objects/structures/ai_core.dm
```diff
@@ -243,6 +243,13 @@
 						else
 							A = new /mob/living/silicon/ai(loc, laws, brain.brainmob)
 
+						if(get_overmap()?.faction == "syndicate") //NSV13 start - Syndicate ship radio & traitor setting
+							A.set_zeroth_law("Accomplish the Syndicate's objectives at all costs.", "Accomplish your Syndicate's objectives at all costs.")
```
DeltaFire15, line 247: SyndieOS lawset (or an adaptation of it akin to NSV crewsimov) instead? Think it may be neat to have neither side's AI immune to law change based interference. Although this is more of a thought and not a required change.

DeltaFire15 (changes requested):

nsv13/code/modules/clothing/helmet_camera.dm
Bokkiewokkie: I forgot :p

code/game/objects/structures/ai_core.dm
Bokkiewokkie: Yeah right now this just makes it law 0 be syndicate and the rest is NSV default (or whatever law board was put in manually when building)

Bokkiewokkie: Maybe it could be fun to have the laws/allegiance change with the IFF, just a thought for later though.

nsv13/code/modules/clothing/helmet_camera.dm
```diff
@@ -64,3 +60,26 @@
...
+	if(builtInCamera) //You already have a camera
+		return FALSE
+	if(!I || !user) //The camera isn't being placed by a person
+		builtInCamera = new /obj/machinery/camera(src)
+		builtInCamera.c_tag = "Helmet Cam #[rand(0,999)]"
+		if(faction == "Syndicate")
+			builtInCamera.network = list("syndicate")
+		else
+			builtInCamera.network = list("ss13")
+			builtInCamera.internal_light = FALSE
```
DeltaFire15, line 74: `internal_light = FALSE` is likely indented one too far since it should do it in either case.

DeltaFire15 (changes requested):

nsv13/code/modules/clothing/helmet_camera.dm
Bokkiewokkie: Yep...

PR #2758, opened 2025-02-02
Delta alert level cannot be manually switched out of.

DeltaFire15:
## About The Pull Request
Currently, you can just switch off delta alert via keycard auth / comms console, despite the fact delta alert is a very special alert level only triggered / disengaged via special causes. This makes it so you cannot switch off delta alert manually, only by resolving the source (which should avoid confusion).

## Why It's Good For The Game
No, you don't get to pretend the ship is actually fine when there is an active nuke.

No.

## Changelog
:cl:
fix: You cannot just change off delta alert.
/:cl:


Bokkiewokkie (approved):
All good, looks like you didn't miss any other cases

DuskLight9978:
The ships fiiiiiine. we dont (do) need this. 

Jokes a side. A lovely fix to see

PR #2762, opened 2025-02-09
Player Missile/Torpedo durability buffs

SerynEngi:
## About The Pull Request

This pull request creates a subtype of all missiles and torpedoes for AI usage and then buffs the durability player-only variants by 200%

## Why It's Good For The Game

Missiles and torpedoes should be effective. Here, we have a single player ship sometimes fighting huge fleets. This should help those weapons actually be effective when facing more than 2 or 3 enemies. 

## Changelog
:cl:
add: Added variants of torpedos and missiles for AI usage
balance: Buffed the durability of player ship missiles and torps
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/overmap.dm
+++ b/nsv13/code/modules/overmap/overmap.dm
@@ -195,6 +195,7 @@
 	var/reserved_z = 0 //The Z level we were spawned on, and thus inhabit. This can be changed if we "swap" positions with another ship.
 	var/list/occupying_levels = list() //Refs to the z-levels we own for setting parallax and that, or for admins to debug things when EVERYTHING INEVITABLY BREAKS
 	var/torpedo_type = /obj/item/projectile/guided_munition/torpedo
+	var/missile_type = /obj/item/projectile/guided_munition/missile
 	var/next_maneuvre = 0 //When can we pull off a fancy trick like boost or kinetic turn?
 	var/flak_battery_amount = 0
 	var/broadside = FALSE //Whether the ship is allowed to have broadside cannons or not
--- a/nsv13/code/modules/overmap/types/nanotrasen.dm
+++ b/nsv13/code/modules/overmap/types/nanotrasen.dm
@@ -268,6 +268,8 @@
 	mass = MASS_MEDIUM
 	sprite_size = 48
 	ai_flags= AI_FLAG_DESTROYER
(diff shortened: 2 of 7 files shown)


SerynEngi:
This was balanced for 1 player VS fleet, and can be tuned up or down fairly easily to desired outcome. The goal here is that it isnt every missile or torp that makes it through, but is instead that some missiles make it more than 3 tiles away from mainship or fightership before getting annihilated when facing a fleet.

I am aware that this may create some _interesting_ scenarios in PVP, however missiles and torps are very expensive to build and take substantially more time to produce. In this scenario, ideally it creates a choice between effective missiles at several times the cost, or fast and heavy hitting NAC shells.

Bokkiewokkie (changes requested):
Looks like there are some instances of Hellfire and Disruptor torpedoes you've missed in this. Otherwise it looks good

SerynEngi:
So, I may have forgotten, in my haste, that hellfire and disruptor torps already have a separate subtype that players use, and unlike _every other torpedo and missile_ I instead buffed the previously existing player variant. Reverted to the second commit.

DeltaFire15 (changes requested):
Leaving decisions on whether this is good to higher dev staff (because I still believe torp is pretty good if not fired straight into the foe's barrage),

however code-wise please replace any `/nerfed` with `/ai`, because that is what they are.

nsv13/code/modules/overmap/overmap.dm
```diff
@@ -195,6 +195,7 @@
 	var/reserved_z = 0 //The Z level we were spawned on, and thus inhabit. This can be changed if we "swap" positions with another ship.
 	var/list/occupying_levels = list() //Refs to the z-levels we own for setting parallax and that, or for admins to debug things when EVERYTHING INEVITABLY BREAKS
 	var/torpedo_type = /obj/item/projectile/guided_munition/torpedo
+	var/missile_type = /obj/item/projectile/guided_munition/missile
```
DeltaFire15, line 198: Unused variable. Hook into `fire_missile()`.

DeltaFire15 (changes requested):

nsv13/code/modules/overmap/overmap.dm
SerynEngi: Thank you! I'd wondered what I missed!

Bokkiewokkie (approved):

PR #2779, opened 2025-03-07
Fix flap and aflap emotes

red031000:
## About The Pull Request

fixes the flap and aflap emotes so that mobs without wings don't flap their non-existant wings

## Why It's Good For The Game

casue it's a bit weird to flap your wings when you don't have any

## Changelog
:cl:
fix: fix mobs being able to flap their wings when they don't have wings
/:cl:

### DIFF
--- a/code/modules/mob/living/emote.dm
+++ b/code/modules/mob/living/emote.dm
@@ -114,7 +114,7 @@
 /datum/emote/living/flap
 	key = "flap"
 	key_third_person = "flaps"
-	message = "flaps their wings"
+	message = "flaps their"
 	restraint_check = TRUE
 	var/wing_time = 10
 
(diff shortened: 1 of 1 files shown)


code/modules/mob/living/emote.dm
```diff
@@ -125,13 +125,33 @@
 		if(H.Togglewings())
 			addtimer(CALLBACK(H,TYPE_PROC_REF(/mob/living/carbon/human, Togglewings)), wing_time)
 
+/datum/emote/living/flap/select_message_type(mob/user, intentional)
+	. = ..()
+	var/wings = FALSE
+	if(ishuman(user))
+		var/mob/living/carbon/human/H = user
+		if(H.dna && H.dna.species)
+			if((H.dna.features["wings"] != "None") || H.dna.features["moth_wings"] != "None")
```
Bokkiewokkie, line 134: These DNA features are inherited from your preferences, and can be set even when you are not a moth (and more often than not are just a random moth wing selection). As such checking for this is only useful if you have already checked if they even have the wing organ.

Bokkiewokkie (changes requested):

code/modules/mob/living/emote.dm
Bokkiewokkie: A good example of what you can do is given in /datum/emote/living/carbon/human/wing (line 122 in emote.dm)

red031000: guh
dumb
will change that

Bokkiewokkie (approved):

DeltaFire15 (approved):

PR #2792, opened 2025-03-26
Syndicate boarders are now more averse to capture

DeltaFire15:
## About The Pull Request
Did you know you can use explosive implants even while incapacitated, crit, or otherwise unable to do most things?
Well, syndicate boarders didn't, so their implants usually remained unused until biological death (unless.. nudged).
Now they do know. Be careful around those.

## Why It's Good For The Game
A player playing one of those would absolutely hit the button when in an inescapable situation, and knpcs deserve the same.

No. (But this is tested)

## Changelog
:cl:
add: Syndicate boarders are fairly likely to react rather explosively to attempts to subdue them.
fix: Some knpc voicelines now correctly declare they are forced by the game.
/:cl:

### DIFF
--- a/nsv13/code/modules/overmap/knpc.dm
+++ b/nsv13/code/modules/overmap/knpc.dm
@@ -273,13 +273,13 @@ GLOBAL_LIST_EMPTY(knpcs)
 	var/mob/living/carbon/human/ai_boarder/H = parent
 	if(H.stat == DEAD)
 		return PROCESS_KILL
-	if(!H.can_resist())
-		if(H.incapacitated()) //In crit or something....
-			return
 	if(world.time >= next_action)
 		next_action = world.time + H.action_delay
 		pick_goal()
 		current_goal?.action(src)
+	if(!H.can_resist())
+		if(H.incapacitated()) //In crit or something....
+			return
 	if(length(path))
 		next_path_step()
 	else //They should always be pathing somewhere...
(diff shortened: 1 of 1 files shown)


nsv13/code/modules/overmap/knpc.dm
```diff
@@ -61,6 +65,7 @@ GLOBAL_LIST_EMPTY(knpcs)
 /datum/component/knpc/Initialize()
 	if(!iscarbon(parent))
 		return COMPONENT_INCOMPATIBLE
+	failsafe_trust = rand((2 SECONDS), (5 SECONDS))
```
Bokkiewokkie, line 68: Why did we define it if it's changed here? Maybe you can incorporate the value into the calculation or let setting it be an override for this (so people can make subtypes with different values easily)

DeltaFire15: Uhhh good question.
But yeah makes sense.

nsv13/code/modules/overmap/knpc.dm
```diff
@@ -65,7 +67,8 @@ GLOBAL_LIST_EMPTY(knpcs)
 /datum/component/knpc/Initialize()
 	if(!iscarbon(parent))
 		return COMPONENT_INCOMPATIBLE
-	failsafe_trust = rand((2 SECONDS), (5 SECONDS))
+	if(failsafe_trust_variance)
+		failsafe_trust = rand(failsafe_trust, (failsafe_trust + failsafe_trust_variance))
```
Bokkiewokkie, line 71: Since failsafe trust variance is pretty much always set there's not really a reason to do an if check I think. Even 0 will still just resolve.

DeltaFire15: I guess, yeah

PR #2799, opened 2025-03-29
Hammerheads pumps rotated to match ship orientation

riquewright:
## About The Pull Request

Mirrors the pumps horizontally as to make them match the ship map's orientation
## Why It's Good For The Game

Less headache when changing the pumps around

</details>

## Changelog
:cl:
tweak: Hammerhead pumps rotated to match ship orientation
/:cl:

### DIFF
--- a/_maps/map_files/Hammerhead/Hammerhead.dmm
+++ b/_maps/map_files/Hammerhead/Hammerhead.dmm
@@ -24839,6 +24839,12 @@
 	},
 /turf/open/floor/wood,
 /area/library)
+"eEb" = (
+/obj/item/toy/plush/hfighter{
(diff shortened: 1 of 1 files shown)


Bokkiewokkie (changes requested):
You put a plushie in space.
Plushies cannot rotate, so setting their `dir` will have no effect.

riquewright:
Seems like I accidentally deleted the email notification about those, my bad

DeltaFire15:
> [...] Plushies cannot rotate [...]

Theoretically would be doable by having a subtype that calls whatever the rotate proc was on it on initialize (but would still be a bit prone to external factors).

PR #2816, opened 2025-05-31
Package wrap inherits more inconveniencing properties

DeltaFire15:
## About The Pull Request
Currently, wrapping items is quite troublesome, as it allows bypassing most, if not all, restrictions for picking up, carrying, or handling things.
This PR makes the wrapped parcel inherit several properties of the wrapped item, such as being unable to pick it up, or density. Additionally, parcels of the highest weight class need two hands to carry, even if they do not contain an object marked as impossible to pick up.
Parcels can also be cut open with any sharp item now, as some would be impossible to open otherwise.

Also resolves some oddities with naval artillery shells from their transfer between files, namely them being onehanded items now. Bokkie mentioned they should be twohanded, so I did that as opposed to preventing any inhand transport as before.
This is not the case for cannonballs, which can only be carried in zero-g.

And as usual also fixes a completely random bug with a disease causing runtimes.

## Why It's Good For The Game
Some things are difficult to carry for a reason. Wrap should not completely bypass any and all properties of such items.
Oh and fix man good, also.

Unfortunately, this is tested.

## Changelog
:cl:
balance: Package-wrap wrapped items (parcels) now inherit several traits of the original item which complicate transport of the item.
add: Parcels can now be cut open with any sharp object.
fix: NAC shell transport behaves more reasonable again, requiring two hands to handle (cannonballs are off limits when gravity is on).
fix: The autobrewery symtom no longer randomly runtimes.
/:cl:


williamcll:
Doesn't this make wrapping redundant?

Bokkiewokkie (approved):

DeltaFire15:
Whoops that branch wasn't merged 😅

PR #2876, opened 2026-04-01
atmospheric retention fields (ports https://github.com/BeeStation/BeeStation-Hornet/pull/11550)

Pockets-byte:
## About The Pull Request

ports bee's port of shiptest's holofields. (https://github.com/BeeStation/BeeStation-Hornet/pull/11550)
sprites are from tg's atmospheric force fields. (https://github.com/tgstation/tgstation/pull/91055)
## Why It's Good For The Game
Solgov gets force-field barriers, so you can experience staring past the glowing blue shield into where the rest of the ship used to be. 
also slapstick comedy when you lose power and all the fancy tech suddenly fails you.
as long as power is maintained they're a flat upgrade to firelocks. they allow people to pass but not gasses or fire. 

<details>

</details>

## Changelog
:cl: Tsar-Salat mc-oofert
add: Added new atmospheric barrier firelocks
code: shieldwalls are now machinery/power
/:cl:


DeltaFire15:
self-assigning this so I get reminded to look at it when it is done.

Pockets-byte:
ready for review!

code/game/area/areas.dm
```diff
@@ -369,6 +370,25 @@ GLOBAL_LIST_EMPTY(teleportlocs)
 				else if(!(D.density ^ opening))
 					INVOKE_ASYNC(D, (opening ? TYPE_PROC_REF(/obj/machinery/door/firedoor, open) : TYPE_PROC_REF(/obj/machinery/door/, close)))
 
+/area/proc/ModifyAtmosshields(opening)
+	if(atmosshields)
+		for(var/eachshield in atmosshields)
+			var/obj/machinery/power/shieldwallgen/atmos/shield = eachshield
```
DeltaFire15: typecast that in the `for` itself, no need to do it with a var in the line after

code/game/area/areas.dm
```diff
@@ -369,6 +370,25 @@ GLOBAL_LIST_EMPTY(teleportlocs)
...
+			if(opening)
+				for(var/b in shield.affecting_areas)
+					var/area/area2 = b
```
DeltaFire15: likewise, typecast in the `for` itself. Also preferably use a more descriptive name than area2.

code/game/area/areas.dm
```diff
@@ -369,6 +370,25 @@ GLOBAL_LIST_EMPTY(teleportlocs)
...
+					if(area2.fire)
+						break
+			if(shield.is_operational & opening & shield.shieldstate > 0)
```
DeltaFire15: `&&` instead of `&` in all cases here.

code/game/area/areas.dm
```diff
@@ -369,6 +370,25 @@ GLOBAL_LIST_EMPTY(teleportlocs)
...
+				INVOKE_ASYNC(shield.toggle())
+			else if(shield.is_operational & shield.shieldstate == 0)
```
DeltaFire15: `isoperational` check should be pulled in front of these ifs to `continue` if not met since you are checking it in either case.

code/game/area/areas.dm
```diff
@@ -369,6 +370,25 @@ GLOBAL_LIST_EMPTY(teleportlocs)
...
+				INVOKE_ASYNC(shield.toggle())
+
+/area/proc/shieldsup()
+	for(var/obj/machinery/power/shieldwallgen/atmos/shield in atmosshields)
+		shield.rapidsetup()
+		shield.breachalert = 1
```
DeltaFire15: should be using `TRUE` / `FALSE` for yes/no cases.

code/game/machinery/shieldgen.dm
```diff
@@ -14,7 +14,7 @@
 /obj/structure/emergency_shield/Initialize(mapload)
 	. = ..()
 	setDir(pick(GLOB.cardinals))
-	air_update_turf()
+	air_update_turf(1)
```
DeltaFire15, line 17: ?
proc doesn't have any arguments.

code/game/machinery/shieldgen.dm
```diff
@@ -97,7 +97,7 @@
 
 /obj/machinery/shieldgen/proc/shields_up()
 	active = TRUE
-	update_icon()
+	update_appearance()
```
DeltaFire15: is there a reason this now does a full update appearance update as opposed to only icon?

code/game/machinery/shieldgen.dm
```diff
@@ -108,7 +108,7 @@
 /obj/machinery/shieldgen/proc/shields_down()
 	active = FALSE
 	move_resist = initial(move_resist)
-	update_icon()
+	update_appearance()
```
DeltaFire15: likewise?

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
 
-#define ACTIVE_SETUPFIELDS 1
-#define ACTIVE_HASFIELDS 2
-/obj/machinery/shieldwallgen
+/obj/machinery/power/shieldwallgen
 	name = "shield wall generator"
 	desc = "A shield generator."
 	icon = 'icons/obj/stationobjs.dmi'
-	icon_state = "Shield_Gen"
+	icon_state = "shield_wall_gen"
```
DeltaFire15, line 229: since you are renaming the icon state, remove the now unused icons from the file (as they're now duplicates as far as I can tell).

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
+	var/shieldstate = SHIELD_NOTACTIVE
+	var/id = null
 	var/locked = TRUE
 	var/shield_range = 8
+	var/shocked = FALSE
+	var/list/affecting_areas
+	var/fields = 0
+	var/buffer = 3000
+	var/max_buffer = 3000
 	var/obj/structure/cable/attached // the attached cable
```
DeltaFire15: unused var.

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
 	max_integrity = 300
-	var/active = FALSE
-	var/power = 0
-	var/maximum_stored_power = 500
+	var/shieldstate = SHIELD_NOTACTIVE
+	var/id = null
 	var/locked = TRUE
 	var/shield_range = 8
+	var/shocked = FALSE
+	var/list/affecting_areas
```
DeltaFire15, line 246: does this need to be on the base shieldwall generator type? Not used on it.

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
+		if(!active_power_usage || surplus() >= active_power_usage)
+			add_load(active_power_usage)
+		else
 			visible_message("<span class='danger'>The [src.name] shuts down due to lack of power!</span>", \
-				"<span class='italics'>You hear heavy droning fade out.</span>")
-			icon_state = "Shield_Gen"
-			active = FALSE
-			for(var/d in GLOB.cardinals)
-				cleanup_field(d)
+				"If this message is ever seen, something is wrong.",
```
DeltaFire15: this quasi-error-message shouldn't be in here because it will never be seen *including* if something goes wrong and just wastes a minimal amount of processing power into another call in visual message that returns out immediately. Use named arguments for later arguments instead (`argname = value`)

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
 	else
-		icon_state = "Shield_Gen"
-		for(var/d in GLOB.cardinals)
-			cleanup_field(d)
+		for(var/direction in GLOB.cardinals)
+			cleanup_field(direction)
+	update_appearance()
+
+/obj/machinery/power/shieldwallgen/proc/rapidsetup()
+	if(buffer & !shieldstate)
```
DeltaFire15: `&&`

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
+		for(var/direction in GLOB.cardinals)
+			setup_field(direction)
+			fields++
+			shieldstate = SHIELD_HASFIELDS
```
DeltaFire15: should probably check for if it actually ended up with fields as opposed to this?
(similar to standard startup procedure in `process()`)

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
 		return SINGULARITY_TRY_MOVE_BLOCK
 
-/obj/machinery/shieldwallgen/can_be_unfasten_wrench(mob/user, silent)
-	if(active)
+/obj/machinery/power/shieldwallgen/can_be_unfasten_wrench(mob/user, silent)
+	if(shieldstate)
 		if(!silent)
 			to_chat(user, "<span class='warning'>Turn off the shield generator first!</span>")
 		return FAILED_UNFASTEN
+	remove_from_areas()
```
DeltaFire15: this is odd (and probably doesn't work).
instead hook into `setAnchored()` and call only `remove_from_areas()` if it's a successful unanchor action (e.g. anchorvalue is 0).

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
-	if(W.tool_behaviour == TOOL_WRENCH)
-		default_unfasten_wrench(user, W, 0)
+/obj/machinery/power/shieldwallgen/wrench_act(mob/living/user, obj/item/item)
+	. = ..()
+	. |= default_unfasten_wrench(user, item, 0)
+//	update_cable_icons_on_turf(T) - Removed because smartwire Revert
+	if(. == SUCCESSFUL_UNFASTEN && anchored)
+		connect_to_network()
+
+/obj/machinery/power/shieldwallgen/setAnchored()
```
DeltaFire15: missing argument (`anchorvalue`)
(also the proc where the stuff from previous comment should go)

code/game/machinery/shieldgen.dm
```diff
@@ -374,35 +428,55 @@
...
 
-/obj/machinery/shieldwallgen/emag_act(mob/user)
+/obj/machinery/power/shieldwallgen/proc/toggle()
+	if(!anchored)
+		return
+	if(!powernet)
+		return
+	if(shieldstate)
+		visible_message("<span class= 'notice'>The [src.name] hums as it powers down.</span>", \
+			"If this message is ever seen, something is wrong.", \
```
DeltaFire15: pointless error message

code/game/machinery/shieldgen.dm
```diff
@@ -374,35 +428,55 @@
...
+			"<span class= 'notice'>You hear heavy droning fade out.</span>")
+		playsound(src, 'sound/machines/synth_no.ogg', 50, TRUE, frequency = 6120)
+		shieldstate = SHIELD_NOTACTIVE
+		log_game("[src] was deactivated by wire pulse at [AREACOORD(src)]")
+	else
+		visible_message("<span class= 'notice'>The [src.name] beeps as it powers up.</span>", \
+			"If this message is ever seen, something is wrong.", \
```
DeltaFire15: pointless error message.

nsv13/code/datums/wires/shieldwallgen.dm
```diff
@@ -0,0 +1,46 @@
...
+	status += "The interface light is [generator.locked ? "red" : "green"]."
+	status += "The activity light is [generator.shieldstate ? "blinking steadily" : "off"]."
+	return status
+
+/datum/wires/shieldwallgen/on_pulse(wire)
+	var/obj/machinery/power/shieldwallgen/generator = holder
+	switch(wire)
+		if(WIRE_SHOCK)
+			generator.shocked = !generator.shocked
+			addtimer(CALLBACK(generator, TYPE_PROC_REF(/obj/machinery/modular_fabricator/autolathe, reset), wire), 60)
```
DeltaFire15: someone sure did some bad copypasta when writing this part originally. Reference the generator type instead of the autolathe.

code/_compile_options.dm
```diff
@@ -60,7 +60,7 @@
 								//	2 for preloading absolutely everything;
 
 #ifdef LOWMEMORYMODE
-#define FORCE_MAP "runtimestation"
+//#define FORCE_MAP "runtimestation"
```
DeltaFire15: debug change present in pushed code.

code/modules/atmospherics/environmental/LINDA_turf_tile.dm
```diff
@@ -203,6 +203,12 @@
 		FD.emergency_pressure_stop()
 	for(var/obj/machinery/door/firedoor/FD in src)
 		FD.emergency_pressure_stop()
+	var/area/A = get_area(src)
+	A.shieldsup()
+	var/area/A2 = get_area(T2)
+	A2.shieldsup()
```
DeltaFire15, line 209: potentially high processing load, we can see about a better way to handle this later (hit me up once the other stuff is resolved).
Can probably have some fun listening signal the generators deploy & signal that onto the turf here.

_maps/map_files/Aetherwhisp/Aetherwhisp1.dmm
DeltaFire15, line 1: should probably be using a `/anchored` type for maps instead of mapediting anchored to true.

(also note to self: there's still debugging stuff on these maps (I saw a void core)

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+		if(T2 in T.atmos_adjacent_turfs)
+			T.atmos_adjacent_turfs[T2] |= ATMOS_ADJACENT_FIRELOCK
+		if(T in T2.atmos_adjacent_turfs)
+			T2.atmos_adjacent_turfs[T] |= ATMOS_ADJACENT_FIRELOCK
+
+/obj/machinery/power/shieldwallgen/proc/remove_from_areas()
+	if(affecting_areas)
+		for(var/I in affecting_areas)
+			var/area/A = I
+			LAZYREMOVE(A.atmosshields, src)
```
DeltaFire15, line 555: do these need to be on `/shieldwallgen` instead of on the `/atmos` subtype?

(similar to comment on `affecting_areas` on the object definition)

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+	var/turf/turf = loc
+	for(var/direction in GLOB.cardinals)
+		for(var/i in 1 to shield_range)
+			turf = get_step(turf, direction)
+	for(var/t in T.atmos_adjacent_turfs & turf.atmos_adjacent_turfs)
+		var/turf/open/T2 = t
+		if(T2 in T.atmos_adjacent_turfs)
+			T.atmos_adjacent_turfs[T2] |= ATMOS_ADJACENT_FIRELOCK
+		if(T in T2.atmos_adjacent_turfs)
+			T2.atmos_adjacent_turfs[T] |= ATMOS_ADJACENT_FIRELOCK
```
DeltaFire15, line 549: note to self: check how this works.

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+	icon_state = "shield_wall_gen_atmos"
+	circuit = /obj/item/circuitboard/machine/shieldwallgen/atmos
+	anchored = FALSE
+	density = FALSE
+	req_access = list()
+	locked = FALSE
+	shield_range = 8
+	var/breachalert = 0
+	layer = WALL_OBJ_LAYER
+	wire_compatible = FALSE
```
DeltaFire15: this doesn't do anything (variable unused on machine level and also `FALSE` by default)

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+
+/obj/machinery/power/shieldwallgen/atmos/roundstart
+	anchored = TRUE
+	shieldstate = SHIELD_SETUPFIELDS
+	active_power_usage = 0
```
DeltaFire15: why do these have 0 power drain?

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+	name = "high power atmospheric barrier generator"
+	desc = "A barrier generator designed for use in hangar bays."
+	circuit = /obj/item/circuitboard/machine/shieldwallgen/atmos/strong
+	shield_range = 20
+	active_power_usage = 10000
+
+/obj/machinery/power/shieldwallgen/atmos/strong/roundstart
+	anchored = TRUE
+	shieldstate = SHIELD_SETUPFIELDS
+	active_power_usage = 0
```
DeltaFire15: likewise. (power usage)

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+
+/obj/machinery/power/shieldwallgen/ruin
+	anchored = TRUE
+	shieldstate = SHIELD_SETUPFIELDS
+	active_power_usage = 0
```
DeltaFire15, line 597: likewise (sort of, unless the area is unpowered)

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+/obj/machinery/power/shieldwallgen/atmos
+	name = "atmospheric barrier generator"
+	desc = "A barrier generator designed for use in ship loading bays."
+	icon = 'icons/obj/stationobjs.dmi'
+	icon_state = "shield_wall_gen_atmos"
+	circuit = /obj/item/circuitboard/machine/shieldwallgen/atmos
+	anchored = FALSE
+	density = FALSE
+	req_access = list()
+	locked = FALSE
```
DeltaFire15: Why are these not locked to engineering or atmos access by default?

code/game/machinery/shieldgen.dm
```diff
@@ -411,6 +485,147 @@
...
+		return
+	if(direction != dir)
+		return
+
+	var/turf/turf = loc
+	var/obj/machinery/power/shieldwallgen/generator
+	var/steps = 0
+	var/opposite_direction = turn(direction, 180)
+
+	for(var/i in 1 to shield_range) //checks out to 8 tiles away for another generator
```
DeltaFire15: comment is wrong due to variable range.

code/game/machinery/shieldgen.dm
DeltaFire15: should have a subtype for atmos shields that only executes this in a single direction (or maybe the proc should just be moved there in its entirety.

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
-	var/maximum_stored_power = 500
+	var/shieldstate = SHIELD_NOTACTIVE
+	var/id = null
 	var/locked = TRUE
 	var/shield_range = 8
+	var/shocked = FALSE
+	var/list/affecting_areas
+	var/fields = 0
+	var/buffer = 3000
+	var/max_buffer = 3000
```
DeltaFire15: These for some reason are now only used on the atmos shield subtype, removing the power buffer from standard shield walls that was there before. Please restore it (see: former "`power`" variable / `use_stored_power()` handling).

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+//atmos blocking shieldwalls for shiptest use
+/obj/machinery/shieldwall/atmos
+	name = "atmospheric barrier"
+	desc = "An energy shield capable of blocking gas movement."
+	icon = 'nsv13/icons/obj/smooth_structures/atmosshield.dmi'
+	icon_state = "atmosshield-3"
+	density = FALSE
+	use_power = NO_POWER_USE
+	CanAtmosPass = ATMOS_PASS_NO
+	CanAtmosPassVertical = 1
```
DeltaFire15: why is vertical atmos pass permitted if it's a barrier that blocks atmos in all four non-z cardinals?

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+	air_update_turf(TRUE)
+
+/obj/machinery/power/shieldwallgen/atmos/toggle()
+	if(!anchored)
+		return
+	if(buffer <= 0)
+		return
+	if(shieldstate)
+		visible_message("<span class= 'notice'>The [src.name] hums as it powers down.</span>", \
+			"If this message is ever seen, something is wrong.", \
```
DeltaFire15: pointless error message (see previous mentions)

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+			"<span class= 'notice'>You hear heavy droning fade out.</span>")
+		playsound(src, 'sound/machines/synth_no.ogg', 50, TRUE, frequency = 6120)
+		shieldstate = SHIELD_NOTACTIVE
+		breachalert = 0
+		log_game("[src] was deactivated by wire pulse at [AREACOORD(src)]")
+	else
+		visible_message("<span class= 'notice'>The [src.name] beeps as it powers up.</span>", \
+			"If this message is ever seen, something is wrong.", \
```
DeltaFire15: pointless error message (see previous mentions)

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+			"<span class= 'notice'>You hear heavy droning.</span>")
+		shieldstate = SHIELD_SETUPFIELDS
+		breachalert = 1
+		log_game("[src] was activated by wire pulse at [AREACOORD(src)]")
```
DeltaFire15: not always true due to other things calling toggle.

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+	if(buffer <= 0)
+		return
+	if(shieldstate)
+		visible_message("<span class= 'notice'>The [src.name] hums as it powers down.</span>", \
+			"If this message is ever seen, something is wrong.", \
+			"<span class= 'notice'>You hear heavy droning fade out.</span>")
+		playsound(src, 'sound/machines/synth_no.ogg', 50, TRUE, frequency = 6120)
+		shieldstate = SHIELD_NOTACTIVE
+		breachalert = 0
+		log_game("[src] was deactivated by wire pulse at [AREACOORD(src)]")
```
DeltaFire15: not always true due to other things calling toggle.

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+		breachalert = 1
+		log_game("[src] was activated by wire pulse at [AREACOORD(src)]")
+
+/obj/machinery/shieldwall/atmos/drain_power(drain_amount)
+	if(needs_power && gen_primary)
+		gen_primary.use_power(drain_amount * 1)
+		gen_primary.buffer -= drain_amount
+		if(gen_secondary) //using power may cause us to be destroyed
+			gen_secondary.use_power(drain_amount * 1)
+			gen_secondary.buffer -= drain_amount
```
DeltaFire15: is the point of buffer not to buffer only if the generator is not powered? This drains it always.

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+
+/obj/machinery/power/shieldwallgen/atmos/process()
+	if(is_operational & buffer <= max_buffer )	//I'm in the dark about this. my brain no work
+		use_power(active_power_usage)
+		buffer = clamp(buffer + 11, 0, max_buffer)
```
DeltaFire15: odd magicnumber in buffer addition

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+/obj/machinery/shieldwall/atmos/drain_power(drain_amount)
+	if(needs_power && gen_primary)
+		gen_primary.use_power(drain_amount * 1)
+		gen_primary.buffer -= drain_amount
+		if(gen_secondary) //using power may cause us to be destroyed
+			gen_secondary.use_power(drain_amount * 1)
+			gen_secondary.buffer -= drain_amount
+
+/obj/machinery/power/shieldwallgen/atmos/process()
+	if(is_operational & buffer <= max_buffer )	//I'm in the dark about this. my brain no work
```
DeltaFire15: `&&`

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+	density = FALSE
+	use_power = NO_POWER_USE
+	CanAtmosPass = ATMOS_PASS_NO
+	CanAtmosPassVertical = 1
+	hardshield = FALSE
+	layer = ABOVE_MOB_LAYER
+	light_color = "#61a3ff"
+	light_system = MOVABLE_LIGHT //for instant visual feedback regardless of lag
+	//Atmos shields suck more power
+	active_power_usage = 5000
```
DeltaFire15: presently buffer at full will last less than one tick before collapsing due to this power usage (if that power usage were used as opposed to a hard 1)
Probably either seperate buffer drain and power usage, handle buffer as a time / counter, or adjust buffer value.

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+		shieldstate = SHIELD_SETUPFIELDS
+		breachalert = 1
+		log_game("[src] was activated by wire pulse at [AREACOORD(src)]")
+
+/obj/machinery/shieldwall/atmos/drain_power(drain_amount)
+	if(needs_power && gen_primary)
+		gen_primary.use_power(drain_amount * 1)
+		gen_primary.buffer -= drain_amount
+		if(gen_secondary) //using power may cause us to be destroyed
+			gen_secondary.use_power(drain_amount * 1)
```
DeltaFire15: these `use_power()` calls presently ignore if the generator does not have an active power usage - is that intentional?

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+	if(shieldstate)
+		if(shieldstate == SHIELD_SETUPFIELDS)
+			fields = 0
+			for(var/direction in GLOB.cardinals)
+				if(setup_field(direction))
+					fields++
+			if(fields)
+				shieldstate = SHIELD_HASFIELDS
+		if(buffer > 10)
+		else
```
DeltaFire15: huh? why is this an if-else instead of just an if for the lower case
(also why the 10 as opposed to if there is no buffer or if a buffer drain fails?)

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+			visible_message("<span class='danger'>The [src.name] shuts down due to lack of power!</span>", \
+				"If this message is ever seen, something is wrong.",
```
DeltaFire15: pointless error message

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+	if(!anchored)
+		to_chat(user, "<span class='warning'>\The [src] needs to be firmly secured to the floor first!</span>")
+		return
+	if(locked && !issilicon(user))
+		to_chat(user, "<span class='warning'>The controls are locked!</span>")
+		return
+	if(!is_operational)
+		to_chat(user, "<span class='warning'>\The [src] needs to be powered!</span>")
+		return
+	if(buffer <= 25)
```
DeltaFire15: probably `define` the minimum stored value.

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+/obj/machinery/power/shieldwallgen/atmos/proc/latetoggle()
+	if(!buffer || !breachalert)
+		return
+	else
+		rapidsetup()
+
+/obj/machinery/shieldwall/atmos/take_damage(damage_amount, damage_type = BRUTE, damage_flag = 0, sound_effect = 1, attack_dir, armour_penetration = 0)
+	. = ..()
+	if(damage_type == BRUTE || damage_type == BURN)
+		drain_power(damage_amount*1000)
```
DeltaFire15: likewise (as with comment about passive buffer drain), will collapse immediately at the smallest damage (though that may be intentional in this case)

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
 	desc = "A shield generator."
 	icon = 'icons/obj/stationobjs.dmi'
-	icon_state = "Shield_Gen"
+	icon_state = "shield_wall_gen"
 	anchored = FALSE
 	density = TRUE
 	req_access = list(ACCESS_TELEPORTER)
 	flags_1 = CONDUCT_1
-	use_power = NO_POWER_USE
+	use_power = IDLE_POWER_USE
```
DeltaFire15: why does this have APC power drain if it's a gridlinked machine? Should move that to the subtype that does run off APCs.

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+	if(damage_type == BRUTE || damage_type == BURN)
+		drain_power(damage_amount*1000)
+
+/obj/machinery/shieldwall/atmos/process()
+	if(needs_power)
+		if(!gen_primary || !gen_primary.shieldstate || !gen_secondary || !gen_secondary.shieldstate)
+			qdel(src)
+			return
+
+		drain_power(1)
```
DeltaFire15: why is `active_power_usage` set on atmos fields if it is never used?

code/game/machinery/shieldgen.dm
```diff
@@ -420,20 +635,26 @@
...
 		setDir(get_dir(gen_primary, gen_secondary))
-	for(var/mob/living/L in get_turf(src))
-		visible_message("<span class='danger'>\The [src] is suddenly occupying the same space as \the [L]!</span>")
-		L.gib()
+	if(hardshield == TRUE)
+		for(var/mob/living/victim in get_turf(src))
+			visible_message("<span class='danger'>\The [src] is suddenly occupying the same space as \the [victim]!</span>")
+			victim.investigate_log("has been gibbed by [src].", ADMIN_INVESTIGATE_TARGET)
+			victim.gib()
 	RegisterSignal(src, COMSIG_ATOM_SINGULARITY_TRY_MOVE, PROC_REF(block_singularity))
```
DeltaFire15: ensure the atmos shieldwall does not possess this trait (e.g. by having a var for this and checking for it here).

code/game/machinery/shieldgen.dm
```diff
@@ -202,166 +202,220 @@
...
+	var/list/affecting_areas
+	var/fields = 0
+	var/buffer = 3000
+	var/max_buffer = 3000
 	var/obj/structure/cable/attached // the attached cable
 
-/obj/machinery/shieldwallgen/Initialize(mapload)
+/obj/machinery/power/shieldwallgen/Initialize(mapload)
 	. = ..()
 	RegisterSignal(src, COMSIG_ATOM_SINGULARITY_TRY_MOVE, PROC_REF(block_singularity_if_active))
```
DeltaFire15: ensure the atmos shield generator does not possess this trait (e.g. by having a var for this and checking for it here).

DeltaFire15 (changes requested):
phew
preliminary review.

will go over it again deeper once things mentioned here are handled (some functionality I know doesn't work right but I don't want to debug that deeper while other issues might be having domino effects)

EDIT:
OH right and the attribution in the changelog should probably just be usernames as opposed to a link since I don't know if the changelog generator would take that well.

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+
+/obj/machinery/power/shieldwallgen/atmos/interact(mob/user)
+	if(interaction_flags_atom & INTERACT_ATOM_NO_FINGERPRINT_INTERACT)
+		add_hiddenprint(user)
 	else
-		if(istype(mover, /obj/item/projectile))
-			return prob(10)
+		add_fingerprint(user)
+	if(interaction_flags_atom & INTERACT_ATOM_UI_INTERACT)
+		return ui_interact(user)
```
DeltaFire15: these are unneeded, except for a generic `add_fingerprint(user)` call (since you know what state the flags are in here)

code/game/machinery/shieldgen.dm
```diff
@@ -443,39 +664,189 @@
...
+	if(interaction_flags_machine & INTERACT_MACHINE_SET_MACHINE)
+		user.set_machine(src)
+	if(.)
+		return
```
DeltaFire15: this is unneeded since . is not changed.

code/game/machinery/shieldgen.dm
Pockets-byte: Holdover from shiptest I believe- these are subtypes for ruins, probably should rename them inline with the ruins subtype I added

code/game/machinery/shieldgen.dm
Pockets-byte: I moved it there because one of the procs would fail because it couldn't find it, looking at the old logic,I wasn't actually sure how much of a difference it made because the old logic took the surplus and stuffed it into the power, but also I'm not confident in recreating the old logic so I put it off

code/game/machinery/shieldgen.dm
Pockets-byte: I have no clue, have to check but it might be a holdover from bee

code/game/machinery/shieldgen.dm
Pockets-byte: I did notice that, figured I'd leave it as is since it's logging and most of what triggers that is either buttons, wiring, or ai I believe

code/game/machinery/shieldgen.dm
Pockets-byte: should be handled by the check for hardfield directly above, I think

code/game/machinery/shieldgen.dm
Pockets-byte: it's used ostentably to recharge the buffer

code/game/area/areas.dm
Pockets-byte: area2 is used in the reference I used, I don't actually know what this does, I can guess but uh- the atmos stuff works by being directly copied from firelock handling

code/game/area/areas.dm
Pockets-byte: isoperational is already wrong anyway, I want to check for buffer not powersupply

code/game/machinery/shieldgen.dm
Pockets-byte: like that in firelocks, assumed it didn't need arguments

code/game/machinery/shieldgen.dm
Pockets-byte: was how it's done on bee, did not know enough if it was a relevant change

code/game/machinery/shieldgen.dm
Pockets-byte: I think it caused some errors when I had it only on the atmoshield, but I have to check again

code/game/machinery/shieldgen.dm
Pockets-byte: unsure

code/game/machinery/shieldgen.dm
Pockets-byte: or really, since you can adjust the power usage in the map I should remove the power usage part

code/game/machinery/shieldgen.dm
Pockets-byte: I discovered that yeah, I suppose it would be better to figure out how to make it more precise for timing but I don't know.

code/game/machinery/shieldgen.dm
Pockets-byte: not really no-

code/game/machinery/shieldgen.dm
Pockets-byte: I'm looking to have it mostly drain from the buffer and the buffer uses power to recharge maybe

code/game/machinery/shieldgen.dm
Pockets-byte: I set the max tiles to ten, so 11 is enough to have +1 buffer while it's running.

code/game/machinery/shieldgen.dm
Pockets-byte: ten so that it actually has a failstate instead of instantly recharging up maybe, there was another thing here at one point so I think that's why it's this way

code/game/machinery/shieldgen.dm
Pockets-byte: was like that on bee, no idea why, I keep forgetting what . does

code/game/machinery/shieldgen.dm
Pockets-byte: it is- for example if we add an emag behavior to make it switch to a hardshield I don't want it to be as strong as the shieldwall
takes about 4-3 hits with an ion saber to down.

code/game/machinery/shieldgen.dm
Pockets-byte: oh active power use- I could use that yeah

code/game/machinery/shieldgen.dm
Pockets-byte: this right?

code/game/machinery/shieldgen.dm
Pockets-byte: what do you mean in a single direction? shieldgens still need to make shields

code/game/machinery/shieldgen.dm
Pockets-byte: unsure, need to test

code/game/machinery/shieldgen.dm
Pockets-byte: used for the abandoned space zoo to curcumvent power requirements

code/game/machinery/shieldgen.dm
Pockets-byte: fixed(I think)

code/game/area/areas.dm
Pockets-byte: fixed, maybe?

code/game/machinery/shieldgen.dm
Pockets-byte: likely fixed

code/game/machinery/shieldgen.dm
Pockets-byte: it was used at one point, still trying to figure out the buffer system

code/game/machinery/shieldgen.dm
Pockets-byte: should be fix

code/game/machinery/shieldgen.dm
Pockets-byte: less magic now

code/game/machinery/shieldgen.dm
Pockets-byte: proc doesn't want any arguments as far as I can tell

code/game/machinery/shieldgen.dm
Pockets-byte: old logic restored, regression fixed.

code/game/machinery/shieldgen.dm
Pockets-byte: might be covered looking at it? setanchored() has some stuff in it

code/game/machinery/shieldgen.dm
Pockets-byte: no longer set on atmos fields.

code/_compile_options.dm
Pockets-byte: fixed

PR #2909, opened 2026-07-24
adds a dedicated fighter shield

Pockets-byte:
solgov fighter now fits it's initial design
the shield is now a replacement for armor, and can be destroyed if you take too many hits on the hull. it properly removes itself when destroyed or removed. 

<details>

image
image

</details>

## Changelog
:cl:
add: added a fighter shield
del: Removed free shields from the solgov fighter, buffed it's health.
/:cl:


nsv13/code/modules/overmap/fighters/_fighters.dm
```diff
@@ -1147,6 +1146,39 @@ due_to_damage: If the removal was caused voluntarily (FALSE), or if it was cause
...
+			overmap_shields.set_stats(0, 0, 0)
+			overmap_shields.RemoveComponent()
+		return //We don't reset our health if the plating was destroyed due to hits, or the increase would be useless. It DOES get reset once we install new armor, though.
+	target.max_integrity = initial(target.max_integrity)
+	if(overmap_shields)
+		target.visible_message("<span class='warning'>The shield generator shuts down.</span>")
+		overmap_shields.set_stats(0, 0, 0)
+		overmap_shields.RemoveComponent()
+	//Remove any overheal.
+	target.obj_integrity = CLAMP(target.obj_integrity, 0, target.max_integrity)
```
Bokkiewokkie:
```suggestion
/obj/item/fighter_component/armour_plating/tier6/remove_from(obj/structure/overmap/target, due_to_damage)
	. = ..()
	if(overmap_shields)
		overmap_shields.set_stats(0, 0, 0)
		overmap_shields.RemoveComponent()
		if(due_to_damage)
			target.visible_message("<span class='warning'>The shield generator shuts down, melting into slag.</span>")
		else
			target.visible_message("<span class='warning'>The shield generator shuts down.</span>")
```
You can refactor this to check for the shields first to avoid duplicate code (which also makes it a bit easier to follow)
The other lines also don't need to be present, as ..() will call the parent proc which already does the same thing.

nsv13/code/modules/overmap/fighters/_fighters.dm
```diff
@@ -1147,6 +1146,39 @@ due_to_damage: If the removal was caused voluntarily (FALSE), or if it was cause
...
+	weight = 3.5
+	obj_integrity = 25
+	max_integrity = 25
+	var/datum/component/overmap_shields/overmap_shields
+
+
+/obj/item/fighter_component/armour_plating/tier6/on_install(obj/structure/overmap/target)
+	. = ..()
+	overmap_shields = target.AddComponent(/datum/component/overmap_shields, 125, 125, 15) //inital integrity, max integrity, and recharge rate. bound to change most likely
+	target.visible_message("<span class='warning'>The shield generator starts up, humming loudly.</span>")
```
Bokkiewokkie:
```suggestion
	target.audible_message("<span class='notice'>\The [src] starts up, humming loudly.</span>")
```
This isn't dangerous, so it can just be a notice, and audible message makes more sense if it's humming.
On top of that, using \The [src] helps with modularity by making the reference flexible, you can apply these to the other text as well.

nsv13/code/modules/overmap/fighters/_fighters.dm
```diff
@@ -1147,6 +1146,39 @@ due_to_damage: If the removal was caused voluntarily (FALSE), or if it was cause
 	obj_integrity = 300
 	max_integrity = 300
 
+/obj/item/fighter_component/armour_plating/tier6
+	name = "Compact Shield Generator Array"
```
Bokkiewokkie:
```suggestion
	name = "compact shield generator array"
```
Avoid capitalization on improper nouns, or use \improper to indicate it as such if the capitalization makes sense (like with acronyms at the start of names). Byond's text system automatically changes all references to be proper otherwise (e.g. on examine it will show "That's the compact shield generator array" instead of "That's a compact shield generator array"). For objects that usually isn't needed unless they are unique (like the nuclear authentication disk)

nsv13/code/modules/overmap/fighters/_fighters.dm
```diff
@@ -1147,6 +1146,39 @@ due_to_damage: If the removal was caused voluntarily (FALSE), or if it was cause
...
+	desc = "A rather heavy, and highly illegal shield generator, paired with a fine, lightly armored self-applicating emmiter mesh. simply looking at this has already doomed you to a life in prison."
+	icon_state = "shield"
+	tier = 6	//it's top tier solgov tech.
+	weight = 3.5
```
Bokkiewokkie:
```suggestion
	weight = 1.3
```
Unsure if you tested this out in game but I wouldn't make it any heavier than the tier 2 armor plates, one benefit of shields is that they don't weigh as much as covering your ship in literal metal. Fighters suffer a lot from heavy armor, which is why the tier 3 one is actually lighter.

Bokkiewokkie (changes requested):

nsv13/code/modules/overmap/fighters/_fighters.dm
Pockets-byte: I honestly didn't think about that- thank you, still learning some of these things

nsv13/code/modules/overmap/fighters/_fighters.dm
Pockets-byte: then I'm a little confused, I missed that that was included, if the parent proc is called, it wouldn't remove the shields because If(due_to_damage) fires, and returns before the shield can remove itself, right?

Bokkiewokkie (approved):

PR #2910, opened 2026-07-24
Total sound death fix

Pockets-byte:
I fixed missing sound channels across the game, hopefully preventing all instances of all channels being purged without intention to do so from happening. 
no more mental bandwidth for more at this moment
Also fixed a years old typo in sound channel datums

I went through and fixed every instance of SEND_SOUND() and RELAY() without a channel by hand
also can confirm stop_sounds was not broken. I intentionally did not fix it for a few instances where it makes sense. 

<details>

</details>

## Changelog
:cl:
fix: fixed sounds using every channel in the game occasionally by making sure all sounds barring a few have a set channel
fix: fixed a years old typo in soundchannels
/:cl:


Bokkiewokkie:
Please update your changelog to actually reflect your changes. I'd also rather have all of the channels here actually set to something that's not a random available one, but I won't punish you for finding a bug and fixing it by forcing you to, and I appreciate you already doing that for several of them regardless. :p

Pockets-byte:
> Please update your changelog to actually reflect your changes. I'd also rather have all of the channels here actually set to something that's not a random available one, but I won't punish you for finding a bug and fixing it by forcing you to, and I appreciate you already doing that for several of them regardless. :p

I'll be honest, at some point my brain gave up and I stopped thinking about it
there were so many....
other problem with that is you have to consider the things they might force out with set channels- like BWOINKs are fine in the admin channel, but if we put damage sounds in SHIP_FX for example, they might shut off the sound of the asteroid magnet, so it's not like you can just choose a channel to put them in q.q you actually have to consider what will be and should be overridden or not.

